PR #45, opened 2018-02-15
fix to create viewer with a title including non-ascii characters on Windows.

DavidJenny:


tomvanmele:
hi david, 

could you give an example of such a title? i would like to run a few tests, because string encoding is very different in python 2.7 and 3.x

thanks,
tom

DavidJenny:
Hi Tom

Actually I was also struggling with the string encodings when I was trying to use the Network OpenGL viewer.  Firstly, we tried something like this (viewer.py, line 132):

b = bytes(mystring, 'utf-8')

b = mystring.encode('utf-8')

I then saw that you had already updated the viewer file in the meantime to this:

glutCreateWindow(str.encode(self.title))

This solved the problem but (on Windows) titles such as “ETH Zürich”, “Grüezi”,  etc. didn’t display correctly. That’s why we added the 'Windows-1252' encoding.

I guess this is only relevant for Windows system and would be better to have a generic solution. ‘utf-8’ encoding did not work for us though.

Kind regards
David

gonzalocasas:
We'll need to find a different solution for this.

PR #57, opened 2018-03-19
Frame and Transformation

romanarust:
Added classes `Frame`, `Transformation`, `Rotation`, etc in geometry/objects and merged some functions with geometry/transformations.py. To avoid ambiguities I changed *rotation_matrix* to *matrix_from_axis_and_angle*, since there are now more options of creating a "rotation_matrix". i.e *matrix_from_euler_angles*, *matrix_from_quaternion*, *matrix_from_frame*, etc. I followed the naming for *matrix_from_translation*, *matrix_from_scale_factors*, etc.


gonzalocasas:
@worbit @tmsmendez @andrewtliew @tomvanmele would you guys have time to check this out? we would really like to have it in soon-ish if possible. Thx!

tomvanmele:
i am on it!

tomvanmele:
i will write comments here as i go through it...

tomvanmele:
always inherit from the base `object` when you create a new class definition...

```python
class Frame():
```
```python
class Frame(object):
```

tomvanmele:
as a suggestion, perhaps factory methods that don't follow the convention `from_xxx` could start with a capital letter since they return an instance...

```python
@classmethod
def worldXY(cls):
    return cls(...)
```
```python
@classmethod
def WorldXY(cls):
    return cls(...)
```

tomvanmele:
i prefer `Parameters` over `Args` for the docstring sections

tomvanmele:
backslashes are not needed in docstrings for line continuations

tomvanmele:
general note (for everyone including me and for the entire framework):

it is confusing if the name of an attribute does not suggest whether it is a method or a property, for example by including an action in the name or not...

this is especially the case if the use of descriptors (`@property`) is mixed with callable attributes to return information. with the latter being used when options can be specified.

for example:

```python
    @property
    def axis_and_angle(self):
        """Returns the axis and the angle of the ``Rotation``.
        """
        return axis_and_angle_from_matrix(self.matrix)
```
```python
    def euler_angles(self, static=True, axes='xyz'):
        """Returns Euler angles from the ``Rotation`` according to specified
        axis sequence and rotation type.

        Returns:
            (:obj:`list` of :obj:`float`): The 3 Euler angles.
        """
        return euler_angles_from_matrix(self.matrix, static, axes)

```

PR #74, opened 2018-03-20
fix #73: try..except all layer-related imports that fail in GHPython context

### LINKED ISSUE #73: Cannot import `compas_rhino` from Grasshopper (GHPython)
Currently it is not possible to import any `compas_rhino` functionality from Grasshopper, mainly because of three lines:
 - [this one on `layers.py`](https://github.com/compas-dev/compas/blob/master/src/compas_rhino/utilities/layers.py#L10). 
 - [this one on `objects.py`](https://github.com/compas-dev/compas/blob/master/src/compas_rhino/utilities/objects.py#L8). 
 - [and this one on `drawing.py`](https://github.com/compas-dev/compas/blob/master/src/compas_rhino/utilities/drawing.py#L37)

Regarding the decision of an additional `compas_grasshopper` package, the impossibility to import the namespace is very confusing, especially because the error message is also very misleading.

To reproduce, simply drop a GHPython component on an empty GH canvas, and type `import compas_rhino`:

image

gonzalocasas:
Since we discussed different options about the GH support, and this is definitely not a long-term solution, I propose it as a PR.

This small fix at least makes the `compas_rhino` module importable, so things like `RhinoMesh` can be used from Grasshopper, and also several of the utility functions at least.

Perhaps the support we discussed for `xdraw_*` (or `draw_*` according to the upcoming rename), could be implemented directly on `compas_rhino`, detecting the context, instead of creating a separate `compas_grasshopper` one.

### DIFF
--- a/src/compas_rhino/utilities/drawing.py
+++ b/src/compas_rhino/utilities/drawing.py
@@ -34,7 +34,6 @@ try:
     from Rhino.DocObjects.ObjectPlotWeightSource import PlotWeightFromObject
 
     find_object = sc.doc.Objects.Find
-    find_layer_by_fullpath = sc.doc.Layers.FindByFullPath
     add_point = sc.doc.Objects.AddPoint
     add_line = sc.doc.Objects.AddLine
     add_dot = sc.doc.Objects.AddTextDot
@@ -49,6 +48,11 @@ except ImportError:
     import platform
     if platform.python_implementation() == 'IronPython':
         raise
+
+try:
+    find_layer_by_fullpath = sc.doc.Layers.FindByFullPath
+except SystemError:
+    find_layer_by_fullpath = None
 
 
 __author__     = ['Tom Van Mele', ]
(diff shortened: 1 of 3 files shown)


gonzalocasas:
ping

tomvanmele:
drawing.py seems okay

tomvanmele:
in `layers.py` i think it would make more sense to just ignore the `purge` keyword when `purge_object` is not available. the way you set it up now, nothing happens when `purge` is `True` and `purge_object` is not available.

tomvanmele:
in `objects.py` you seem to have done as suggested in the previous comment. or am i missing a key difference between the two?!

gonzalocasas:
@brgcode you're totally right. i'll fix it asap. thx!

tomvanmele (commented):
so, as said in the comments, it seems that the absence of purge_object is handled differently in layers.py and objects.py. The latter seems correct...

gonzalocasas:
@brgcode fixed review findings on https://github.com/compas-dev/compas/pull/74/commits/d9ba7cd46c152e0faaef109a62fb3977a0f66f48

PR #117, opened 2018-07-27
Add URDF support and compas.robots model

gonzalocasas:
This pull request adds parser support for the [URDF format](http://wiki.ros.org/urdf/) to be able to exchange robot model descriptions and prepare compatibility with [ROS](http://www.ros.org/).

In detail, the additions are:
 * URDF parser infrastructure
 * Robot model representations for all standard URDF nodes and some extensions (more should be added in the future)
 * A memoize decorator, just because I needed one

There is also documentation for the new modules and some basic unit tests for core features of them: 
image


romanarust:
Hi Gonzalo,
I'm trying to load the UR5 urdf but I get a key error:
```
File "C:\Users\rustr\workspace\compas\src\compas\files\urdf.py", line 121, in parse_element
    parser_type = cls._parsers[element.tag]
KeyError: 'plugin'
```
Maybe because "plugin" has no child? If I remove the plugin tag, I'll get the next key error at "color" (= single line child)

Please try the following gist:
https://gist.github.com/romanarust/1b34ae0eb60f984bfc08e64210a1530e

gonzalocasas:
@romanarust thanks, I'll fix it. So far, I was not doing anything special about unknown tags (since there are lots of extensions to URDF) but perhaps a generic tag parsing for unknown tags would be a good tradeoff.

gonzalocasas:
@romanarust could you pls check again? Besides testing it with your URDF, I also tested against a large collection of URDF from the [Drake project](https://github.com/RobotLocomotion/drake), and made several adjustments to defaults and other non-standard attributes/elements.

PR #120, opened 2018-07-31
Add continuous integration via Travis

gonzalocasas:
Configuration to add continuous integration to compas using [Travis-CI](https://travis-ci.org/). The proposed config does the following:
 * It runs the installer of the repository on both Python 2.7 and 3.6
 * It runs unit tests on both configurations (2.7 and 3.6) (for now, there's almost no unit tests, but hopefully that will change)
 * It builds the documentation (only on 2.7)
 * It deploys the statically generated documentation to Github Pages (only on `develop` branch, but that can be changed). Currently configured to deploy to the [compas-dev/main](https://github.com/compas-dev/main) repo.

Additionally, this means that all pull requests are checked by Travis on all configurations, so if the unit test suite grows, this becomes a great safety net to avoid regressions.

To get this to work, one of the admins (@brgcode ?) needs to:
 * Sign in to Travis
 * Add this repository
 * Generate a personal token with the `public_repo` scope from the github web ui and add it as a Travis variable ([check here for more details](https://docs.travis-ci.com/user/deployment/pages/#setting-the-github-token))

I tested this on a fork of compas and was able to build and push to [some random github page](http://gnz.io/compas-docs/) the generated documentation:

For reference, the settings I used on Travis are the following:
image

I would like to enable the `invoke check` as well once we have this merged, but right now it throws a million warnings about PEP8 violations, unused code and a bunch of other stuff that we should eventually fix but it's not that critical. ;)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,45 @@
+language: python
+
+matrix:
+  include:
+    - python: 2.7
+    - python: 3.6
+
+before_install:
+- if [[ "$TRAVIS_PYTHON_VERSION" == "3.6" ]]; then
+    wget https://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh -O miniconda.sh;
+    bash miniconda.sh -b -p $HOME/miniconda;
+    export PATH="$HOME/miniconda/bin:$PATH";
+    hash -r;
+    conda config --set always_yes yes --set changeps1 no;
+    conda update -q conda;
+    conda info -a;
+    conda create -y -q -n compas_3x -c conda-forge python=3.6;
+    source ~/miniconda/bin/activate compas_3x;
+    pip install -r requirements-dev.txt;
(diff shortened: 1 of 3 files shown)


tomvanmele:
as of now, travis has "third-party access" to compas-dev and is installed as a GitHub app with access to all repositories.

not sure if the token thing is applicable to organisations. at least, there is no obvious way to set it. i can set a token for the brgcode account, but i am not sure that is the way to do this...

also, since we will move the docs back into the compas repo, the built documentation should now be transferred into the corresponding folder of github.io

gonzalocasas:
@brgcode for the time being, a personal token is the way to go because Travis deployer doesn't have the option of using anything else but that. I guess they will add support for deploy keys later on.

tomvanmele:
done, i think :)

screen shot 2018-08-06 at 22 19 13

PR #134, opened 2018-08-19
Fix/homogenize xdraw_mesh API across CAD packages

gonzalocasas:
I hope I’m not missing something obvious, but I’ve been failing to make the "canonical" example of COMPAS work: i.e. mesh drawing on multiple CADs.

A few (older) docs use `mesh_draw`, but that has been mostly removed (only available on `compas_ghpython`), so I tried using `xdraw_mesh`, but Rhino, Blender and Grasshopper packages all 3 have slightly different APIs: `compas_blender` requires a name, `compas_rhino` has a bug in the `xdraw_wrapper` that doesn’t pass `kwargs` properly, and `compas_ghpython` fails if a name is specified.

So, I tried to fix those very basic inconsistencies because we'll do a workshop this Wednesday and need this to work.

Also I added `compas_ghpython` to the docs because it was missing.

### DIFF
--- a/docs/reference.rst
+++ b/docs/reference.rst
@@ -6,5 +6,6 @@ Reference
     :maxdepth: 1
 
     reference/compas
-    reference/compas_rhino
     reference/compas_blender
+    reference/compas_ghpython
+    reference/compas_rhino
--- a/docs/reference/compas_ghpython.rst
+++ b/docs/reference/compas_ghpython.rst
@@ -1,0 +1,2 @@
+
+.. automodule:: compas_ghpython
--- a/src/compas/utilities/names.py
+++ b/src/compas/utilities/names.py
@@ -18,7 +18,7 @@ __all__ = [
 
 
 def random_name(n=17):
-    return ''.join(random.choice(string.lowercase) for _ in range(n))
+    return ''.join(random.choice(string.ascii_lowercase) for _ in range(n))
 
 
 # ==============================================================================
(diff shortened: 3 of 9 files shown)


tomvanmele:
the changes to compas_rhino seem fine. will leave it up to andrew to comment on the blender stuff.

however, just for the record, i want to point out that there is a fundamental difference between `mesh_draw` and `xdraw_mesh`. `mesh_draw` is a helper function for drawing compas meshes and is being deprecated in favour of `MeshArtist`. `xdraw_mesh` is just a convenience wrapper for the Rhino mesh drawing function.

tomvanmele:
i would also prefer that the ghpython package follow the *new-style* conventions for drawing compas meshes using artists. compas_rhino.helpers is only there for backwards compatibility and will be removed as soon as all of its functionality has been integrated in artists, modifiers, inspectors, ...

gonzalocasas:
@brgcode thx. I notice now that this PR breaks some stuff in the artists, so I just fixed that (https://github.com/compas-dev/compas/pull/134/commits/6c9766323336cb3c43faf9501304ad1281d46b48). Until now, I was actually very confused as to what the `*Artists` were. (So I was indeed missing something obvious 😉).

I agree we should switch `compas_ghpython` to this new style of drawing with artists, but if possible, I'd like to merge this fix as is, and work on the artists on a separate PR. I will open an issue for that, because there are a few questions that would need to be answered before tackling that.

PR #145, opened 2018-08-27
Feature/ some changes and updates on the robot model

romanarust:


src/compas_ghpython/helpers/mesh.py
```diff
@@ -75,10 +75,11 @@ def mesh_draw(mesh,
     key_index = {key: index for index, key in enumerate(mesh.vertices())}
     xyz = [mesh.vertex_coordinates(key) for key in mesh.vertices()]
     faces = []
-
     for fkey in mesh.face:
         face = mesh.face_vertices(fkey)  # ordered=True
-        faces.append([key_index[k] for k in face])
+        face = [key_index[k] for k in face]
+        if len(face) == len(set(face)): # sometimes vertex keys are not unique
```
gonzalocasas: I think this was a side effect of the precision issue we saw the other day. Should we revert this bit?

src/compas/robots/model/link.py
```diff
@@ -140,74 +137,65 @@ def __init__(self, name, type=None, visual=[], collision=[], inertial=None, **kw
...
+            cjoint.origin.create(parent_transformation)
+            transformation = Transformation.from_frame(cjoint.origin)
             if cjoint.axis:
                 cjoint.axis.create(transformation)
             clink = cjoint.child_link
-            clink.create(urdf_importer, meshcls, cjoint.origin)
+            clink.create(urdf_importer, meshcls, transformation)
 
-    def update(self, joint_state, parent_transformation, reset_transformation):
+    def update(self, joint_state, parent_transformation, reset_transformation, collision=False):
```
gonzalocasas: Shouldn't `collision` default to `True`? Is there really a case where one wants to transform only the visual geometry and not the collision one?

gonzalocasas (approved):
It generally looks good to me! I added two little questions, once those are clear, I'll merge it. :)

src/compas_ghpython/helpers/mesh.py
romanarust: yes, true, revert please

src/compas/robots/model/link.py
romanarust: you are right!

PR #148, opened 2018-08-28
Use symlinks for Rhino install

gonzalocasas:
Instead of adding the `site-packages` path to Rhino search paths (which causes side problems as observed in the last days), this pull requests switches the install method to symlink `COMPAS` packages to the Rhino lib folder instead.

Python 2.7 needs to be pollyfilled with support using `mklink`, otherwise, python 3.x has native `os.symlink` support on Windows.

I've tested this on Windows 10, with python 2.7 and 3.6. `mklink` is supported since Windows Vista.

### DIFF
--- a/src/compas_rhino/__init__.py
+++ b/src/compas_rhino/__init__.py
@@ -23,11 +23,49 @@ compas_rhino
 """
 from __future__ import absolute_import
 
+import os
+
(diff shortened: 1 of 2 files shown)


tomvanmele:
cool!

PR #159, opened 2018-08-30
Remove appdirs dependency

gonzalocasas:
The functions of the `appdirs` package that were in use have been "vendored", and are now directly included in the framework. I also had to fix the OS detection of the library, because it was not handling IronPython correctly.

Also made a small change to the install scripts to make them reusable from additional packages.


src/compas/_os.py
```diff
@@ -0,0 +1,212 @@
...
+These are internal functions of the framework.
+Not intended to be used outside compas* packages.
+"""
+import os
+import sys
+
+PY3 = sys.version_info[0] == 3
+system = sys.platform
+
+# IronPython support (OMG)
```
tomvanmele, line 12: there is a function for this ``compas.is_ironpython()``

src/compas/_os.py
```diff
@@ -0,0 +1,212 @@
...
+    Returns
+    -------
+    str
+        Full path to the user-specific data dir.
+    """
+    if system == 'win32':
+        if appauthor is None:
+            appauthor = appname
+        const = roaming and "CSIDL_APPDATA" or "CSIDL_LOCAL_APPDATA"
+        path = os.path.normpath(_get_win_folder(const))
```
tomvanmele: is ``const`` supposed to be ``True`` or ``False``?
why the quotes?

gonzalocasas: This part is copy/pasted verbatim from [appdirs](https://github.com/ActiveState/appdirs/blob/master/appdirs.py#L80), but I also found it confusing. If roaming is True, const will be the string `CSIDL_APPDATA`, if not, it will be `CSIDL_LOCAL_APPDATA`

src/compas/_os.py
gonzalocasas: Yes, but that's defined in `compas.__init__` which imports this file, and it would create a circular dependency.

src/compas/_os.py
tomvanmele: ``const = "CSIDL_APPDATA" if roaming else "CSIDL_LOCAL_APPDATA"``

gonzalocasas:
Fixed both findings on https://github.com/compas-dev/compas/pull/159/commits/7791b88150de90c532442e44a7ef447b2fe6cf64

PR #167, opened 2018-09-04
Yield pipes instead of extending lists over and over

gonzalocasas:
Very small performance fix on `xdraw_pipes`.


gonzalocasas:
If possible, please review/merge before 0.3.1 release.

PR #180, opened 2018-09-26
Fix STL binary detection

gonzalocasas:
After trying out a few things, I settled for doing this ascii failing over to binary, which seems to be the more stable approach. It's also similar to the way [numpy-stl](https://pypi.org/project/numpy-stl/) implements it.

Added a few test cases to make sure this works for the 3 different cases I could think of: ascii, binary but with `solid` in the header, binary with purely binary header.


tomvanmele:
cool thanks for fixing

PR #187, opened 2018-10-01
Feature/ghpython artists

romanarust:
Implementation of `compas_ghpython.artists` similar to `compas_rhino.artists`, only without layer and clearing methods.
Inconsistencies are however in `helpers.mesh`. Here I would propose that `mesh.draw` shouldn't have so much drawing options `compas_rhino`: `mesh_draw(mesh, show_faces=True, show_vertices=False, show_edges=False)`, since there is anyway `mesh_draw_vertices`, `mesh_draw_edges`, `mesh_draw_faces`... 
Instead `mesh_draw` should only draw `Rhino.Mesh` similar to the `MeshArtist` with `MeshArtist.draw`, `MeshArtist.draw_vertices`, `MeshArtist.draw_edges`, `MeshArtist.draw_faces`.


tomvanmele:
i would just remove the helpers. in `compas_rhino` this is just a leftover from old code that i have not yet removed because i need to fix a few circular imports first...

tomvanmele:
hi romana,

there seem to be some merge conflicts with the master branch.
can you resolve them?
it seems this is only possible for the author of the pull request...

best,
tom

PR #195, opened 2018-10-31
new/modified mesh class methods

robin-oval:
new
mesh topology: is_empty, euler, genus
polyedge topology: boundaries
mesh geometry: area, centroid, normal

modified
edges_on_boundary with option to get them pointing outwards

### DIFF
--- a/src/compas/datastructures/mesh/mesh.py
+++ b/src/compas/datastructures/mesh/mesh.py
@@ -1385,6 +1385,69 @@ class Mesh(FromToPickle,
             return False
         return not any(4 != len(self.face_vertices(fkey)) for fkey in self.faces())
 
+    def is_empty(self):
+        """Boolean whether the mesh is empty.
+
+        Parameters
+        ----------
+
+        Returns
+        -------
+        bool
+            True if no vertices. False otherwise.
+        """
+
+        if self.number_of_vertices() == 0:
+            return True
+
+        return False
+
+    def euler(self):
+        """Calculate the Euler characterisic.
+
+        Parameters
+        ----------
+
+        Returns
+        -------
+        int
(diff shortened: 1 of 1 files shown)


tomvanmele:
i like these additions very much.
however, i am trying to reduce the complexity and responsibilities of the mesh class on a separate development branch.
in that context, should these functions not be defined in a separate module rather then as part of the mesh class definition?

robin-oval:
I agree that the original mesh module is growing too big. As they can be quite specific, would you suggest a mesh_topology and a mesh_geometry modules with mesh functions instead of methods?

tomvanmele:
if you fetch the branch "reorganise_data_and_geo" you can see what i am working towards.

basically, i am making `compas.geometry`, `compas.topology`, `compas.numerical` completely independent. datastructure-related functionality will be accessible from `compas.datastructures`.  

in `compas.datastructures` the definition of a data structure should become leaner and the algorithmic stuff moved into separate modules.

usage would then be:

```python
from compas.datastructures import mesh_contours
from compas.datastructures import mesh_subdivide
from compas.datastructures import mesh_join
from compas.datastructures import mesh_...
```

robin-oval:
I see. It will be much more intuitive indeed. 
Should I modify my PR according to the structure in the reorganise_data_and_geo branch or do I modify it once this structure is in master?
Or do I do this PR on the reorganise_data_and_geo branch?

tomvanmele:
no, leave it as it is. we can fix later...

robin-oval:
alright

robin-oval:
are these okay for now then?

tomvanmele:
cool, merged, thanks robin

PR #196, opened 2018-10-31
complement network

robin-oval:
https://en.wikipedia.org/wiki/Complement_graph


gonzalocasas (commented):
Nice! I would add a couple of words about what a complement graph is.

robin-oval:
True. Done!

src/compas/topology/duality.py
```diff
@@ -181,6 +184,34 @@ def network_dual(network, cls=None):
 
     return dual
 
+def network_complement(network, cls = None):
```
tomvanmele: remove spaces around "=" please

src/compas/topology/duality.py
```diff
@@ -12,6 +14,7 @@
     'mesh_dual',
     'network_dual',
     'network_find_faces',
+    'network_complement'
```
tomvanmele: i am reworking the structure of the packages and location of algorithms in a separate branch, so this might become an irrelevant question at some point, but just for argument's sake i will ask it anyway, is `duality.py` the right location for this?

robin-oval: I put it here because of the relation with the original graph, which can be compared to the one between a primal and a dual graph. But it's not duality. So I added a new location for complementarity.

src/compas/topology/duality.py
robin-oval: done

src/compas/topology/duality.py
tomvanmele: cool but it seems you now need to resolve a few conflicts with the base branch :)

robin-oval: done!

PR #197, opened 2018-11-02
geometry utilities: weighted_centroid_points; Line.point; Polyline.point

robin-oval:


src/compas/geometry/average.py
```diff
@@ -58,6 +60,25 @@ def centroid_points(points):
...
+def weighted_centroid_points(points, weights):
+    """Compute the weighted centroid of a set of points.
+
+    Parameters
+    ----------
+    points : list
+        A list of point coordinates.
+    weights : list
+        A list of weight floats.
+
```
tomvanmele, line 160: are there no bounds on the weights?

src/compas/geometry/objects/line.py
```diff
@@ -144,6 +144,17 @@ def __iter__(self):
...
 
+    # ==========================================================================
+    # queries
+    # ==========================================================================
+
+    def point(self, t):
+        """Point: The point from the start to the end at a specific normalized parameter."""
+        if t < 0 or t > 1:
+            return None
+        v = self.direction * (t * self.length)
```
tomvanmele, line 163: should we not also check for the bounds and then return one of the points directly if appropriate?

src/compas/geometry/objects/polyline.py
```diff
@@ -128,6 +130,43 @@ def __iter__(self):
...
+    # queries
+    # ==========================================================================
+
+    def point(self, t, snap = False):
+        """Point: The point from the start to the end at a specific normalized parameter.
+        If snap is True, return the closest polyline point."""
+
+        if t < 0 or t > 1:
+            return None
+
```
tomvanmele: check for boundary values and return points immediately if appropriate?

src/compas/geometry/average.py
robin-oval: No they weights can be negative. I added a comment to mention it.

src/compas/geometry/objects/line.py
robin-oval: I imagine one would use line.start or .end directly for the bounds? The method is still valid for the extremities. But I can add something like if you wish.
`if t == 0:
return self.start
elif t==1:
return.end`

src/compas/geometry/objects/polyline.py
robin-oval: Same thing as for the Line class.
I could also create new methods like:
'def start(self):
return self.points[0]
def end(self):
return self.points[-1]'

robin-oval:
should I add the little additions I mention or is it not necessary?

tomvanmele:
what do you mean

src/compas/geometry/average.py
tomvanmele: but the weights are not bound to an interval between 0 and +/- 1?

robin-oval:
returning directly the bounds if t=0 or 1, though not necessary to add

src/compas/geometry/objects/line.py
tomvanmele: would make sense in my opinion

src/compas/geometry/average.py
robin-oval: no the weights have no bounds

src/compas/geometry/objects/line.py
robin-oval: ok

tomvanmele:
after the small updates this is ready to be merged...

robin-oval:
done!

tomvanmele:
we should also start writing tests for all of these things.
will add a template soon...

PR #198, opened 2018-11-02
Feature/mesh offset and thicken + mesh and network disconnected parts and explode

robin-oval:
Offset mesh + thicken mesh. Optional alignment of boundary edges outwards in the method mesh.edges_on_boundary.


robin-oval:
are these additions okay?

tomvanmele:
there seem to be conflicts with the base branch..

src/compas/datastructures/mesh/mesh.py
```diff
@@ -25,8 +25,10 @@
 from compas.geometry import subtract_vectors
 from compas.geometry import normal_polygon
 from compas.geometry import area_polygon
-from compas.geometry import bestfit_plane
-from compas.geometry import flatness
+
```
tomvanmele: are `bestfit_plane` and `flatness` not directly importable from `geometry`?

src/compas/datastructures/mesh/operations/explode.py
```diff
@@ -0,0 +1,90 @@
...
+__email__      = 'oval@arch.ethz.ch'
+
+__all__ = [
+    'mesh_disjointed_parts',
+    'mesh_explode'
+]
+
+def mesh_disjointed_parts(mesh):
+    """Get the disjointed parts in a mesh as lists of faces.
+
```
tomvanmele: what is the difference compared to `compas.topology.connected_components`?

src/compas/datastructures/mesh/operations/join.py
```diff
@@ -4,7 +4,7 @@
 
 from compas.utilities import geometric_key
 
-from compas.geometry import mesh_cull_duplicate_vertices
+from compas.geometry.algorithms.purging import mesh_cull_duplicate_vertices
```
tomvanmele: not importable from `compas.geometry`?

src/compas/datastructures/mesh/operations/explode.py
robin-oval: none... didn't see this
I'll remove the disjointed_parts functions and use connected_components in the explode functions

src/compas/datastructures/mesh/mesh.py
robin-oval: it works when I use the function but the checks fail

src/compas/datastructures/mesh/operations/join.py
robin-oval: it works when I use the function but the checks fail

src/compas/datastructures/mesh/mesh.py
tomvanmele: fixed a few imports. try again, i think it will work now...

src/compas/datastructures/mesh/operations/join.py
tomvanmele: fixed a few imports. try again, i think it will work now...

tomvanmele:
hi robin,

you create circular imports with `offest.py` in `compas.geometry`.
this will be much easier to avoid in the new package structure.

for now, comment out the mesh-specific stuff in the geometry package and all will be fine (and comment out the problematic imports of course). we can fix the problem as soon as a i push the new package structure...

best,
tom

src/compas/datastructures/mesh/operations/explode.py
robin-oval: done

robin-oval:
I commented out the new functions in compas.geometry that create circular imports by using meshes for now.

PR #200, opened 2018-11-06
Feature/join_lines_to_polylines

robin-oval:
Join lines into polylines that stop at points connecting more than two lines.


src/compas/topology/connectivity.py
```diff
@@ -41,6 +45,53 @@ def connectivity_from_edges(edges):
     """"""
     raise NotImplementedError
 
+def join_lines_to_polylines(lines):
+    """Join polylines from lines. The polylines stop at points connectng more than two lines.
```
gonzalocasas: tiny typo
```suggestion
    """Join polylines from lines. The polylines stop at points connecting more than two lines.
```

src/compas/topology/connectivity.py
```diff
@@ -41,6 +45,53 @@ def connectivity_from_edges(edges):
...
+        # delete polyline edges from the list of univisted edges
+        for u, v in pairwise(polyline):
+            if (u, v) in edges_to_visit:
+                edges_to_visit.remove((u, v))
+            elif (v, u) in edges_to_visit:
+                edges_to_visit.remove((v, u))
+
+        polylines.append(polyline)
+
+    return [[network.vertex_coordinates(vkey) for vkey in polyline]for polyline in polylines]
```
gonzalocasas: missing space
```suggestion
    return [[network.vertex_coordinates(vkey) for vkey in polyline] for polyline in polylines]
```

<sub>(sorry, `whitespace_nazi_mode=off`) ;)</sub>

src/compas/topology/connectivity.py
```diff
@@ -41,6 +45,53 @@ def connectivity_from_edges(edges):
...
+    polylines: list
+        The polylines. If the polyline is closed, the two extremities are the same.
+
+    """
+
+    # create graph from line extremities
+    network = Network.from_lines([(line[0], line[-1]) for line in lines])
+
+    polylines = []
+    edges_to_visit = list(network.edges())
```
gonzalocasas: Since you're doing frequent lookups on `edges_to_visit` inside the loop, I suspect it might be slow-ish for networks with lots of edges. Maybe you could change this to use a dictionary instead?

gonzalocasas (commented):
[LGTM](https://www.urbandictionary.com/define.php?term=LGTM) ;)

src/compas/topology/connectivity.py
robin-oval: good point! I switched to a set instead of a list.

gonzalocasas: 👍

robin-oval:
all good?

src/compas/topology/connectivity.py
```diff
@@ -41,6 +46,61 @@ def connectivity_from_edges(edges):
     """"""
     raise NotImplementedError
 
+def join_polylines(polylines, stops = []):
+    """Join polylines. The polylines stop at points connectng more than two lines and to optional additional points.
```
tomvanmele: typo...

src/compas/topology/connectivity.py
```diff
@@ -2,10 +2,15 @@
 from __future__ import absolute_import
 from __future__ import division
 
+from compas.datastructures.network import Network
```
tomvanmele: only second level imports!

src/compas/topology/connectivity.py
tomvanmele: from summary not very clear to me what function does. examples?

robin-oval: I made it a more atomic and intuitive function: join_lines. The lines are connected into a network from which polylines are extracted as sequences of lines that stop at vertices that are not two-valent. Optional splits can be included. I added an example in the documentation.

src/compas/topology/connectivity.py
```diff
@@ -41,10 +46,88 @@ def connectivity_from_edges(edges):
...
+    -------
+    polylines: list
+        The joined polylines. If the polyline is closed, the two extremities are the same.
+
+    Examples
+    --------
+    Joining the lines (a, b), (b, c), (c, d), (c, e) and (e, f), where a ... f are different point coordinates will resut in the following polylines (a, b, c), (c, d) and (c, e, f).
+    
+    .. code-block:: python
+
```
tomvanmele, line 69: need proper indentation here

PR #212, opened 2018-12-02
Update issue templates

tomvanmele:


gonzalocasas (commented):
I think it's good, but I would consider adding a question to understand the "role" or "type of user" requesting the need.

Taking it from "user stories" in Agile, we could even suggest to frame the feature request in the typical user story template of: `As a <role/type of user>, I want <something/some goal> so that <benefit/some reason>.` (see [here](http://www.agilemodeling.com/artifacts/userStory.htm))

tomvanmele:
added it quite literally. we can change it later if it turns out that it doesn't really work...

gonzalocasas:
image

PR #217, opened 2018-12-04
Add additional possibilities to filter mesh's vertices/edges/faces

Achillx:
Added features:
* `vertices_where()`,`edges_where()`,`faces_where()` now will return results even if the `attr` is a list.

* I've also added functions that can accept functions/lambdas as a condition instead of `dict`.
`vertices_where_predicate()`, `edges_where_predicate()`, `faces_where_predicate()`

### DIFF
--- a/src/compas/datastructures/_mixins/filters.py
+++ b/src/compas/datastructures/_mixins/filters.py
@@ -95,8 +95,8 @@ class VertexFilter(object):
 
         Parameters
         ----------
-        predicate : function
-            The condition you want to evaluate. The function takes 2 parameters: ``key``, ``attr`` and should return ``True`` or ``False``.
+        predicate : callable
+            The condition you want to evaluate. The callable takes 2 parameters: ``key``, ``attr`` and should return ``True`` or ``False``.
         data : bool, optional
             Yield the vertices and their data attributes.
             Default is ``False``.
@@ -124,10 +124,10 @@ class VertexFilter(object):
             mesh.add_face([v0,v1,v3])
             mesh.add_face([v1,v2,v3])
 
-            for v_key in mesh.vertices_where_predicate(lambda key, attr: attr['x']==100.0):
+            for v_key in mesh.vertices_where_predicate(lambda key, attr: attr['x'] == 100.0):
                 print v_key
 
-            for v_key, v_data in mesh.vertices_where_predicate(lambda key, attr: 50.0<=attr['y']<=150.0, True):
+            for v_key, v_data in mesh.vertices_where_predicate(lambda key, attr: 50.0 <= attr['y'] <= 150.0, True):
                 print v_key, v_data
 
             for v_key in mesh.vertices_where_predicate(lambda key, attr: 'extra_attr2' in attr):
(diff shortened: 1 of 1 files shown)


Achillx:
@brgcode I wanted to add reviewers, but I do not have permissions.

tomvanmele:
very much like the lambda function version.
a few questions about the list version of the original function.
what happens if the value of the condition is also a list?

src/compas/datastructures/_mixins/filters.py
```diff
@@ -78,6 +90,62 @@ def vertices_where(self, conditions, data=False):
...
+
+            from compas.datastructures import Mesh
+
+            mesh = Mesh()
+
+            v0 = mesh.add_vertex( x = 0.0, y = 0.0, z = 0.0, extra_attr1 = 0, extra_attr2 = [])
+            v1 = mesh.add_vertex( x = 100.0, y = 0.0, z = 0.0, extra_attr1 = 2, extra_attr2 = [3,5,9])
+            v2 = mesh.add_vertex( x = 100.0, y = 100.0, z = 0.0, extra_attr1 = 1, extra_attr2 = ['value1', 'value2'])
+            v3 = mesh.add_vertex( x = 0.0, y = 100.0, z = 0.0, extra_attr1 = 2, extra_attr2 = [3,7,12])
+
```
tomvanmele, line 123: please remove white space between the parenthesis.
only white space after the comma there...

src/compas/datastructures/_mixins/filters.py
```diff
@@ -78,6 +90,62 @@ def vertices_where(self, conditions, data=False):
                 else:
                     yield key
 
+    def vertices_where_predicate(self, predicate, data=False):
+        """Get vertices for which a certain condition or set of conditions is true using a lambda function.
+
+        Parameters
+        ----------
+        predicate : function
+            The condition you want to evaluate.
```
tomvanmele: explain the expected call signature of the lambda function

tomvanmele:
actually, this is the case for all condition types. so maybe we should better document the allowed structure of the condition dict...

Achillx:
> very much like the lambda function version.
> a few questions about the list version of the original function.
> what happens if the value of the condition is also a list?

There are so many possibilities and that is went for the _predicate versions. This way users know what type of data are storing as attr, and they know how to ask for them.

src/compas/datastructures/_mixins/filters.py
Achillx: Done.

src/compas/datastructures/_mixins/filters.py
Achillx: Done, but ``invoke docs`` gives me some errors and I cannot check if it displays correctly.

tomvanmele: use ``callable`` instead of ``function``

src/compas/datastructures/_mixins/filters.py
```diff
@@ -265,7 +265,7 @@ def edges_where_predicate(self, predicate, data=False):
             mesh.set_edge_attributes(random_edge, ['extra_attr1', 'extra_attr2'], [2, [3,5,9]])
 
 
-            for u, v, e_data in mesh.edges_where_predicate(lambda u, v, attr: attr['extra_attr1'] == 2, True):
+            for u, v, e_data in mesh.edges_where_predicate(lambda u, v, attr: attr['extra_attr1']==2, True):
```
tomvanmele: there the whitespace is correct. ony case where there is no whitespace is in the case of function arguments....

src/compas/datastructures/_mixins/filters.py
Achillx: I've change them.

src/compas/datastructures/_mixins/filters.py
Achillx: ok

gonzalocasas:
👯

PR #218, opened 2018-12-05
Remote procedure calls

tomvanmele:
Base implementation and default service...


tomvanmele:
@gonzalocasas what do you think?

tomvanmele:
@tmsmendez @andrewtliew could you look through this? would like to merge...

andrewliew:
Looks exciting, approved!

gonzalocasas:
I'll test it today on windows a little bit, but generally  it looks good!

tmsmendez:
Coolio
.

tmsmendez (approved):
check

andrewliew (approved):

services/default.py
```diff
@@ -0,0 +1,27 @@
...
+
+import compas
+
+from compas.rpc import Server
+from compas.rpc import Service
+from compas.rpc import kill
+from compas.rpc import ping
+
+
+class DefaultService(Service):
```
gonzalocasas, line 14: This default service and default entry point should be part of `src`, probably inside `compas.rpc`. Otherwise, it will not be distributed with the sdist, and even with my local setup, it didn't work because I'm using the symlinked mode to an editable install.

services/default.py
```diff
@@ -0,0 +1,27 @@
...
+    pass
+
+
+if __name__ == '__main__':
+
+    server = Server(("localhost", 8888))
```
gonzalocasas, line 20: The default port of `8888` is not a great idea, I had other stuff running there because it's just too common of a default, and it took me a while to understand why it was not working.

services/test.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+if __name__ == '__main__':
+
+    server = Server(("localhost", 8888))
+
+    server.register_function(ping)
+    server.register_function(kill)
+
+    server.register_instance(DefaultService())
```
gonzalocasas, line 52: This test is not really testing the `TestService`, but in any case, this test service should be more part of the test suite than an outsider folder.

src/compas/rpc/__init__.py
```diff
@@ -0,0 +1,27 @@
...
+********************************************************************************
+compas.rpc
+********************************************************************************
+
+.. currentmodule:: compas.rpc
+
+
+"""
+
+from __future__ import print_function, division, absolute_import
```
gonzalocasas, line 11: omg, I thought you were the champion of import tidiness, but this one does not follow the `force_single_line` directive ;)

src/compas/rpc/proxy.py
```diff
@@ -0,0 +1,276 @@
...
+        self._package = package
+        self._python = python
+        self._url = url
+        self._port = port
+        self._service = service
+        self._process = None
+        self._server = None
+        self._function = None
+        self.profile = None
+        self.stop_server()
```
gonzalocasas, line 111: I think there's a problem here: if I change the python interpreter from `pythonw` to `python` to see the console output, I notice the old server doesn't seem to really stop, and then there are two or more windows handling proxy stuff.

image

PS: The text/log output on both windows was added just to verify

src/compas/rpc/proxy.py
```diff
@@ -0,0 +1,276 @@
...
+
+        Raises
+        ------
+        RPCServerError
+            If the server providing the requested service cannot be reached after
+            100 contact attempts (*pings*).
+
+        """
+        python = self._python
+        script = os.path.join(compas.HOME, 'services', self._service)
```
gonzalocasas, line 128: `compas.HOME` is not a good way to get to this one because it's not the same when the package is installed. And the fact that regardless of what you pass as parameter, it is always joined with `compas.HOME` means you cannot really override it with anything unless is part of the compas distribution.

Instead, I would rely on the fact that compas is installed on the external python, and then run something like `python -m compas.services.run_service` (or similar)

src/compas/rpc/proxy.py
```diff
@@ -0,0 +1,276 @@
...
+        -------
+        The `args` and `kwargs` have to be JSON-serialisable.
+        This means that, currently, only native Python objects are supported.
+        The returned results will also always be in the form of built-in Python objects.
+
+        """
+        idict = {'args': args, 'kwargs': kwargs}
+        istring = json.dumps(idict, cls=DataEncoder)
+
+        # if self.serializer == 'json':
```
gonzalocasas, line 198: Delete commented out code?

src/compas/rpc/server.py
```diff
@@ -0,0 +1,116 @@
...
+
+    Notes
+    -----
+    Should be used together with an instance of `compas.rpc.Server`.
+
+    """
+    return 1
+
+
+# def list_methods_wrapper(dispatcher):
```
gonzalocasas, line 81: Also delete?

gonzalocasas (commented):
I know I'm late to the party, but here's a bit of feedback. To be honest, I could not make it work from Grasshopper. The only time I managed to get it to work was when I replaced `pythonw` with `python` for testing.

services/default.py
tomvanmele: suggestions?

services/test.py
tomvanmele: i just deleted it. was old code...

src/compas/rpc/__init__.py
tomvanmele: point taken :)
fixed!

services/default.py
tomvanmele: done: compas/rpc/services/default.py

src/compas/rpc/proxy.py
tomvanmele: no, still needs to be integrated...

src/compas/rpc/server.py
tomvanmele: no, also still needs to be integrated

src/compas/rpc/proxy.py
tomvanmele: made it local to the package

tomvanmele:
@gonzalocasas could you have a look at the new version...

PR #219, opened 2018-12-11
Dealing with conda environments

gonzalocasas:
This is a first attempt at supporting conda environments for the remote call features, i.e. `RPC Proxy` and `XFunc`. The idea is rather simple: on Rhino install time, we create source code that allows locating the conda environment where everything is installed. I think this is correct since, on install, there's only one environment available, hence the env variables must be correct for the execution to even run.

I've also applied some changes that I suggested as problematic on the `rpc` pull request.

In summary, the changes are:
 - During install, create a `compas_bootstrapper` module inside Rhino lib folder that allows re-locating conda's current environment (including root if none is created)
 - Use `compas_bootstrapper` to start the correct python on `XFunc` and `Proxy`.
 - Move `default` rpc service into `src` and start it using `python -m` syntax.
 - Use stdlib `tempfile` module (less problematic than defaulting to current)
 - Change default rpc port to `8118` which is less popular than `8888`, also, if a port is specified, it is passed to the default service, so that it matches.
 - Default to Rhino 6.0 for all install/uninstall operations


gonzalocasas:
@brgcode I merged your latest changes back into this PR and found a way to get `pythonw` to work. With this PR applied, I can now correctly run `XFunc` and `Proxy` calls into the conda environment where I have installed `compas`. The shutdown also works fine now. 😄  Could you check it out and let me know what you think?

src/compas/rpc/services/default.py
```diff
@@ -15,13 +15,22 @@ class DefaultService(Service):
 
 if __name__ == '__main__':
 
+    import sys
     import threading
```
tomvanmele, line 19: not used anymore

src/compas/rpc/proxy.py
```diff
@@ -11,19 +11,17 @@
 except ImportError:
     from xmlrpc.client import ServerProxy
 
-from subprocess import Popen
+from subprocess import Popen, PIPE, STDOUT
```
tomvanmele, line 14: are we doing anything with these?

src/compas/rpc/proxy.py
```diff
@@ -127,10 +122,10 @@ def start_server(self):
 
         """
         python = self._python
-        script = os.path.join(HERE, 'services', 'default.py')
+        args = [python, '-m', 'compas.rpc.services.default', str(self._port)]
```
tomvanmele, line 125: somehow these get printed to my editor, but i can't find a corresponding print statement...

src/compas/rpc/proxy.py
```diff
@@ -127,10 +122,10 @@ def start_server(self):
...
         address = "{}:{}".format(self._url, self._port)
 
-        self._process = Popen([python, script])
+        self._process = Popen(args, stdout=PIPE, stderr=STDOUT)
```
tomvanmele, line 128: are we doing anything with this PIPE?
should it send live feedback from the process and/or server?
if so, where do we plot it?

tomvanmele (approved):
on line 108:
is it necessary to stop and start the server as part of the initialisation of the proxy object?
should we not try to reuse a server connection if it is available?

also, i tried using it from two separate scripts and use the sticky dict to pass the proxy around, which works fine, and is a better illustration of the point of starting a server...

tomvanmele:
```python
# launch.py

import scriptcontext as sc
from compas.rpc import Proxy

numerical = Proxy('compas.numerical')
sc.sticky['numerical'] = numerical
```
```python
# use.py

import scriptcontext as sc

import compas
from compas_rhino.artists import MeshArtist

numerical = sc.sticky['numerical']
fd_numpy = numerical.fd_numpy

mesh = Mesh.from_obj(compas.get('faces.obj'))

mesh.update_default_vertex_attributes({'px': 0.0, 'py': 0.0, 'pz': 0.0})
mesh.update_default_edge_attributes({'q': 1.0})

key_index = mesh.key_index()

xyz   = mesh.get_vertices_attributes('xyz')
edges = [(key_index[u], key_index[v]) for u, v in mesh.edges()]
fixed = [key_index[key] for key in mesh.vertices_where({'vertex_degree': 2})]
q     = mesh.get_edges_attribute('q', 1.0)
loads = mesh.get_vertices_attributes(('px', 'py', 'pz'), (0.0, 0.0, 0.0))

xyz, q, f, l, r = fd_numpy(xyz, edges, fixed, q, loads)

for key, attr in mesh.vertices(True):
    index = key
    attr['x'] = xyz[index][0]
    attr['y'] = xyz[index][1]
    attr['z'] = xyz[index][2]
    attr['rx'] = r[index][0]
    attr['ry'] = r[index][1]
    attr['rz'] = r[index][2]

for index, (u, v, attr) in enumerate(mesh.edges(True)):
    attr['f'] = f[index][0]
    attr['l'] = l[index][0]

artist = MeshArtist(mesh)
artist.draw_vertices()
artist.draw_edges()
artist.redraw()

src/compas/rpc/proxy.py
gonzalocasas: we don't do anything with the pipes yet, but we could. But after a few different tries, passing `stdout` and `stderr` values to `Popen` was the things that got `pythonw` to work on my setup. Without them, something deep inside `httplib` was throwing a `BadStatusLine` exception.

Eventually, we could do something useful with their content, but right now I can't think of anything generic enough (except maybe just exposing it somewhere via the `Proxy` so that they can be accessed from calling code.

src/compas/rpc/proxy.py
gonzalocasas: yep, see the other comment :)

src/compas/rpc/services/default.py
gonzalocasas: it is used, I pass the port param to the default service, so that launching it matches the port specified on the other side (otherwise one would need to manually edit the default service or create an alternative service matching the port)

src/compas/rpc/proxy.py
gonzalocasas: How is your setup? I tested from Grasshopper to CPython, but I haven't tried other modes. Is the example you commented below the one you use and prints this (from Rhino I assume)?

gonzalocasas:
@brgcode re: stop & start server at init. Yes, that would be totally possible. Should I give it a try in this PR or we merge this and try that in a separate one?

src/compas/rpc/services/default.py
tomvanmele: no, i meant `threading`

src/compas/rpc/proxy.py
tomvanmele: yes, i tried in Rhino using the built-in editor.
didn't always print this, though...

tomvanmele:
merge!

PR #220, opened 2018-12-11
Revert "Remote procedure calls"

tomvanmele:
Reverts compas-dev/compas#218


gonzalocasas:
@brgcode why revert?

tomvanmele:
well let's fix it properly before merging it in, no?

gonzalocasas:
I proposed already several changes on #219, we could revert, but since we haven't released, and this doesn't break existing functionality, I don't think we really need a revert. But, it's up to you. ;)

PR #221, opened 2018-12-11
Feature/signed angle

romanarust:
Added signed_angle function to vectors `angle_vectors_signed(u, v, normal, ..)` : 
Checks if the angle has to be signed such that rotating vector u about the normal by angle would result in a vector that looks into the same direction as v.


gonzalocasas (approved):
LGTM!

tomvanmele:
could you add a test to the test suite?

PR #226, opened 2019-01-10
fix problem of get_point_coordinates

ghost:
Check if the variable guids is iterative so get_point_coordinates(guids) will return the correct coordinates when the guid is not iterable

### DIFF
--- a/src/compas_rhino/utilities/objects.py
+++ b/src/compas_rhino/utilities/objects.py
@@ -226,6 +226,10 @@ def get_points(layer=None):
 
 
 def get_point_coordinates(guids):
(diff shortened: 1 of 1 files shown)


gonzalocasas (changes requested):
It would be more pythonic if we can avoid the type-checking and rely on [duck-typing](https://en.wikipedia.org/wiki/Duck_typing) instead, so, rather than doing this check, we can wrap the `for guid in guids:` loop in a `try..except TypeError:`, and only in that case, assume it's a non-iterable and treat it as single guid.

Performance-wise, this also saves one call for the common case in which the input is actually the expected one.

PR #231, opened 2019-01-16
Speed up STL binary loading time

gonzalocasas:
This PR speeds up the loading time of STL binary files in two ways: 
 * removing the need to calculate `geometry_key`s for each vertex (and instead relying on the raw bytes of the vertex representation)
 * grouping io operations, specifically doing one file read for the full triangle, instead of many small reads (this might have a bigger positive impact on lame OSs with pathetic io performance such as Windows).

On my system (Python 3.6, Windows), a complex mesh loading went down more than 50%, (according to `timeit`, it's it's about 40% of the time it used to take). The first execution is on current master, the second is on this PR:

```
$ python -m timeit -n 100 -r 5 "import compas;from compas.datastructures import Mesh;Mesh.from_stl('link_4.stl')"
100 loops, best of 5: 445 msec per loop

$ python -m timeit -n 100 -r 5 "import compas;from compas.datastructures import Mesh;Mesh.from_stl('link_4.stl')"
100 loops, best of 5: 179 msec per loop
```

In terms of functionality, nothing has changed. I tested the change on a complex mesh (a robot link from one of our ABBs), loading it with compas and rendering it using `compas_rhino`, and it results in a valid mesh that looks proper:
image

### DIFF
--- a/src/compas/datastructures/mesh/_mesh.py
+++ b/src/compas/datastructures/mesh/_mesh.py
@@ -647,10 +647,11 @@ class Mesh(FromToPickle,
 
         """
         mesh = cls()
-        for x, y, z in iter(vertices):
-            mesh.add_vertex(x=x, y=y, z=z)
-        for face in iter(faces):
-            mesh.add_face(face)
+        for vkey, vertex in enumerate(vertices):
+            x, y, z = vertex
+            mesh.add_vertex(key=vkey, x=x, y=y, z=z)
+        for fkey, face in enumerate(faces):
+            mesh.add_face(face, fkey=fkey)
         return mesh
 
     @classmethod
--- a/src/compas/files/stl.py
+++ b/src/compas/files/stl.py
@@ -104,7 +104,7 @@ class STLReader(object):
                 name = None
 
             elif parts[0] == 'facet':
-                facet = {'normal': None, 'vertices': None, 'attributes': None}
+                facet = {'normal': None, 'vertices': None}
                 if parts[1] == 'normal':
                     facet['normal'] = [float(parts[i]) for i in range(2, 5)]
 
@@ -146,10 +146,6 @@ class STLReader(object):
     #
     # ==========================================================================
 
-    def _read_uint8(self):
-        bytes_ = self.file.read(1)
-        return struct.unpack('<B', bytes_)[0]
-
     def _read_uint16(self):
         bytes_ = self.file.read(2)
         return struct.unpack('<H', bytes_)[0]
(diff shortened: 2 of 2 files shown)


src/compas/datastructures/mesh/_mesh.py
```diff
@@ -647,10 +647,11 @@ def from_vertices_and_faces(cls, vertices, faces):
...
-        for x, y, z in iter(vertices):
-            mesh.add_vertex(x=x, y=y, z=z)
-        for face in iter(faces):
-            mesh.add_face(face)
+        for vkey, vertex in enumerate(vertices):
+            x, y, z = vertex
+            mesh.add_vertex(key=vkey, x=x, y=y, z=z)
+        for fkey, face in enumerate(faces):
+            mesh.add_face(face, fkey=fkey)
         return mesh
```
tomvanmele, line 655: nice, that makes a lot of sense

src/compas/files/stl.py
```diff
@@ -146,10 +146,6 @@ def read_solids_ascii(self):
     #
     # ==========================================================================
 
-    def _read_uint8(self):
-        bytes_ = self.file.read(1)
-        return struct.unpack('<B', bytes_)[0]
-
```
tomvanmele, line 152: why delete this?

src/compas/files/stl.py
```diff
@@ -158,10 +154,6 @@ def _read_uint32(self):
         bytes_ = self.file.read(4)
         return struct.unpack('<I', bytes_)[0]
 
-    def _read_float(self):
-        bytes_ = self.file.read(4)
-        return struct.unpack('<f', bytes_)[0]
-
```
tomvanmele, line 164: why delete this?

src/compas/files/stl.py
```diff
@@ -176,32 +168,20 @@ def read_header_binary(self):
...
+        normal = floats_[0:3]
+        vertices = (floats_[3:6], floats_[6:9], floats_[9:12])
+        keys = (bytes_[12:24], bytes_[24:36], bytes_[36:48])
 
-    def read_attributes_binary(self):
-        count = self.read_attribute_byte_count()
-        if count > 0:
-            return self.file.read(count)
-        return None
+        # Skip two-byte attributes, it's not used anywhere (by anyone - on this planet)
```
tomvanmele, line 181: other planets?

tomvanmele (approved):
let's just try it and see what happens

tomvanmele:
should we suggest the potential use of geometric keys as a post-processing step?

PR #234, opened 2019-01-18
Add await_callback function

gonzalocasas:
When using the robotic fabrication package (and many other libraries) that uses callbacks to handle asynchronous code to signal completion, it is very easy to end up in callback hell, and a significant added complexity in written code that needs to deal with this, and while in many places it does make sense to keep code totally asynchronous, there are situations in which the application code just needs to wait for it before proceeding, and not having access to more advanced tooling such as `asyncio` (since it's not available for IronPython) means the code ends up being pretty convoluted.

This PR introduces a little utility function that helps with that case. Basically a poor man's `asyncio` awaiter. 😛 It deals with exceptions thrown in a thread as well and re-arranging callback parameters into meaningful return values.  

> It makes async code great again™️ 

I added several tests and an example, but for completeness, here's once again. Assuming an asynchronous method like: `def request(url, callback):`, one can write the following bit:

```python
from compas.utilities import await_callback

response = await_callback(request, url='https://api.something.ch/')
```

### DIFF
--- a/src/compas/utilities/__init__.py
+++ b/src/compas/utilities/__init__.py
@@ -16,6 +16,16 @@ animation
     gif_from_images
 
 
+async
+=====
+
+.. autosummary::
+    :toctree: generated/
+    :nosignatures:
+
+    await_callback
+
+
 colors
 ======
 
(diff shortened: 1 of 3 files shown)


romanarust:
Great! LGTM!

PR #235, opened 2019-01-20
Add re-planarity to windows

gonzalocasas:
Planarity support for Windows is good to go!

 - Fixed [planarity](https://anaconda.org/conda-forge/planarity) (the C lib) to support Windows
 - Created [python-planarity](https://anaconda.org/conda-forge/python-planarity) conda recipe for the bindings
  - Updated [compas-feedstock](https://github.com/conda-forge/compas-feedstock/pull/16) to re-add planarity

And now this PR updates the `requirements.txt` to re-introduce `planarity` dependency and adds a couple of unit tests for it.

Also removed all the `pip` upgrading, since it seems to be a cause of problems (see this conversation: https://github.com/conda-forge/compas-feedstock/pull/16#discussion_r249262049).

### DIFF
--- a/tests/datastructures/test_network.py
+++ b/tests/datastructures/test_network.py
@@ -1,7 +1,29 @@
+import pytest
+
 import compas
 
 from compas.datastructures import Network
 from compas.datastructures import network_is_planar
+
+
+@pytest.fixture
+def k5_network():
+    network = Network()
+    network.add_edge('a', 'b')
+    network.add_edge('a', 'c')
+    network.add_edge('a', 'd')
+    network.add_edge('a', 'e')
+
+    network.add_edge('b', 'c')
+    network.add_edge('b', 'd')
+    network.add_edge('b', 'e')
+
+    network.add_edge('c', 'd')
+    network.add_edge('c', 'e')
+
+    network.add_edge('d', 'e')
+
+    return network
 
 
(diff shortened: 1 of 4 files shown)


tests/datastructures/test_network.py
```diff
@@ -7,3 +32,12 @@ def test_add_vertex():
     assert network.add_vertex(x=0, y=0, z=0) == 1
     assert network.add_vertex(key=2) == 2
     assert network.add_vertex(key=0, x=1) == 0
+
+
+def test_non_planar(k5_network):
```
gonzalocasas, line 37: @brgcode at first, I used the test in `__main__` of `_planarity` for the unit test, but weirdly enough, that graph returns true for planarity on Python 2.7 and false on 3.x! So I reverted to a canonical planarity test case.

tomvanmele: that makes sense. because i manually add a "crossing" edge between two specific vertices. since the ordering of the vertices is not the same on py27 and py3x this does not necessarily result in a non-planar graph in both cases... would make more sense to just load a non-planar graph instead.

tomvanmele (approved):

tests/datastructures/test_network.py
gonzalocasas: 👍

PR #237, opened 2019-01-22
Rename `Robot` to `RobotModel`

gonzalocasas:
As discussed today, this renames the robot class to `RobotModel`


romanarust:
;) That was fast! Thank you!

PR #242, opened 2019-02-07
Pull Request for handling mesh loading from URDFs

arpastrana:
Proposal for compas.robots.model.robot py and compas.robots.resources.basic.py


.pytest_cache/README.md
```diff
@@ -0,0 +1,8 @@
+# pytest cache directory #
+
+This directory contains data from the pytest's cache plugin,
+which provides the `--lf` and `--ff` options, as well as the `cache` fixture.
+
+**Do not** commit this to version control.
```
gonzalocasas: 😉  please do not commit any of the `.pytest_cache` directory.

src/compas/robots/model/robot.py
```diff
@@ -290,7 +290,7 @@ def load_geometry(self, *resource_loaders, **kwargs):
         force = kwargs.get('force', False)
 
         loaders = list(resource_loaders)
-        loaders.insert(0, DefaultMeshLoader())
+        loaders.append(DefaultMeshLoader())
```
gonzalocasas: I don't think this is correct, the `DefaultMeshLoader` should be inserted at the first position so that it's really the default one, otherwise other loaders will take precedence. Is there a reason why this should change?

src/compas/robots/resources/basic.py
```diff
@@ -68,8 +69,9 @@ def can_load_mesh(self, url):
         #  - a one-letter scheme in Windows
         #  - file scheme
         is_local_file = len(scheme) in (0, 1) or scheme == 'file'
+        is_existing_file = os.path.isfile(url)
```
gonzalocasas: I think this is the wrong place to check this. If `is_local_file` is False, we should not even try to check for file existence so I would suggest moving the check inside the `if` block (I know that now it looks pretty because all vars are assigned in one block but still).

src/compas/robots/resources/basic.py
```diff
@@ -169,6 +171,34 @@ def _get_local_path(self, url):
         return self.build_path(*path.split('/'))
 
 
+class LocalFolderMeshLoader(AbstractMeshLoader):
```
gonzalocasas: I'm not really convinced about this additional loader. It does almost the same thing than the default one, and that can probably be the reason why you had to push the default one to the end of the list. That is likely to cause more confusion down the road because they are really really close in terms of what they do. What would be needed on the default one so that it can handle also this case you have?

gonzalocasas (changes requested):

src/compas/robots/resources/basic.py
arpastrana: I agree on this one, because indeed it is very similar to the default loader. What I am trying to address is the scenario where the `shape.filename` is relative to the directory where the loaded `urdf` file lives. However, I haven't found a way to store it and then retrieve it later on as the `RobotModel` class doesn't store this information when parsed from `urdf`. Could it be that we store the `urdf`directory in `RobotModel.attr` dictionary, and query it afterwards? Although I am aware storing the `urdf` directory is not strictly necesary if the geometry can be processed with the other loaders.

src/compas/robots/resources/basic.py
arpastrana: Something more like this?

```suggestion
is_local_file = len(scheme) in (0, 1) or scheme == 'file'

if is_local_file: 
    if os.path.isfile(url):
        return True
```

.pytest_cache/README.md
arpastrana: Sorry about that! Didn't notice at first!

src/compas/robots/model/robot.py
arpastrana: My understanding is as follows: 
If the user specifies `resource_loaders` (I presume that by default, no resource loader is passed in), why wouldn't the user-defined `resource_loaders` would be processed in first?

tomvanmele: not sure if this exactly applies to the implementation here, but generally speaking things should be processed in the opposite order they inherit. so default first, user-specified later. that way defaults are present and can be overwritten by more specific implementations...

arpastrana: Fair enough. I will undo this part!

src/compas/robots/resources/basic.py
arpastrana: Actually, would passing some `kwargs` to `DefaultMeshLoader` do? Say, for example:

```suggestion
class DefaultMeshLoader(AbstractMeshLoader):
    def __init__(self, **kwargs):
        super(DefaultMeshLoader, self).__init__()
        self.attr = dict() or kwargs

    def can_load_mesh(self, url):
        url = self._get_local_path(url)
        scheme = urlparse(url).scheme
        # continues

    def _get_local_path(self, url):
        filepath = self.attr.get('filepath') 
        if filepath:
            return os.path.join(filepath, url)
        return url
```

tomvanmele:
@gonzalocasas i leave this to you, right?

src/compas/robots/resources/basic.py
```diff
@@ -55,12 +55,34 @@ def load_mesh(self, url):
...
-    def __init__(self):
+    Attributes
+    ----------
+    kwargs (optional): dict
+        Additional keyword arguments.
+    """
+
+    def __init__(self, **kwargs):
         super(DefaultMeshLoader, self).__init__()
+        self.attr = dict() or kwargs
```
gonzalocasas: I think this should be reversed: `kwargs or dict()`. Now it basically always returns `kwargs`, because `dict()` is always `False` (because it's a valid but empty dictionary)

src/compas/robots/resources/basic.py
```diff
@@ -70,18 +92,49 @@ def can_load_mesh(self, url):
...
+        Parameters
+        ----------
+        url : str
+            Mesh location.
+
+        Returns
+        -------
+        url: str
+            Extended mesh location if filepath in kwargs. 
+            Else, ir returns None.
```
gonzalocasas:
```suggestion
            Else, it returns None. 
```

src/compas/robots/resources/basic.py
```diff
@@ -55,12 +55,34 @@ def load_mesh(self, url):
...
+            Mesh URL.
+
+        Returns
+        -------
+        bool
+            ``True`` if the URL points to a local and valid file.
+            Otherwise ``False``.
+        """
+
+        url = self._get_local_path(url) or url
```
gonzalocasas: Minor detail, but I would make `_get_local_path()` return `url` instead of `None`, so that client code is simpler instead of having to know how to select the default.

src/compas/robots/resources/basic.py
```diff
@@ -70,18 +92,49 @@ def can_load_mesh(self, url):
         is_local_file = len(scheme) in (0, 1) or scheme == 'file'
 
         if is_local_file:
-            return True
+            if os.path.isfile(url):
```
gonzalocasas, line 95: This one breaks when the file uses the `file:///` scheme, e.g. if you try to check for 'file:///C:/Users/myuser/some_file_that_does_exist.obj' it will return `False` because it cannot deal with the scheme.

Check the `load_mesh` function, it handles that by stripping it if needed, and I guess you need to do the same thing here.

src/compas/robots/resources/basic.py
```diff
@@ -70,18 +92,49 @@ def can_load_mesh(self, url):
...
+        url : str
+            Mesh location.
+
+        Returns
+        -------
+        url: str
+            Extended mesh location if filepath in kwargs. 
+            Else, ir returns None. 
+        """
+        filepath = self.attr.get('filepath')
```
gonzalocasas: How about `basepath` instead of `filepath` here?

gonzalocasas (changes requested):
It's looking good, please address the comments I just added and then we merge.
(and btw, sorry for the delay!)

src/compas/robots/resources/basic.py
arpastrana: Agree!

src/compas/robots/resources/basic.py
arpastrana: Typo fixed.

src/compas/robots/resources/basic.py
arpastrana: I was actually thinking about what convention to use regarding the `filepath` name. `basepath` sounds good to me, and I will actually use the same variable name to the dictionary entry as well as with the local variable (e.g. `basepath = self.attr.get('basepath')`).

src/compas/robots/resources/basic.py
arpastrana: Agreed.  Now `_get_local_path()` would look like:
```
    def _get_local_path(self, url):
        basepath = self.attr.get('basepath') 
        if basepath:
            return os.path.join(basepath, url)
        return url
```

gonzalocasas: 👍

src/compas/robots/resources/basic.py
arpastrana: Actually, it's great you point this out because I had overlooked to look for the `basepath` when calling `load_mesh()`  in the `DefaultLoader`. However, this makes me think if we should rather hadle the `file:///` stripping part in the `_get_local_path()` method instead of within `load_mesh()`.

Something like: 
```
    def _get_local_path(self, url):
        if url.startswith('file:///'):
            url = url[8:]
        
        basepath = self.attr.get('basepath') 
        if basepath:
            return os.path.join(basepath, url)

        return url
```

Which would imply calling this function _twice_, and changing line 92 from `is_local_file = len(scheme) in (0, 1) or scheme == 'file'` to something like: `is_local_file = len(scheme) in (0, 1)`

What do you think?

gonzalocasas: Good point, let me check how it looks like now

src/compas/robots/resources/basic.py
```diff
@@ -70,17 +92,54 @@ def can_load_mesh(self, url):
...
+        :class:`Mesh`
+            Instance of a mesh.
+        """
+        # if url.startswith('file:///'):
+        #     url = url[8:]
+        
+        url = self._get_local_path(url)
+        return _mesh_import(url, url)
+
+    def _get_local_path(self, url):
```
gonzalocasas: Please rename to `_get_mesh_url` , because at this point, it's not defined whether the input is a local path or a remote url.

src/compas/robots/resources/basic.py
```diff
@@ -70,17 +92,54 @@ def can_load_mesh(self, url):
...
+        ----------
+        url : str
+            Mesh location
+
+        Returns
+        -------
+        :class:`Mesh`
+            Instance of a mesh.
+        """
+        # if url.startswith('file:///'):
```
gonzalocasas: you can get rid of this commented out code now ;)

gonzalocasas (commented):
Looks good! The only relevant comment would be to rename that function but other than that, it's good to go.

If you have the time, and want to dig into it, you could write a unit test for this class. If not, then just make sure you test locally the 4 cases it now supports: 
 - remote files (starting with `http` or `https`) (only for `.obj` files)
 - local files starting with the `file://` scheme
 - local files with absolute path and no scheme (`/Users/gcasas/Desktop/bla.obj`), or the one-letter scheme/variant (i.e. the Windows drive) e.g. `/c/Users/gcasas/Desktop/bla.obj`
 - local files with relative path (added in this PR!)

src/compas/robots/resources/basic.py
```diff
@@ -82,7 +82,7 @@ def can_load_mesh(self, url):
             Otherwise ``False``.
         """
 
-        url = self._get_local_path(url)
+        url = self._get_local_url(url)
```
gonzalocasas: I think the suggested `_get_mesh_url` is more appropriate, because it does not return only local urls, it can be remote as well.

arpastrana: Sorry, I overlooked the suggestion was `_get_mesh_url` instead of `_get_local_url`. I will fix that.

arpastrana:
The only thing I'm missing now is to carry the 4 loading test locally. If something goes wrong, I'll commit the necessary changes.

src/compas/robots/resources/basic.py
arpastrana: So this one we could say is resolved with the latest revisions, right?

arpastrana:
I carried the local tests, and they went well with no further changes 👍 Apart from having loaded the stanford bunny (I couldn't find a quicker alternative for a remote `urfl` with `obj`), I would just add that with the current snippet passing `DefaultMeshLoader` into `robot.load_geometry()` looks a bit _strange_.

[load_urdf.zip](https://github.com/compas-dev/compas/files/2885313/load_urdf.zip)

```
import compas
import os

from compas.robots import RobotModel
from compas.robots import DefaultMeshLoader

file_dir = C:\Users\armandop\Desktop\load_urdf\
file_names = [beam_rel, beam_abs, beam_file, beam_http] 
load = True

if load:
    for name in file_names:
        location = os.path.abspath(os.path.join(file_dir, name + ".urdf"))
        robot = RobotModel.from_urdf_file(location)

        if name == "beam_rel":
            robot.load_geometry(DefaultMeshLoader(basepath=file_dir))  # this one is funny
        else:
            robot.load_geometry()
```

gonzalocasas:
It does, but I think it's ok. The use case for relative files is rather limited, in general, linked files inside urdf use the package scheme and are built around the discoverability mechanisms of ROS so this is seldom needed.

So I'll just go ahead and merge this! Thanks!

PR #243, opened 2019-02-07
Feature/mesh properties

robin-oval:
fix mesh centroid and mesh normal
new face skewness, face aspect ratio, face curvature and vertex curvature
math function average, variance, standard deviation
improvement pairwise and window utilities with wrap option

### DIFF
--- a/src/compas/datastructures/mesh/_mesh.py
+++ b/src/compas/datastructures/mesh/_mesh.py
@@ -10,7 +10,7 @@ from ast import literal_eval
 
 from math import pi
 
-from compas.utilities import avrg
+from compas.utilities import average
 
 from compas.files import OBJ
 from compas.files import PLY
@@ -2514,8 +2514,6 @@ class Mesh(FromToPickle,
 
         """
 
-
-
         return 2 * pi - sum([angle_points(mesh.vertex_coordinates(vkey), mesh.vertex_coordinates(u), mesh.vertex_coordinates(v)) for u, v in pairwise(self.vertex_neighbors(vkey, ordered = True), wrap = True)])
 
     # --------------------------------------------------------------------------
@@ -2711,7 +2709,7 @@ class Mesh(FromToPickle,
 
         max_deviation = max([distance_point_plane(self.vertex_coordinates(vkey), plane) for vkey in self.vertices()])
 
-        average_distances = avrg([distance_point_point(self.vertex_coordinates(vkey), self.face_centroid(fkey)) for vkey in self.vertices()])
+        average_distances = average([distance_point_point(self.vertex_coordinates(vkey), self.face_centroid(fkey)) for vkey in self.vertices()])
 
         return max_deviation / average_distances
 
--- a/src/compas/datastructures/mesh/_mesh.py
+++ b/src/compas/datastructures/mesh/_mesh.py
@@ -2514,7 +2514,7 @@ class Mesh(FromToPickle,
 
         """
 
-        return 2 * pi - sum([angle_points(mesh.vertex_coordinates(vkey), mesh.vertex_coordinates(u), mesh.vertex_coordinates(v)) for u, v in pairwise(self.vertex_neighbors(vkey, ordered = True), wrap = True)])
+        return 2 * pi - sum([angle_points(mesh.vertex_coordinates(vkey), mesh.vertex_coordinates(u), mesh.vertex_coordinates(v)) for u, v in pairwise(self.vertex_neighbors(vkey, ordered = True) + self.vertex_neighbors(vkey, ordered = True)[:1])])
 
     # --------------------------------------------------------------------------
     # edge geometry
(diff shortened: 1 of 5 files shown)


src/compas/utilities/itertools_.py
```diff
@@ -126,18 +126,24 @@ def repeatfunc(func, times=None, *args):
     return starmap(func, repeat(args, times))
 
 
-def pairwise(iterable):
+def pairwise(iterable, wrap=False):
     """s -> (s0,s1), (s1,s2), (s2, s3), ..."""
+    iterable_type = type(iterable)
+    if wrap:
+        iterable = iterable + iterable_type(iterable[:1])
```
gonzalocasas: I think this doesn't work for many cases, e.g. a generator cannot be sliced this way, and it cannot be re-constructed either.

src/compas/utilities/functions.py
```diff
@@ -19,6 +22,58 @@ def fibonacci(n, memo={}):
         memo[n] = fibonacci(n - 2, memo) + fibonacci(n - 1, memo)
     return memo[n]
 
+def avrg(list):
```
gonzalocasas: I would suggest sticking to standard naming for these functions, in particular, taking from `numpy` stats functions, this would be `average`.

src/compas/utilities/functions.py
```diff
@@ -19,6 +22,58 @@ def fibonacci(n, memo={}):
...
+    float
+        The variance value.
+
+    """
+
+    m = avrg(list)
+
+    return (sum([(i - m) ** 2 for i in list]) / float(len(list))) ** .5
+
+def st_dev(list):
```
gonzalocasas: Same as above, I'd stick to numpy naming and call this `std`.

src/compas/utilities/functions.py
```diff
@@ -5,6 +5,9 @@
 
 __all__ = [
     'fibonacci',
+    'avrg',
```
gonzalocasas: Better to move all these new functions into a specialized module `statistics` (and as a personal rant, `utilities` in general is such an annoyingly mixed bag of things I would eventually reconsider its existence as well, but that's beyond the scope of this PR).

src/compas/utilities/functions.py
tomvanmele: agreed, and in any case words should not be abbreviated. what sounds like a meaningful abbreviation to you might not to someone else...

src/compas/utilities/itertools_.py
tomvanmele: also agreed

src/compas/utilities/functions.py
tomvanmele: i would just call this standard_deviation

src/compas/datastructures/mesh/_mesh.py
```diff
@@ -2495,6 +2495,29 @@ def vertex_normal(self, key):
...
+        References
+        ----------
+        .. [1] Botsch, Mario, et al. *Polygon mesh processing.* AK Peters/CRC Press, 2010.
+
+        """
+
+
+
+        return 2 * pi - sum([angle_points(mesh.vertex_coordinates(vkey), mesh.vertex_coordinates(u), mesh.vertex_coordinates(v)) for u, v in pairwise(self.vertex_neighbors(vkey, ordered = True), wrap = True)])
+
```
tomvanmele, line 2518: what's up with the multiple line breaks?!

src/compas/datastructures/mesh/_mesh.py
```diff
@@ -2620,6 +2643,78 @@ def face_flatness(self, fkey):
...
+        """
+
+        plane = bestfit_plane([self.vertex_coordinates(vkey) for vkey in self.vertices()])
+
+        max_deviation = max([distance_point_plane(self.vertex_coordinates(vkey), plane) for vkey in self.vertices()])
+
+        average_distances = avrg([distance_point_point(self.vertex_coordinates(vkey), self.face_centroid(fkey)) for vkey in self.vertices()])
+
+        return max_deviation / average_distances
+
```
tomvanmele, line 2715: this is more of a question than a comment, but should these not be in separate modules? they have little to do with the data structure itself... (i am aware that i have also not always been consistent in this)

src/compas/utilities/itertools_.py
robin-oval: true, I tried to find a not too ugly way to make it work

src/compas/utilities/functions.py
robin-oval: full names now: average, variance, standard_deviation

src/compas/utilities/functions.py
robin-oval: done

src/compas/utilities/itertools_.py
tomvanmele: i think this can just be the responsibility of the user

```python
for u, v in pairwise(vertices + vertices[0:1]):
    print(u, v)
```

robin-oval: ok I removed the wrap option in pairwise and function and I updated the mesh property functions

PR #245, opened 2019-02-08
add meshes_join, mesh_weld, meshes_join_and_weld

robin-oval:
I removed the old meshes_join to have three consistent functions, and I am not sure where to have mesh_unweld_vertices in datastructures.operations.weld.

### DIFF
--- a/src/compas/datastructures/mesh/join.py
+++ b/src/compas/datastructures/mesh/join.py
@@ -4,52 +4,100 @@ from __future__ import division
 
 from copy import deepcopy
 
+from compas.utilities import pairwise
+
 from compas.utilities import geometric_key
-
-from compas.datastructures.mesh.clean import mesh_delete_duplicate_vertices
-
+from compas.utilities import reverse_geometric_key
 
 __all__ = [
+    'mesh_weld',
     'meshes_join',
+    'meshes_join_and_weld'
 ]
 
-
-def meshes_join(meshes, delete_duplicates=False, precision='3f'):
-    """Join multiple meshes. These meshes are usually assumed
-    to be mesh patches that nicely align along their boundaries.
-
+def meshes_join(meshes, cls = None):
+    """Join meshes without welding.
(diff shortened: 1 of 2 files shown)


src/compas/datastructures/mesh/join.py
```diff
@@ -4,57 +4,105 @@
...
-from compas.utilities import geometric_key
-
-from compas.datastructures.mesh.clean import mesh_delete_duplicate_vertices
+from compas.utilities import pairwise
 
+from compas.utilities import geometric_key
+from compas.utilities import reverse_geometric_key
 
 __all__ = [
+    'mesh_weld',
```
tomvanmele, line 13: is this the opposite of `mesh_unweld_vertices`? should it then not be in `weld.py`

src/compas/datastructures/mesh/join.py
```diff
@@ -4,57 +4,105 @@
...
+    # modify vertex indices in the faces
+    faces = [ [vertex_map[geometric_key(mesh.vertex_coordinates(vkey), precision)] for vkey in mesh.face_vertices(fkey)] for fkey in mesh.faces()]
 
-        for fkey,attr in mesh.faces(True):
-            vertices = mesh.face_vertices(fkey)
-            new_vertices = [key_map[key] for key in vertices]
-            mesh_all.add_face(new_vertices)
+    return cls.from_vertices_and_faces(vertices, faces)
+
+def meshes_join_and_weld(meshes, precision = None, cls = None):
```
tomvanmele, line 74: no spaces around `=` in assignment of default values for arguments

tomvanmele: also should `weld` not be an optional flag in `meshes_join` instead of a separate function?

tomvanmele (changes requested):
can you add tests?
not sure how to enforce this, but it should be part of any pull request...

PR #246, opened 2019-02-08
Feature/mesh offset thicken 2

robin-oval:
new mesh_offset and mesh_thicken functions
I do not know why
from compas.datastructures.mesh import mesh_flip_cycles
does not work

### DIFF
--- a/src/compas/datastructures/mesh/join.py
+++ b/src/compas/datastructures/mesh/join.py
@@ -4,52 +4,100 @@ from __future__ import division
 
 from copy import deepcopy
 
+from compas.utilities import pairwise
+
 from compas.utilities import geometric_key
-
-from compas.datastructures.mesh.clean import mesh_delete_duplicate_vertices
-
+from compas.utilities import reverse_geometric_key
 
 __all__ = [
+    'mesh_weld',
     'meshes_join',
+    'meshes_join_and_weld'
 ]
 
-
-def meshes_join(meshes, delete_duplicates=False, precision='3f'):
-    """Join multiple meshes. These meshes are usually assumed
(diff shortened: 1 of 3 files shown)


tomvanmele:
there seems to be quite a bit of overlap with one of your other pull requests. can we split this properly over two and merge them in the right order?

tomvanmele:
also, it is a bit unclear what you mean by `mesh_flip_cycles` does not work, because i just ran it on a few examples and it seems to work just fine...

robin-oval:
sometimes the second level import does not work in a new branch but works in the master branch...

src/compas/datastructures/mesh/join.py
```diff
@@ -4,57 +4,105 @@
 
 from copy import deepcopy
 
-from compas.utilities import geometric_key
-
-from compas.datastructures.mesh.clean import mesh_delete_duplicate_vertices
+from compas.utilities import pairwise
```
gonzalocasas: this doesn't seem to be used.

tomvanmele:
@robin-oval 
> sometimes the second level import does not work in a new branch but works in the master branch...

i was commenting on a different case

```python
# in compas.topology.xxx.py
from compas.datastructures.network import Network
# should be
from compas.datastructures import Network
```
if you are importing from within the same package you should import from the same level as the module you are importing into...
```python
# in compas.datastructures.mesh.geodesics.py
from compas.datastructures import trimesh_cotangent_laplacian_matrix
# should be
from compas.datastructures.mesh.matrices import trimesh_cotangent_laplacian_matrix
```

robin-oval:
actually this PR calls a function included in another PR

should I remove the older PR or wait for it to be merged?

PR #247, opened 2019-02-08
uncomment join_lines

robin-oval:


src/compas/topology/connectivity.py
```diff
@@ -5,9 +5,11 @@
 from compas.utilities import pairwise
 from compas.utilities import geometric_key
 
+from compas.datastructures.network import Network
```
tomvanmele: topology can't import from datastructures.
maybe this works now, but it will certainly produce errors later

tomvanmele: also don't import from beyond the second level => `from compas.datastructures import Network`

tomvanmele (commented):
i find it problematic that a function in topology imports from datastructures. this will create problems down the road...

tomvanmele (changes requested):
i find it problematic that a function in topology imports from datastructures. this will create problems down the road...

robin-oval:
that's true I forgot
I converted it into network_polylines

src/compas/datastructures/network/operations/join.py
```diff
@@ -42,10 +45,88 @@ def network_join_edges(network, key):
     # set attributes based on average of two joining edges?
     network.add_edge(a, b)
 
+def network_polylines(network, splits = []):
```
tomvanmele: no space around `=` in assignment of default values of function arguments

tomvanmele: also, using a mutable object as default value can have side effects. better to use `None`

tomvanmele (changes requested):
other than the minor changes, i approve...

src/compas/datastructures/network/operations/join.py
```diff
@@ -54,8 +54,8 @@ def network_polylines(network, splits = []):
     ----------
     network : Network
         A network.
-    splits : list
-        List of point coordinates for optional splits.
+    splits : list, None
```
tomvanmele: proper way to write this would be
```python
splits : list, optional
    List of point coordinates for optional splits.
    Default is ``None``
```

tomvanmele (approved):
approved

PR #249, opened 2019-02-08
fix exception in angle_vectors_signed if vectors aligned

robin-oval:


src/compas/geometry/angles.py
```diff
@@ -92,10 +92,15 @@ def angle_vectors_signed(u, v, normal, deg=False, threshold=1e-3):
...
     normal_uv = cross_vectors(u, v)
-    # check if normal_uv has the same direction as normal
-    angle_btw_normals = angle_vectors(normal, normal_uv)
-    if angle_btw_normals > threshold:
-        angle *= -1
+
+    # exception if u and v aligned and no normal exists
+    if normal_uv == [0., 0., 0.]:
+        pass
+    else:
```
tomvanmele: should you not check for the length of the normal and compare it to an epsylon value? and perhaps

```python
if length_vector(normal_uv) > eps:
    # do stuff
```
rather than `pass` and then `else`

tomvanmele (changes requested):

src/compas/geometry/angles.py
robin-oval: done

tomvanmele (approved):

PR #250, opened 2019-02-12
binomial coefficient function

robin-oval:


tomvanmele:
there is already a function like this in `compas.geometry.Bezier`. could you check how they relate?

robin-oval:
true! I moved that one from compas.geometry._primitives.curves to compas.utilities.functions and fixed the imports

tomvanmele (changes requested):
can you write a test for this one? should be a fairly easy one...

tests/utilities/test_functions.py
```diff
@@ -0,0 +1,8 @@
+import functools
+
+from compas.utilities import binomial
+
+
+def test_binomial():
+    assert binomial(3, 1) == 3
+    assert binomial(21, 10) == 352716
```
tomvanmele: can you make sure to test the boundary cases?
for example, what if one or both of the number is zero?
see the geometry tests for setting up tests that are expected to fail.
also you can parametrize the input so the two assertions count as two tests...

tests/utilities/test_functions.py
```diff
@@ -1,8 +1,19 @@
...
+	[
+	pytest.param(3, 1, 3),
+	pytest.param(21, 10, 352716),
+	pytest.param(10, 0, 1),
+	pytest.param(0, 0, 1),
+	pytest.param(1, 2, 0, marks=pytest.mark.xfail(raises=ValueError)),
+	pytest.param(2, -1, 0, marks=pytest.mark.xfail(raises=ValueError)),
+	]
+)
+
```
tomvanmele: no spacing here, because the stuff above is a decorator for the function and should be "attached" to it...

src/compas/utilities/functions.py
```diff
@@ -19,10 +21,37 @@ def fibonacci(n, memo={}):
         memo[n] = fibonacci(n - 2, memo) + fibonacci(n - 1, memo)
     return memo[n]
 
+def binomial(n, k):
+    """Returns the binomial coefficient of the :math:`x^k` term in the
+    polynomial expansion of the binmoial power :math:`(1 + x)^n` [wikipedia2017j]_.
```
gonzalocasas:
```suggestion
    polynomial expansion of the binomial power :math:`(1 + x)^n` [wikipedia2017j]_.
```

src/compas/utilities/__init__.py
```diff
@@ -72,6 +72,7 @@
     :nosignatures:
 
     fibonacci
+    binomial_coefficient
```
gonzalocasas, line 75: Shouldn't this be called `binomial` or the function renamed to `binomial_coefficient`? Or are these two different things?

robin-oval: I think binomial_coefficient is the correct name.

gonzalocasas: I agree, but what I meant is, you called the function `binomial` but referenced in the docs as `binomial_coefficient`, so doc will fail to find it. Unless, there's a second function called `binomial_coefficient` that I'm not seeing...

gonzalocasas: see the warning in travis: https://travis-ci.com/compas-dev/compas/jobs/177444951#L658

robin-oval: ah yes!

robin-oval: I am now using binomial_coefficient everywhere

tomvanmele (approved):

PR #251, opened 2019-02-12
extension data structure from vertices and edges/faces

robin-oval:
the keys of the elements in Mesh.from_vertics_and_faces() and Network.from_vertices_and_edges() can be speficied using a dictionary instead of a list
I added an example for in the main

### DIFF
--- a/src/compas/datastructures/mesh/_mesh.py
+++ b/src/compas/datastructures/mesh/_mesh.py
@@ -619,11 +619,10 @@ class Mesh(FromToPickle,
 
         Parameters
         ----------
-        vertices : list
-            A list of vertices, represented by their XYZ coordinates.
-        faces : list
-            A list of faces.
-            Each face is a list of indices referencing the list of vertex coordinates.
+        vertices : list, dict
+            A list of vertices, represented by their XYZ coordinates, or a dictionary of vertex keys pointing to their XYZ coordinates.
+        faces : list, dict
+            A list of faces, represented by a list of indices referencing the list of vertex coordinates, or a dictionary of face keys pointing to a list of indices referencing the list of vertex coordinates.
 
         Returns
         -------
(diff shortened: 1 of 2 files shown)


src/compas/datastructures/mesh/_mesh.py
```diff
@@ -644,11 +643,21 @@ def from_vertices_and_faces(cls, vertices, faces):
 
         """
         mesh = cls()
-        for vkey, vertex in enumerate(vertices):
-            x, y, z = vertex
-            mesh.add_vertex(key=vkey, x=x, y=y, z=z)
-        for fkey, face in enumerate(faces):
-            mesh.add_face(face, fkey=fkey)
+
+        if type(vertices) == list:
```
tomvanmele: technically, the preferred way of doing this kind of thing would be through *duck typing*. if you are going to check for specific types, the method you used will not always work. something like the following would be better

```python
import collections

# ...

if isinstance(vertices, collections.Mapping):
    # stuff

**Review by tomvanmele — CHANGES_REQUESTED** [review]
type checking as you implemented it, is not correct...

File: src/compas/datastructures/mesh/_mesh.py (hunk 1 above)
Comment (tomvanmele) [reply src/compas/datastructures/mesh/_mesh.py]: https://docs.python.org/3/library/collections.abc.html

Comment (robin-oval) [reply src/compas/datastructures/mesh/_mesh.py]: great, I used collections.Mapping for dictionaries and collections.MutableSequence for lists

File: src/compas/datastructures/mesh/_mesh.py [hunk 2]
```diff
@@ -644,17 +645,17 @@ def from_vertices_and_faces(cls, vertices, faces):
         """
         mesh = cls()
 
-        if type(vertices) == list:
+        if isinstance(vertices, collections.MutableSequence):
```
tomvanmele: can it not just be `collections.Sequence`?
doesn't have to be a list, can also be a tuple...

src/compas/datastructures/mesh/_mesh.py
```diff
@@ -644,17 +645,17 @@ def from_vertices_and_faces(cls, vertices, faces):
...
             for x, y, z in iter(vertices):
                 mesh.add_vertex(x=x, y=y, z=z)
-        elif type(vertices) == dict:
+        elif isinstance(vertices, collections.Mapping):
             for key, xyz in vertices.items():
                 mesh.add_vertex(key = key, attr_dict = {i: j for i, j in zip(['x', 'y', 'z'], xyz)})
         
-        if type(faces) == list:
+        if isinstance(faces, collections.MutableSequence):
```
tomvanmele: can it not just be `collections.Sequence`?
doesn't have to be a list, can also be a tuple...

tomvanmele (changes requested):

src/compas/datastructures/mesh/_mesh.py
gonzalocasas: I know this is merged already so I'm late to the party, but I think this change rules out iterators, which is a real pity. As suggested by @brgcode , I think best would be *duck typing* style, but if you want to keep the type check, please try with `collections.Iterable` and add a bunch of simple tests that make sure this works with the basic stuff: lists, tuples, iterators/generators, dicts, and ideally also a custom iterable class.

tomvanmele: i already changed this locally, will push it soon. i think the check for mapping is useful, but that should it. i already had problems with numpy arrays for example.

PR #253, opened 2019-02-13
fix mesh.face_adjacency_halfedgeS

robin-oval:
check for multiple halfedges, occurring due to two-valent singularities as in [a, b, e, d] and [b, c, d, e]
/|\
\|/


tomvanmele:
not sure i get the point.
there can only be one pair of half-edges separating two faces.
that is the essence of a half-edge data structure...

PR #257, opened 2019-02-18
some analytical curves

robin-oval:
evaluation of parametric curves: circle, ellipse, archimedean spiral, logarithmic spiral and helix

### DIFF
--- a/src/compas/geometry/analytical.py
+++ b/src/compas/geometry/analytical.py
@@ -10,7 +10,7 @@ __all__ = [
 	'helix_evaluate'
 ]
 
-def circle_evaluate(t, r, z):
+def circle_evaluate(t, r, z=0):
 	"""Evalutes a circle at a parameter.
 
 	Parameters
@@ -19,8 +19,8 @@ def circle_evaluate(t, r, z):
 		Parameter
 	r : float
 		Constant
-	z : float
-		Height
+	z : float, optional
+		Height. Default is 0.
 
 	Notes
 	-----
@@ -34,7 +34,7 @@ def circle_evaluate(t, r, z):
 	"""
 	return (r * cos(t), r * sin(t), z)
 
-def ellipse_evaluate(t, r, z):
+def ellipse_evaluate(t, r, z=0):
 	"""Evalutes an ellipse at a parameter.
 
 	Parameters
@@ -45,8 +45,8 @@ def ellipse_evaluate(t, r, z):
 		Constant
 	b : float
 		Constant
-	z : float
-		Height
+	z : float, optional
+		Height. Default is 0.
 
 	Notes
 	-----
@@ -60,7 +60,7 @@ def ellipse_evaluate(t, r, z):
 	"""
 	return (a * cos(t), b * sin(t), z)
 
-def archimedean_spiral_evaluate(t, a, b):
+def archimedean_spiral_evaluate(t, a, b, z=0):
 	"""Evalutes a spiral at a parameter. The analytical polar equation is r = a + b * theta.
 
 	Parameters
@@ -71,11 +71,13 @@ def archimedean_spiral_evaluate(t, a, b):
 		Constant
 	b : float
 		Constant
+	z : float, optional
+		Height. Default is 0.
 
 	Returns
 	-------
 	tuple
-		The (x, y) coordinates.
+		The (x, y, z) coordinates.
 
 	Notes
 	-----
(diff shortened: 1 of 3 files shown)


src/compas/geometry/analytical.py
```diff
@@ -0,0 +1,159 @@
...
+	The radius is r.
+
+	Returns
+	-------
+	tuple
+		The (x, y, z) coordinates.
+
+	"""
+	return (r * cos(t), r * sin(t), z)
+
```
tomvanmele, line 38: two blank lines between function definitions

src/compas/geometry/analytical.py
```diff
@@ -0,0 +1,159 @@
...
+	The radius along the x and y axes are a and b, respectively.
+
+	Returns
+	-------
+	tuple
+		The (x, y, z) coordinates.
+
+	"""
+	return (a * cos(t), b * sin(t), z)
+
```
tomvanmele, line 65: two blank lines

src/compas/geometry/analytical.py
```diff
@@ -0,0 +1,159 @@
...
+	The angle of the tangent at the beginning is equal to a.
+	The length of an arc of the spiral from 0 t is equal to L(t) = b / 2 * t ** 2.
+
+	References
+	----------
+	.. [1] GeoGebra. *Archimedean Spiral built by parametric equations*.
+		   Available at: https://www.geogebra.org/m/dZuH5hWa.
+	"""
+	return (b * t * cos(t + a), b * t * sin(t + a), z)
+
```
tomvanmele, line 98: two blank lines

src/compas/geometry/analytical.py
```diff
@@ -0,0 +1,159 @@
...
+	tuple
+		The (x, y, z) coordinates.
+
+	References
+	----------
+	.. [1] GeoGebra. *An equiangular spiral - parametric equation*.
+		   Available at: https://www.geogebra.org/m/zsHgCvq7.
+	"""
+	return (a * exp(b * t) * cos(t), a * exp(b * t) * sin(t), z)
+
```
tomvanmele, line 125: two blank lines

src/compas/geometry/analytical.py
```diff
@@ -0,0 +1,159 @@
...
+	a : float
+		Constant
+	b : float
+		Constant
+	c : float
+		Constant
+
+	Returns
+	-------
+	tuple
```
tomvanmele: i would return a list. you never know where the output will be used and some functions will fail if given a tuple because they will try modify it...

src/compas/geometry/analytical.py
```diff
@@ -0,0 +1,159 @@
...
+		The (x, y, z) coordinates.
+
+	Notes
+	-----
+	An interpreatation of the constants a and b are:
+		- the radius of the helix is a, and
```
tomvanmele: no indentation. otherwise your list will be interpreted as a block quote

src/compas/geometry/analytical.py
```diff
@@ -0,0 +1,159 @@
...
+	a : float
+		Constant
+	b : float
+		Constant
+	z : float, optional
+		Height. Default is 0.
+
+	Returns
+	-------
+	tuple
```
tomvanmele: return a list instead of a tuple

src/compas/geometry/analytical.py
```diff
@@ -0,0 +1,159 @@
...
```
tomvanmele: return a list instead of a tuple

src/compas/geometry/analytical.py
```diff
@@ -0,0 +1,159 @@
...
+	z : float, optional
+		Height. Default is 0.
+
+	Notes
+	-----
+	The radius along the x and y axes are a and b, respectively.
+
+	Returns
+	-------
+	tuple
```
tomvanmele: return a list instead of a tuple

src/compas/geometry/analytical.py
```diff
@@ -0,0 +1,159 @@
...
+	z : float, optional
+		Height. Default is 0.
+
+	Notes
+	-----
+	The radius is r.
+
+	Returns
+	-------
+	tuple
```
tomvanmele: return a list instead of a tuple

tomvanmele (changes requested):
please also add tests. should be very simple in this case...

tomvanmele (approved):

PR #259, opened 2019-02-19
face_circle

robin-oval:
face circle for triangular faces, none otherwise

possible improvement: bestfit circle
not included because there's only a numpy bestfit circle

### DIFF
--- a/src/compas/datastructures/mesh/_mesh.py
+++ b/src/compas/datastructures/mesh/_mesh.py
@@ -2632,33 +2632,7 @@ class Mesh(FromToPickle,
 
         """
         return area_polygon(self.face_coordinates(fkey))
-
-    def face_circle(self, fkey):
-        """Get data on circumcentre of triangular face.
-
-        Parameters
-        ----------
-        fkey : Key
-            The face key.
-
-        Returns
-        -------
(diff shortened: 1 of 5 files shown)


src/compas/datastructures/mesh/_mesh.py
```diff
@@ -2633,6 +2633,32 @@ def face_area(self, fkey):
         """
         return area_polygon(self.face_coordinates(fkey))
 
+    def face_circle(self, fkey):
```
tomvanmele: not sure if it is advisable to add trimesh specific functions to the general mesh implementation. perhaps we could make this the rule: if N-gon if can be added directly, if trimesh/quadmesh specific => add as separate function...

tomvanmele (commented):
not sure if the function should be added to the mesh implementation...

robin-oval:
should I add it into a new file like compas.datastructures.mesh.geometry.py as trimesh_face_circle? along with mesh_face_skewness, mesh_face_aspect_ratio, mesh_vertex_curvature...

robin-oval:
I created a mesh.geometry file and made a trimesh_face_circle function.

PR #260, opened 2019-02-19
class method mesh.from_polylines

robin-oval:
creates a mesh from polylines, each polyline becoming an edge

based on mesh from lines but different because the "curved" polylines can come to a vertex with a different angle and therefore a different order than lines


src/compas/datastructures/mesh/_mesh.py
```diff
@@ -614,6 +614,46 @@ def from_lines(cls, lines, delete_boundary_face=False, precision=None):
...
+        corner_vertices = [geometric_key(xyz) for polyline in boundary_polylines + other_polylines for xyz in [polyline[0], polyline[-1]]]
+        boundary_vertices = [geometric_key(xyz) for polyline in boundary_polylines for xyz in polyline]
+
+        mesh = cls.from_lines([(u, v) for polyline in boundary_polylines + other_polylines for u, v in pairwise(polyline)])
+
+        # remove the vertices that are not from the polyline extremities and the faces with all their vertices on the boundary
+        vertex_keys = [vkey for vkey in mesh.vertices() if geometric_key(mesh.vertex_coordinates(vkey)) in corner_vertices]
+        vertex_map = {vkey: i for i, vkey in enumerate(vertex_keys)}
+        vertices = [mesh.vertex_coordinates(vkey) for vkey in vertex_keys]
+        faces = [[vertex_map[vkey] for vkey in mesh.face_vertices(fkey) if geometric_key(mesh.vertex_coordinates(vkey)) in corner_vertices] for fkey in mesh.faces() if sum([geometric_key(mesh.vertex_coordinates(vkey)) not in boundary_vertices for vkey in mesh.face_vertices(fkey)])]
```
tomvanmele: please write this out. it is unreadable and it will not produce a (noticeable) speed improvement...

tomvanmele (commented):
tests?

src/compas/datastructures/mesh/_mesh.py
robin-oval: I split it into more reasonable pieces

robin-oval:
I added a test

tomvanmele:
still don't really get it, but let's just merge and see what happens :)

PR #261, opened 2019-02-26
Update in plane.transform() to account for nested list

arpastrana:


src/compas/geometry/_primitives/plane.py
```diff
@@ -243,8 +243,8 @@ def transform(self, matrix):
             The transformation matrix.
 
         """
-        point = transform_points([self.point], matrix)
-        normal = transform_vectors([self.normal], matrix)
+        point = transform_points([self.point], matrix)[0]
```
gonzalocasas, line 246: ~Was this a bug? Does it fail without this change?~ Maybe you could add a little unit test for it ~(especially if it was failing before and your change fixes it)~..

gonzalocasas: sorry, I just see the linked issue ;)

arpastrana:
Here's a small unit-test. Without the try-except, it fails with `TypeError: float() argument must be a string or a number, not 'list'`. Moreover, I noticed the slicing proposed here in #261 is already implemented in line 554 of  `_primitives/vector.py`  

```
import math
from compas.geometry import Plane
from compas.geometry import Vector

vector = Vector(0, 0, 1)
plane = Plane([0, 0, 0], vector)
frame = Frame.worldXY()

for angle in range(0, 90, 15):
	print('---')
	rot = Rotation.from_axis_and_angle([1, 0, 0], math.radians(angle))
	
	vector.transform(rot)
	print(vector)

	frame.transform(rot)
	print(frame)

	try:
		plane.transform(rot)
		print(plane)
	except TypeError:
		print('something went wrong!')
```

romanarust:
There is actually an incosistency in the `plane.transform()` in comparison to transform functions in other primitives. They all perform the Transform through their attributes. So rather than `point = transform_points([self.point], matrix)[0]` and then overwriting, it should be `self.point.transform(transformation)` and `self.vector.transform(transformation)` (passing an `Transformation` class). I am actually about to fix this inconsistency amongst other stuff in the branch *feature/geometry_primitives_update*, which is soon to merge. But in the meantime I'll merge this one.

https://github.com/compas-dev/compas/blob/feature/geometry_primitives_update/src/compas/geometry/_primitives/plane.py

PR #267, opened 2019-03-14
Make installer a bit more resilient

gonzalocasas:
- Fixed detection issue that caused broken symlinks to not be removable (e.g. when the target was deleted but link stays in place)
- Always retry with `create_symlink` polyfill: Interestingly enough, the polyfill does NOT require admin rights, so, we first try with the native function if available, but if that fails (e.g. due to lack of admin rights), we retry with the polyfill.
- Fixed an exception on Py2k: `FileNotFoundError` is not defined on python 2, so the installer was broken.

### DIFF
--- a/src/compas/_os.py
+++ b/src/compas/_os.py
@@ -66,8 +66,10 @@ def select_python(python_executable):
 def absjoin(*parts):
     return os.path.abspath(os.path.join(*parts))
 
+
 # Cache whatever symlink function works (native or polyfill)
 _os_symlink = None
+
 
 def create_symlink_polyfill():
     def symlink_ms(source, link_name):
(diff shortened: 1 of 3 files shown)


gonzalocasas:
I tested this on Windows with Python 2.x and 3.x. @brgcode could you pls test on Mac?

tomvanmele (approved):
there are some blank lines missing in `_os.py`, but other than that everything seems to work fine...

gonzalocasas:
> there are some blank lines missing in `_os.py`, but other than that everything seems to work fine...

shame on me! fixed on https://github.com/compas-dev/compas/pull/267/commits/d27f07cf63a3b2036285825930c484fc8bfe3c86

tomvanmele:
will you merge?

PR #270, opened 2019-03-25
Fix xfunc and rpc environment activation

gonzalocasas:
Conda environments need to add some things to the `PATH` when activated, at least on Windows. This fixes it for `XFunc` and `RPC` calls ([report on Forum](https://forum.compas-framework.org/t/rhino-6-0-error-no-module-name-numpy/156/9)).

### Checklist

1. [x] Add the change to the `CHANGELOG.md` file.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Fix `XFunc` and `RPC` environment activation.
 - Fix exception on Rhino Mac.
 - Fix missing import on `compas_rhino.geometry`.
 
--- a/src/compas/_os.py
+++ b/src/compas/_os.py
@@ -6,6 +6,11 @@ Not intended to be used outside compas* packages.
 import os
 import sys
 
+try:
+    NotADirectoryError
+except NameError:
+    class NotADirectoryError(Exception):
+        pass
 
 PY3 = sys.version_info[0] == 3
 system = sys.platform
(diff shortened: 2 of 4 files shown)


gonzalocasas:
I think this is required to be released before Wednesday's `compas_fea` workshop, because some participants already have the requirements installed in an conda environment.

tomvanmele (commented):
* `NotADirectoryError` does not exist in Python 2.7 (at least not on Mac).
* Tried a fresh `rpc` environment and RhinoMac seems to be able to find the correct Python
* For now on Mac (because of Matplotlib) the command to install and uninstall should be `pythonw -m compas_rhino.xxx`
* it seems that on mac the equivalent of `PYTHON_DIRECTORY, 'Library', 'bin'` would have to be `PYTHON_DIRECTORY, 'lib'`
* To add environment variables, the equivalent using the Windows `Process.StartInfo` would be `self._process.StartInfo.EnvironmentVariables = env`

gonzalocasas:
> `NotADirectoryError` does not exist in Python 2.7 (at least not on Mac).

where do you see that one? did i add it now?

> For now on Mac (because of Matplotlib) the command to install and uninstall should be `pythonw -m compas_rhino.xxx`

Before changing that, I would ask why does matplotlib play a role when installing for rhino? Do you have one import too many somewhere?

> it seems that on mac the equivalent of `PYTHON_DIRECTORY, 'Library', 'bin'` would have to be `PYTHON_DIRECTORY, 'lib'`

thx, I wasn't sure and I didn't have a mac to test, I'll add it.

> To add environment variables, the equivalent using the Windows `Process.StartInfo` would be `self._process.StartInfo.EnvironmentVariables = env`

thx! will add that

tomvanmele:
you are right about the plotters. they should not have an effect on this process.

andrew's `devo` stuff imports `matplotlib` for some implicit plotting which should be handled through a callback. will ask him to change that...

the `NotADirectoryError` is part of the `remove_symlink` function. apparently i had never run in to that problem in a py2-based environment before...

PR #273, opened 2019-03-26
fix in mesh weld operations to fix face vertices like [a,b, c, c] to [a, b, c]

robin-oval:
fix in mesh weld operations to fix face vertices like [a,b, c, c] to [a, b, c]


gonzalocasas:
@robin-oval lgtm! could you please add a line to the `CHANGELOG.md` file with your fix?

robin-oval:
`CHANGELOG.md` edited! has to be done for each PR?

gonzalocasas:
@robin-oval ideally, yep, otherwise we need to go through all changes before releasing a new version if we want to provide a meaningful changelog.

CHANGELOG.md
```diff
@@ -16,6 +16,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Removed
 
+## [0.5.1] 2019-03-26
+
+### Changed
```
gonzalocasas: This looks strange... you should not add a new release, you need to enter your changes in the top-most (`Unreleased`) section. But I don't understand why you have here 0.5.1 with a wrong date... did you have a merge conflict and manually resolved?

robin-oval: no I just wrote it down like this but I'll correct it

gonzalocasas: ah ok! the idea is that you're basically queuing up your change in there, since you don't know on which version it will be released (because it's not defined whether next release will be a patch, minor or major one).

btw, have you seen the "pull request template" we added just recently? It's only a small template for the description of the PR on github, but I'd be great to get your feedback about it since you are the author of plenty of PRs.

robin-oval: ok thanks!

I think the template helps, though I skipped the part on the `CHANGELOG.md` file...
Maybe having the template including the Added/Changed/Removed of the `CHANGELOG.md` file would make sense.

gonzalocasas: thx! just updated the template to mention the headings: https://github.com/compas-dev/compas/blob/master/.github/pull_request_template.md

about this PR (and sorry for nitpicking!), the line is now good, but it's still under an already released version, could you move it from line 34 all the way up to be under the `Unreleased` section heading, so, around line 13 now? thx!

robin-oval: ah right I was editing an old version of the `CHANGELOG.md... done!

gonzalocasas: thx!

PR #275, opened 2019-04-01
Fix/surface mesh map

robin-oval:
Fix `mesh_uv_to_xyz`in `RhinoSurface`. Bug fix in a **backwards-compatible** manner.


tomvanmele:
can you resolve the conflicts?

robin-oval:
conflict resolved

src/compas_rhino/geometry/surface.py
```diff
@@ -430,7 +430,7 @@ def mesh_uv_to_xyz(self, mesh, cls=None):
             cls = type(mesh)
 
         vertices, faces = mesh.to_vertices_and_faces()
-        vertices = [self.point_uv_to_xyz(uv0[:2]) for uv0 in vertices]
+        vertices = {vkey: self.point_uv_to_xyz(uv0[:2]) for vkey, uv0 in vertices.items()}
```
tomvanmele: what is `uv0`? (or why the zero)

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Fix `mesh_uv_to_xyz`in `RhinoSurface`.
```
tomvanmele: space before in

src/compas_rhino/geometry/surface.py
robin-oval: because I output a 3D point (u, v, 0) and not a 2D point (u, v)

robin-oval: so uv0[:2] is (u, v)

tomvanmele: i don't understand.
isn't the point of a UV parametrisation to reduce a 3D point to a 2-component representation?
you seem to slice the result to the `uv` component, so why is there a zero?
as far as i know `rs.EvaluateSurface` also takes only `u` and `v` as input parameters...

tomvanmele: let me phrase it differently (because i see all the mapping functions do this): is there any situation in which this zero component would be anything other than zero?

CHANGELOG.md
robin-oval: done

src/compas_rhino/geometry/surface.py
robin-oval: true it's misleading, I changed to 2D inputs and outputs for uv
the user will do the formatting

tomvanmele:
can you resolve the conflicts?

PR #283, opened 2019-05-02
volmesh class update

juney-lee:
- various updates and clean up of the volmesh class (ongoing and more to come...)
- halffaces are restructured as list of vertex keys (like the faces of a mesh)

### DIFF
--- a/src/compas/datastructures/volmesh/volmesh.py
+++ b/src/compas/datastructures/volmesh/volmesh.py
@@ -3,7 +3,6 @@ from __future__ import absolute_import
 from __future__ import division
 
 from ast import literal_eval as _eval
-from math import sqrt
 from copy import deepcopy
 
 from compas.files.obj import OBJ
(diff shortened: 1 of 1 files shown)


gonzalocasas:
@juney-lee if possible, please don't delete the template of the PR, and instead use it to tick the type of change you're suggesting. Also, pls add an entry to the `CHANGELOG.md` file ;)

src/compas/datastructures/volmesh/volmesh.py
```diff
@@ -459,8 +586,10 @@ def add_halfface(self, vertices, fkey=None):
             raise Exception('Corrupt halfface.')
 
         fkey = self._get_face_key(fkey)
+        self.halfface[fkey] = vertices
+        self.facedata[fkey] = self.default_face_attributes
 
-        self.halfface[fkey] = {}
+        edge_attr = self.default_edge_attributes
```
tomvanmele, line 592: this should be a copy

src/compas/datastructures/volmesh/volmesh.py
```diff
@@ -567,19 +729,8 @@ def edges(self, data=False):
...
-        #         vertices = self.halfface_vertices(fkey, ordered=True)
-        #         vset = frozenset(vertices)
-        #         if vset not in seen:
-        #             faces.append(vertices)
-        #         seen.add(vset)
-        # return faces
+        """"Return unique" halfface keys."""
         seen = set()
         faces = []
         for fkey in self.halfface:
```
tomvanmele, line 736: should this not be `for fkey in self.halffaces()`

tomvanmele: is there such a thing as `self.halffaces()`?

tomvanmele (changes requested):
the implementation should be equivalent with the other data structures.
* use iterators for all accessor functions
* all data should be accessible through methods

PR #291, opened 2019-06-18
Face neighborhood

Achillx:
Added `face_neighborhood` function in `compas.datastructures.Mesh` class.
It work the same way as `vertex_neighborhood`.

### Checklist

### DIFF
--- a/src/compas/datastructures/mesh/_mesh.py
+++ b/src/compas/datastructures/mesh/_mesh.py
@@ -2165,6 +2165,44 @@ class Mesh(FromToPickle,
                 nbrs.append(nbr)
         return nbrs
 
+    def face_neighborhood(self, key, ring=1):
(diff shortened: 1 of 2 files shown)


src/compas/datastructures/mesh/_mesh.py
```diff
@@ -2165,6 +2165,44 @@ def face_neighbors(self, fkey):
...
+            for key in nbrs:
+                temp += self.face_neighbors(key)
+
+            nbrs.update(temp)
+
+            i += 1
+
+        if key in nbrs:
+            nbrs.remove(key)
+
```
tomvanmele: is this necessary if you are using a set?

Achillx: nope! I don't know why I added it in the first place!

PR #292, opened 2019-06-18
modify point plane distance for SDF (compas_vol)

worbit:
As we discussed, I will try to base all the `compas_vol` classes on existing compas classes wherever possible. For the `plane` class - because the goal is to create **signed** distance fields - it is necessary that it returns not the absolute value of the distance but both negative _and_ positive values.
If you prefer, I can also add an additional function, e.g. `def distance_point_plane_signed(point, plane)` if this is preferable?

- [x] Bug fix in a **backwards-compatible** manner. (not sure ;)

### Checklist


gonzalocasas (changes requested):
@worbit it's not exactly backwards-compatible because it changes the behavior of an existing function ;)

Also, the docstring explicitly mentions that it returns the absolute value, so it should be changed to reflect the new behavior.

On the other hand, maybe it is worth avoiding the breaking change altogether, either adding an optional argument to this function, or adding a second function that does the non-abs version of distance.

romanarust:
I second the "adding an additional function" `distance_point_plane_signed(point, plane)` to make it more explicit.

worbit:
thank you @romanarust and @gonzalocasas for your comments. I will add a new function, as this is in line with all the `_xy` and `_sqrd` variations, that are also separate functions and not implemented with optional arguments.

src/compas/geometry/distance.py
```diff
@@ -371,6 +371,63 @@ def distance_point_plane(point, plane):
...
+    Parameters
+    ----------
+    point : list
+        Point coordinates.
+    plane : tuple
+        A point and a vector defining a plane.
+
+    Returns
+    -------
+    d : float
```
gonzalocasas:
```suggestion
    float
```

src/compas/geometry/distance.py
```diff
@@ -371,6 +371,63 @@ def distance_point_plane(point, plane):
     return fabs(dot_vectors(vector, normal))
```
gonzalocasas: Maybe consider changing `distance_point_plane` to simply call `distance_point_plane_signed` and return the `fabs` of it, so:
```python
def distance_point_plane(point, plane):
    return fabs(distance_point_plane_signed(point, plane))

```

src/compas/geometry/_primitives/cylinder.py
```diff
@@ -0,0 +1,293 @@
...
+
+__all__ = ['Cylinder']
+
+
+class Cylinder(object):
+    """A cylinder is defined by a plane, a radius and a height.
+
+    Attributes
+    ----------
+    plane: :class:`compas.geometry.Plane`
```
romanarust: I suggest creating the cylinder with a Circle(plane, radius) and the height, and have that 2 as attributes.

src/compas/geometry/_primitives/cylinder.py
```diff
@@ -0,0 +1,293 @@
...
+        The radius of the cylinder.
+    height: float
+        The height of the cylinder.
+
+    Examples
+    --------
+    >>> from compas.geometry import Plane
+    >>> from compas.geometry import Cylinder
+    >>> plane = Plane([0, 0, 0], [0, 0, 1])
+    >>> cylinder = Cylinder(plane, 5, 7)
```
romanarust:
```suggestion
    >>> circle = Circle(plane, 5)
    >>> cylinder = Cylinder(circle, 7)
```

src/compas/geometry/_primitives/cylinder.py
```diff
@@ -0,0 +1,293 @@
...
+        return cylinder
+
+    # ==========================================================================
+    # descriptors
+    # ==========================================================================
+
+    @property
+    def plane(self):
+        """Plane: The plane of the cylinder."""
+        return self._plane
```
romanarust: With the circle as attr, you can still use setter and getter for plane and radius
```suggestion
        return self.circle.plane
```

src/compas/geometry/_primitives/cylinder.py
```diff
@@ -0,0 +1,293 @@
...
+        """Transform the cylinder.
+
+        Parameters
+        ----------
+        transformation : :class:`Transformation`
+            The transformation used to transform the cylinder.
+
+        Examples
+        --------
+        >>> from compas.geometry import Frame
```
romanarust, line 229: To make example section shorter, suggest to remove imports and put them either at the beginning of the file or pass them to the doctest (see below).

src/compas/geometry/_primitives/cylinder.py
```diff
@@ -0,0 +1,293 @@
...
+    cylinder.transform(T)
+    print(cylinder)
+
+    print(Plane.worldXY().data)
+    data = {'plane': Plane.worldXY().data, 'radius': 5., 'height': 7.}
+    cylinder = Cylinder.from_data(data)
+    print(cylinder)
+
+    import doctest
+    doctest.testmod()
```
romanarust, line 294: Like that you pass the imports to the examples
```suggestion
    doctest.testmod(globs=globals())
```

src/compas/geometry/_primitives/torus.py
```diff
@@ -0,0 +1,281 @@
...
+class Torus(object):
+    """A torus is defined by a plane and two radii.
+
+    Attributes
+    ----------
+    plane : :class:`compas.geometry.Plane`
+        The plane of the torus.
+    radius1: float
+        The radius of the axis.
+    radius2: float
```
romanarust: Would call it `pipe_radius` or `radius_pipe`

src/compas/geometry/_primitives/torus.py
```diff
@@ -0,0 +1,281 @@
...
+
+
+class Torus(object):
+    """A torus is defined by a plane and two radii.
+
+    Attributes
+    ----------
+    plane : :class:`compas.geometry.Plane`
+        The plane of the torus.
+    radius1: float
```
romanarust: Would call it `axis_radius` or `radius_axis`

src/compas/geometry/_primitives/torus.py
```diff
@@ -0,0 +1,281 @@
...
+    
+    @property
+    def radius2(self):
+        """float: The radius of the pipe."""
+        return self._radius2
+
+    @radius2.setter
+    def radius2(self, radius):
+        self._radius2 = float(radius)
+
```
romanarust, line 97: would add `inner_radius` and `outer_radius` to the properties and maybe also add `Torus.from_inner_and_outer_radius()`,  as it seems to be often used as constuctor.

src/compas/geometry/_primitives/torus.py
```diff
@@ -0,0 +1,281 @@
...
+        --------
+        >>> from compas.geometry import Plane
+        >>> from compas.geometry import Torus
+        >>> torus = Torus(Plane.worldXY(), 5, 2)
+        >>> sdict = {'plane': Plane.worldXY().data, 'radius1': 5., 'radius2': 2.}
+        >>> sdict == torus.data
+        True
+
+        """
+        return {'plane': Plane.worldXY(),
```
romanarust, line 131:
```suggestion
        return {'plane': self.plane.data,
```

src/compas/geometry/_primitives/torus.py
```diff
@@ -0,0 +1,281 @@
...
+                'radius1': self.radius1,
+                'radius2': self.radius2}
+
+    @data.setter
+    def data(self, data):
+        self.plane = data['plane']
```
romanarust:
```suggestion
        self.plane = Plane.from_data(data['plane'])
```

src/compas/geometry/distance.py
```diff
@@ -365,10 +365,65 @@ def distance_point_plane(point, plane):
     --------
     >>>
 
+    """
+    return fabs(distance_point_plane_signed(point, plane))
+
+
+def distance_point_plane_signed(point, plane):
+    """Compute the distance from a point to a plane defined by three points.
```
romanarust:
```suggestion
    """Compute the signed distance from a point to a plane defined by point and normal.
```

src/compas/geometry/distance.py
```diff
@@ -315,7 +315,7 @@ def distance_point_line_sqrd_xy(point, line):
 
 
 def distance_point_plane(point, plane):
-    r"""Compute the distance from a point to a plane defined by three points.
+    """Compute the distance from a point to a plane defined by three points.
```
romanarust:
```suggestion
    """Compute the distance from a point to a plane defined by point and normal.
```

src/compas/geometry/_primitives/cylinder.py
```diff
@@ -0,0 +1,293 @@
...
+        Returns
+        -------
+        dict
+            The cylinder data.
+
+        """
+        return self.data
+
+    @property
+    def center(self):
```
romanarust, line 148: I find `center` here misleading as this is the center of the bottom circle and not the center of the cylinder

src/compas/geometry/_primitives/cylinder.py
```diff
@@ -0,0 +1,293 @@
...
+    def center(self, point):
+        self.plane.point = point
+
+    @property
+    def area(self):
+        """Float: The area of the cylinder."""
+        return (pi * (self.radius**2) * 2) + (2 * pi * self.radius * self.height)
+
+    @property
+    def circumference(self):
```
romanarust: shouldn't that rather be the property of the circle

src/compas/geometry/_primitives/torus.py
```diff
@@ -0,0 +1,281 @@
+from __future__ import print_function
+from __future__ import absolute_import
+from __future__ import division
+
+from compas.geometry import Plane
+
+__all__ = ['Torus']
+
+
+class Torus(object):
```
romanarust, line 12: to complete, maybe add here `volume` as well

romanarust (changes requested):
Hi! Great, just proposing some smaller changes.

src/compas/geometry/_primitives/cylinder.py
worbit: I *can* add this, but as I copied most of the new cylinder class from other existing primitives, it would have to be changed there as well in order to be coherent?

src/compas/geometry/_primitives/cylinder.py
worbit: center will be the actual center of the cylinder as well, as the distance function will be such that it extends half the height both above and below the circle's plane. property center now returns the circle's center though.

src/compas/geometry/_primitives/cylinder.py
worbit: removed

src/compas/geometry/_primitives/torus.py
worbit: added `volume`and `area`as property

src/compas/geometry/distance.py
worbit: i can change that, but it is somewhat inconsistent throughout the file. most distance methods write `Returns d : float` while their `_xy` versions write `Returns float`.

gonzalocasas: I believe it is incorrect to add the name of the variable being returned. Check [this doc](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_numpy.html) for a sort of authoritative reference on how to do numpy-style docstrings.

src/compas/geometry/_primitives/cylinder.py
romanarust: I don't think this needs to be coherent, it is only practical to make example strings shorter: sometimes needed, sometimes not

src/compas/geometry/_primitives/cylinder.py
romanarust: ah, ok. but then it is strange that the circle is in the center of the cylinder and not represents the base: the bottom of the cylinder.

src/compas/geometry/distance.py
worbit: ok, I will remove it in all the functions (at least in `distance.py`) then.

src/compas/geometry/_primitives/cylinder.py
worbit: in volumetric modelling, all objects are defined as a distance function centered around the origin. a sphere has the center at the origin, a box extends half of each edge length to either side and likewise, a cylinder extends to top and bottom. it is not an extrusion of a circle. but i see your point, it might not correspond to *expected behavior* of any user. any suggestions?

worbit: maybe rename to `center_of_mass`? for primitives, this would be easy to calculate of course. for more complex objects a bit less trivial...

romanarust: Ah yes, `center_of_mass` could be a solution

tomvanmele:
how are we doing on this?
maybe next time, do one PR per primitive (or "unit of change").
otherwise the review is complicated :)

worbit:
I am aware of this, please accept my apologies ;)
this is why I refrained from pushing more lately until the current PR is edited and merged (or not).
please let me know what best works for you. if all at once is not feasible, we can also integrate them one by one as you suggest (meaning: reject this PR and I add the changes to a new branch one at a time)

worbit:
I merged the current state of branch **master** into **compas_vol** and the tests ran successfully. Ready to go?

tomvanmele:
go!

PR #298, opened 2019-06-24
new: quaternion functions and a Quaternion class

funkchaser:
Adding features concerning quaternions:

1. `transformations.quaternions.py` contains new functions operating on or returning quaternions (excluding to/from matrix, which are already in `transformations.matrices.py`)
2. `xforms.quaternion.py` contains new `Quaternion` class 

### DIFF
--- a/src/compas/geometry/transformations/quaternions.py
+++ b/src/compas/geometry/transformations/quaternions.py
@@ -310,8 +310,8 @@ def axis_angle_from_quaternion(q):
     -------
     >>> q = [1,1,0,0]
     >>> axis,angle = axis_angle_from_quaternion(q)
-    >>> print("axis = ", axis, " angle = ", angle, "(%s°)"%math.degrees(angle))
-    axis =  [1.0, 0.0, 0.0]  angle =  1.5707963267948966 (90.0°)
+    >>> print("axis = ", axis, " angle = ", angle)
+    axis =  [1.0, 0.0, 0.0]  angle =  1.5707963267948966
     """
 
     m = matrix_from_quaternion(q)
--- a/src/compas/geometry/transformations/quaternions.py
+++ b/src/compas/geometry/transformations/quaternions.py
@@ -1,7 +1,7 @@
 """
 This module contains functions that operate on and/or return quaternions.
 
-Default convention to represent a quaternion in this module is a tuple of four real values [qw,qx,qy,qz].
+Default convention to represent a quaternion in this module is a list of four real values [qw, qx, qy, qz].
 The first value qw is the scalar (real) part, and qx,qy,qz form the vector (complex, imaginary) part, so that:
 q = qw + qx*i + qy*j + qz*k,
 where i,j,k are basis components.
@@ -37,7 +37,8 @@ from compas.geometry.transformations import matrix_from_euler_angles
 from compas.geometry.transformations import axis_and_angle_from_matrix
 from compas.geometry.transformations import matrix_from_axis_and_angle
 
-#from compas.geometry.basic import allclose
+from compas.geometry.basic import allclose
+
 import math
 
 
@@ -46,26 +47,14 @@ __all__ = [
     'quaternion_unitize',
     'quaternion_is_unit',
     'quaternion_multiply',
-    'quaternion_canonic',
+    'quaternion_canonize',
     'quaternion_conjugate',
     'quaternion_from_euler_angles',
     'euler_angles_from_quaternion',
     'quaternion_from_axis_angle',
     'axis_angle_from_quaternion']
 
-
-# ----------------------------------------------------------------------
-ATOL = 1e-16  # absolute tolerance
-RTOL = 1e-3  # relative tolerance
-
-
-def isclose(a, b, rtol=RTOL, atol=ATOL):
-    # https://docs.scipy.org/doc/numpy/reference/generated/numpy.isclose.html#numpy.isclose
-    # absolute(a - b) <= (atol + rtol * absolute(b))
-
-    return abs(a - b) <= (atol + rtol * abs(b))
-
-# ----------------------------------------------------------------------
+ATOL = 1e-8  # absolute tolerance
 
 
 def quaternion_norm(q):
@@ -74,15 +63,15 @@ def quaternion_norm(q):
     Parameters
     ----------
     q : list
-        Quaternion as a tuple of four real values [qw,qx,qy,qz].
+        Quaternion as a list of four real values ``[qw, qx, qy, qz]``.
 
     Returns
     -------
     float
         The length (euclidean norm) of a quaternion.
 
-    Note
-    ----
+    References
+    ----------
     .. _mathworld quaternion norm
     http://mathworld.wolfram.com/QuaternionNorm.html
     """
@@ -91,10 +80,11 @@ def quaternion_norm(q):
 
 def quaternion_unitize(q):
     """Makes a quaternion unit-length.
+
     Parameters
     ----------
     q : list
-        Quaternion as a tuple of four real values [qw,qx,qy,qz].
+        Quaternion as a list of four real values ``[qw, qx, qy, qz]``.
 
     Returns
     -------
@@ -102,49 +92,53 @@ def quaternion_unitize(q):
         A quaternion of length 1.0.
     """
     n = quaternion_norm(q)
-    if isclose(n, 0.0):
+    if allclose([n], [0.0], ATOL):
         raise ValueError("The given quaternion has zero length.")
     else:
         return [x/n for x in q]
 
 
-def quaternion_is_unit(q):
+def quaternion_is_unit(q, tol=ATOL):
     """Checks if a quaternion is unit-length.
+
     Parameters
     ----------
     q : list
-        Quaternion as a tuple of four real values [qw,qx,qy,qz].
+        Quaternion as a list of four real values ``[qw, qx, qy, qz]``.
 
     Returns
     -------
     bool
-        True if the quaternion is unit-length, and False if otherwise.
+        ``True`` if the quaternion is unit-length, and ``False`` if otherwise.
     """
     n = quaternion_norm(q)
-    return isclose(n, 1.0)
+    return allclose([n], [1.0], tol)
 
 
 def quaternion_multiply(r, q):
     """Multiplies two quaternions.
+
     Parameters
     ----------
     r : list
-        Quaternion as a tuple of four real values [qw,qx,qy,qz].
+        Quaternion as a list of four real values ``[qw, qx, qy, qz]``.
     q : list
-        Quaternion as a tuple of four real values [qw,qx,qy,qz].
+        Quaternion as a list of four real values ``[qw, qx, qy, qz]``.
 
     Returns
     -------
-    list
-        A quaternion p = r * q.
+    list of float
+        A quaternion p = r*q.
 
-    Note
-    ----
+    Notes
+    -----
     Multiplication of two quaternions r * q can be interpreted as applying rotation r to an orientation q,
     provided both r and q are unit-length.
     The result is also unit-length.
     Multiplication of quaternions is not commutative!
 
+    References
+    ----------
     .. _mathworld quaternion:
     http://mathworld.wolfram.com/Quaternion.html
     """
@@ -163,15 +157,15 @@ def quaternion_canonize(q):
     Parameters
     ----------
     q : list
-        Quaternion as a tuple of four real values [qw,qx,qy,qz].
+        Quaternion as a list of four real values ``[qw, qx, qy, qz]``.
 
     Returns
     -------
     list
         A quaternion in a canonic form.
 
-    Note
-    ----
+    Notes
+    -----
     Canonic form means the scalar component is a non-negative number.
 
     """
@@ -187,15 +181,15 @@ def quaternion_conjugate(q):
     Parameters
     ----------
     q : list
-        Quaternion as a tuple of four real values [qw,qx,qy,qz].
+        Quaternion as a list of four real values ``[qw, qx, qy, qz]``.
 
     Returns
     -------
     list
         A conjugate quaternion.
 
-    Note
-    ----
+    References
+    ----------
     .. _mathworld quaternion conjugate
     http://mathworld.wolfram.com/QuaternionConjugate.html
 
@@ -217,20 +211,20 @@ def quaternion_from_euler_angles(e, static=True, axes='xyz'):
     euler_angles : list of float
         Three numbers that represent the angles of rotations about the specified axes.
     static : bool, optional
-        If True, the rotations are applied to a static frame.
-        If False, the rotations are applied to a rotational frame.
-        Defaults to True.
+        If ``True``, the rotations are applied to a static frame.
+        If ``False``, the rotations are applied to a rotational frame.
+        Defaults to ``True``.
     axes : str, optional
         A three-character string specifying the order of the axes.
-        Defaults to 'xyz'.
+        Defaults to ``'xyz'``.
 
     Returns
     -------
     list of float
-        Quaternion [qw,qx,qy,qz] as a list of four real values.
+        Quaternion ``[qw, qx, qy, qz]`` as a list of four real values.
     """
 
-    m = matrix_from_euler_angles(e, static=True, axes='xyz')
+    m = matrix_from_euler_angles(e, static, axes)
     q = quaternion_from_matrix(m)
     return q
 
@@ -241,14 +235,14 @@ def euler_angles_from_quaternion(q, static=True, axes='xyz'):
     Parameters
     ----------
     quaternion : list of float
-        Quaternion [qw,qx,qy,qz] as a list of four real values.
+        Quaternion ``[qw, qx, qy, qz]`` as a list of four real values.
     static : bool, optional
-        If True, the rotations are applied to a static frame.
-        If False, the rotations are applied to a rotational frame.
-        Defaults to True.
+        If ``True``, the rotations are applied to a static frame.
+        If ``False``, the rotations are applied to a rotational frame.
+        Defaults to ``True``.
     axes : str, optional
         A three-character string specifying the order of the axes.
-        Defaults to 'xyz'.
+        Defaults to ``'xyz'``.
 
     Returns
     -------
@@ -259,8 +253,6 @@ def euler_angles_from_quaternion(q, static=True, axes='xyz'):
     e = euler_angles_from_matrix(m, static, axes)
     return e
 
-# ----------------------------------------------------------------------
-
 
 def quaternion_from_axis_angle(axis, angle):
     """Returns a quaternion describing a rotation around the given axis by the given angle.
@@ -268,24 +260,24 @@ def quaternion_from_axis_angle(axis, angle):
     Parameters
     ----------
     axis : list
-        Coordinates [x,y,z] of the rotation axis vector.
+        Coordinates ``[x,y,z]`` of the rotation axis vector.
     angle : float
         Angle of rotation in radians.
 
     Returns
     -------
     quaternion : list
-        Quaternion [qw,qx,qy,qz] as a list of four real values.
+        Quaternion ``[qw, qx, qy, qz]`` as a list of four real values.
 
     Example
     -------
     >>> axis =  [1.0, 0.0, 0.0]
     >>> angle = math.pi/2
     >>> q = quaternion_from_axis_angle(axis,angle)
-    >>> print(q)
-    [0.7071067811865476, 0.7071067811865475, 0.0, 0.0]
+    >>> allclose(q, [math.sqrt(2)/2, math.sqrt(2)/2, 0, 0])
+    True
     """
-    m = matrix_from_axis_and_angle(axis, angle, point=None, rtype='list')
+    m = matrix_from_axis_and_angle(axis, angle, point, rtype)
     q = quaternion_from_matrix(m)
     return q
 
@@ -296,12 +288,12 @@ def axis_angle_from_quaternion(q):
     Parameters
     ----------
     quaternion : list
-        Quaternion [qw,qx,qy,qz] as a list of four real values.
+        Quaternion ``[qw, qx, qy, qz]`` as a list of four real values.
 
     Returns
     -------
     axis : list of float
-        Coordinates [x,y,z] of the rotation axis vector.
+        Coordinates ``[x,y,z]`` of the rotation axis vector.
     angle : float
         Angle of rotation in radians.
 
@@ -311,8 +303,8 @@ def axis_angle_from_quaternion(q):
     >>> axis,angle = axis_angle_from_quaternion(q)
     >>> allclose(axis, [1., 0., 0.])
     True
-    >>> isclose(angle, 1.57079, 1e-5)
-    axis =  [1.0, 0.0, 0.0]  angle =  1.5707963267948966
+    >>> allclose([angle], [math.pi/2], 1e-5)
+    True
     """
 
     m = matrix_from_quaternion(q)
--- a/src/compas/geometry/transformations/quaternions.py
+++ b/src/compas/geometry/transformations/quaternions.py
@@ -310,7 +310,9 @@ def axis_angle_from_quaternion(q):
     -------
     >>> q = [1,1,0,0]
     >>> axis,angle = axis_angle_from_quaternion(q)
-    >>> print("axis = ", axis, " angle = ", angle)
+    >>> allclose(axis, [1., 0., 0.])
+    True
+    >>> isclose(angle, 1.57079, 1e-5)
     axis =  [1.0, 0.0, 0.0]  angle =  1.5707963267948966
     """
 
--- a/src/compas/geometry/transformations/quaternions.py
+++ b/src/compas/geometry/transformations/quaternions.py
@@ -231,7 +231,7 @@ def quaternion_from_euler_angles(e, static=True, axes='xyz'):
         Quaternion [qw,qx,qy,qz] as a list of four real values.
     """
 
-    m = matrix_from_euler_angles(e, static=True, axes='xyz')
+    m = matrix_from_euler_angles(e, static, axes)
     q = quaternion_from_matrix(m)
     return q
 
--- a/src/compas/geometry/transformations/quaternions.py
+++ b/src/compas/geometry/transformations/quaternions.py
@@ -1,33 +1,53 @@
 """
 This module contains functions that operate on and/or return quaternions.
 
-Default convention to represent a quaternion in this module is a list of four real values [qw, qx, qy, qz].
-The first value qw is the scalar (real) part, and qx,qy,qz form the vector (complex, imaginary) part, so that:
-q = qw + qx*i + qy*j + qz*k,
-where i,j,k are basis components.
+Notes
+-----
+The default convention to represent a quaternion :math:`q` in this module is by four real values **w**, **x**, **y**, **z**.
+The first value **w** is the scalar (real) part, and **x**, **y**, **z** form the vector (complex, imaginary) part [1]_, so that:
 
-Quaternion algebra basics:
-i*i = j*j = k*k = i*j*k = -1
-i*j = k    j*i = -k
-j*k = i    k*j = -i
-k*i = j    i*k = -j
+:math:`q = w + xi + yj + zk`
 
-Quaternion as rotation:
-A rotation through an angle theta around an axis defined by a euclidean unit vector u = ux*i + uy*j + uz*k
+where :math:`i, j, k` are basis components with following multiplication rules [2]_:
+
+:math:`ii = jj = kk = ijk = -1`
+
+:math:`ij = k,\\qquad ji = -k`
+
+:math:`jk = i,\\qquad kj = -i`
+
+:math:`ki = j,\\qquad ik = -j`
+
+Quaternions are associative but not commutative.
+
+
+**Quaternion as rotation.**
+A rotation through an angle :math:`\\theta` around an axis defined by a euclidean unit vector :math:`u = u_{x}i + u_{y}j + u_{z}k`
 can be represented as a quaternion:
-q = cos(theta/2) + sin(theta/2) * [ux*i + uy*j + uz*k]
+
+:math:`q = cos(\\frac{\\theta}{2}) + sin(\\frac{\\theta}{2})  [u_{x}i + u_{y}j + u_{z}k]`
+
 i.e.:
-qw = cos(theta/2)
-qx = sin(theta/2) * ux
-qy = sin(theta/2) * uy
-qz = sin(theta/2) * uz
+
+:math:`w = cos(\\frac{\\theta}{2})`
+
+:math:`x = sin(\\frac{\\theta}{2})  u_{x}`
+
+:math:`y = sin(\\frac{\\theta}{2})  u_{y}`
+
+:math:`z = sin(\\frac{\\theta}{2})  u_{z}`
 
 For a quaternion to represent a rotation or orientation, it must be unit-length.
+A quaternion representing a rotation :math:`p` resulting from applying a rotation :math:`r` to a rotation :math:`q`, i.e.:
+:math:`p = rq`,
+is also unit-length.
+
 
 References
 ----------
-.. https://github.com/matthew-brett/transforms3d/blob/master/transforms3d/quaternions.py
-.. http://mathworld.wolfram.com/Quaternion.html
+.. [1] http://mathworld.wolfram.com/Quaternion.html
+.. [2] http://mathworld.wolfram.com/HamiltonsRules.html
+.. [3] https://github.com/matthew-brett/transforms3d/blob/master/transforms3d/quaternions.py
 """
 
 from compas.geometry.transformations import quaternion_from_matrix
@@ -54,7 +74,7 @@ __all__ = [
     'quaternion_from_axis_angle',
     'axis_angle_from_quaternion']
 
-ATOL = 1e-8  # absolute tolerance
+ATOL = 1e-6  # absolute tolerance
 
 
 def quaternion_norm(q):
@@ -63,7 +83,7 @@ def quaternion_norm(q):
     Parameters
     ----------
     q : list
-        Quaternion as a list of four real values ``[qw, qx, qy, qz]``.
+        Quaternion as a list of four real values ``[w, x, y, z]``.
 
     Returns
     -------
@@ -84,12 +104,12 @@ def quaternion_unitize(q):
     Parameters
     ----------
     q : list
-        Quaternion as a list of four real values ``[qw, qx, qy, qz]``.
+        Quaternion as a list of four real values ``[w, x, y, z]``.
 
     Returns
     -------
     list
-        A quaternion of length 1.0.
+        Quaternion of length 1 as a list of four real values ``[nw, nx, ny, nz]``.
     """
     n = quaternion_norm(q)
     if allclose([n], [0.0], ATOL):
@@ -104,7 +124,9 @@ def quaternion_is_unit(q, tol=ATOL):
     Parameters
     ----------
     q : list
-        Quaternion as a list of four real values ``[qw, qx, qy, qz]``.
+        Quaternion as a list of four real values ``[w, x, y, z]``.
+    tol : float, optional
+        Requested decimal precision.
 
     Returns
     -------
@@ -121,19 +143,19 @@ def quaternion_multiply(r, q):
     Parameters
     ----------
     r : list
-        Quaternion as a list of four real values ``[qw, qx, qy, qz]``.
+        Quaternion as a list of four real values ``[rw, rx, ry, rz]``.
     q : list
         Quaternion as a list of four real values ``[qw, qx, qy, qz]``.
 
     Returns
     -------
-    list of float
-        A quaternion p = r*q.
+    list
+        Quaternion :math:`p = rq` as a list of four real values ``[pw, px, py, pz]``.
 
     Notes
     -----
-    Multiplication of two quaternions r * q can be interpreted as applying rotation r to an orientation q,
-    provided both r and q are unit-length.
+    Multiplication of two quaternions :math:`p = rq` can be interpreted as applying rotation :math:`r` to an orientation :math:`q`,
+    provided that both :math:`r` and :math:`q` are unit-length.
     The result is also unit-length.
     Multiplication of quaternions is not commutative!
 
@@ -157,12 +179,12 @@ def quaternion_canonize(q):
     Parameters
     ----------
     q : list
-        Quaternion as a list of four real values ``[qw, qx, qy, qz]``.
+        Quaternion as a list of four real values ``[w, x, y, z]``.
 
     Returns
     -------
     list
-        A quaternion in a canonic form.
+        Quaternion in a canonic form as a list of four real values ``[cw, cx, cy, cz]``.
 
     Notes
     -----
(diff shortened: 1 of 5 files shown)


src/compas/geometry/transformations/quaternions.py
```diff
@@ -0,0 +1,324 @@
...
+    'quaternion_canonic',
+    'quaternion_conjugate',
+    'quaternion_from_euler_angles',
+    'euler_angles_from_quaternion',
+    'quaternion_from_axis_angle',
+    'axis_angle_from_quaternion']
+
+
+"""
+This module contains functions that operate on and/or return quaternions.
```
gonzalocasas: The module docstring should go to the very top of the file, before the imports

src/compas/geometry/transformations/quaternions.py
```diff
@@ -0,0 +1,324 @@
...
+
+Default convention to represent a quaternion in this module is a tuple of four real values [qw,qx,qy,qz].
+The first value qw is the scalar (real) part, and qx,qy,qz form the vector (complex, imaginary) part, so that:
+q = qw + qx*i + qy*j + qz*k,
+where i,j,k are basis components.
+
+Quaternion algebra basics:
+i*i = j*j = k*k = i*j*k = -1
```
gonzalocasas: I have the feeling this will not render very nicely. I think rst / Sphinx supports the `.. math::` directive to write math formulae (not sure the specific format required, it might be latex)

src/compas/geometry/transformations/quaternions.py
```diff
@@ -0,0 +1,324 @@
...
+    The result is also unit-length.
+    Multiplication of quaternions is not commutative!
+
+    .. _mathworld quaternion:
+    http://mathworld.wolfram.com/Quaternion.html
+    """
+    rw, rx, ry, rz = r
+    qw, qx, qy, qz = q
+    pw = rw*qw - rx*qx - ry*qy - rz*qz
+    px = rw*qx + rx*qw + ry*qz - rz*qy
```
gonzalocasas, line 170: I think you don't have `autopep8` and/or `flake8` installed, otherwise it would have inserted spaces between the arithmetic operators. Please install them and then run `Format document` to quickly fix all those little quirks.

src/compas/geometry/transformations/quaternions.py
```diff
@@ -0,0 +1,324 @@
...
+    py = rw*qy - rx*qz + ry*qw + rz*qx
+    pz = rw*qz + rx*qy - ry*qx + rz*qw
+    return [pw, px, py, pz]
+
+
+def quaternion_canonic(q):
```
gonzalocasas: I would prefer to name this `quaternion_canonize` to be aligned with most other functions that are named with a verb, rather than a noun.

src/compas/geometry/xforms/quaternion.py
```diff
@@ -0,0 +1,118 @@
+from compas.geometry.transformations import quaternion_multiply
+from compas.geometry.transformations import quaternion_conjugate
+from compas.geometry.transformations import quaternion_unitize
+from compas.geometry.transformations import quaternion_canonic
+from compas.geometry.transformations import quaternion_norm
+from compas.geometry.transformations import quaternion_is_unit
+
+__all__ = ['Quaternion']
```
gonzalocasas, line 12: I assume this module needs to be imported at the `compas.geometry` level and be documented next to Frame, Rotation, etc.

src/compas/geometry/xforms/quaternion.py
```diff
@@ -0,0 +1,118 @@
...
+
+class Quaternion(object):
+    def __init__(self, w,x,y,z):
+
+        # TODO: add checks for value type
```
gonzalocasas: Rather than type checking, you could simply convert to float. Normally that will be a no-op, since the arguments will be floats already. If they are convertible, they will be, and if they are not, and exception will be thrown, which is correct.

src/compas/geometry/xforms/quaternion.py
```diff
@@ -0,0 +1,118 @@
...
+
+        self.w = w
+        self.x = x
+        self.y = y
+        self.z = z
+
+    def __iter__(self):
+        return iter([self.w, self.x, self.y, self.z])
+
+    def __str__(self):
```
gonzalocasas, line 95: Consider also adding `__repr__`

src/compas/geometry/xforms/quaternion.py
```diff
@@ -0,0 +1,118 @@
...
+        return "Quaternion = %s" % list(self)
+
+    def __mul__(self, other):
+        """Multiply operator for two quaternions.
+
+        Parameters
+        ----------
+        R
+            First argument, a Quaternion object.
+        Q
```
gonzalocasas: Here there parameter names do not match the definition. Should be only one and called `other`

gonzalocasas (changes requested):
It is generally looking good! I added a few comments here and there, and I would suggest a couple more things:
 - It would be important to have a `from_quaternion` on the `Frame` class and a match `from_frame` in the `Quaternion` class, to make this all more integrated.
 - This type of code lends itself very well to be unit tested, if possible, I would suggest adding at least a few tests for some of the main functions (e.g. normalize, unitize, canonize and various conversions).

src/compas/geometry/transformations/quaternions.py
```diff
@@ -0,0 +1,324 @@
...
+        A three-character string specifying the order of the axes.
+        Defaults to 'xyz'.
+
+    Returns
+    -------
+    list of float
+        Quaternion [qw,qx,qy,qz] as a list of four real values.
+    """
+
+    m = matrix_from_euler_angles(e, static=True, axes='xyz')
```
romanarust:
```suggestion
    m = matrix_from_euler_angles(e, static, axes)
```

src/compas/geometry/transformations/quaternions.py
```diff
@@ -0,0 +1,324 @@
...
+    -------
+    quaternion : list
+        Quaternion [qw,qx,qy,qz] as a list of four real values.
+
+    Example
+    -------
+    >>> axis =  [1.0, 0.0, 0.0]
+    >>> angle = math.pi/2
+    >>> q = quaternion_from_axis_angle(axis,angle)
+    >>> print(q)
```
romanarust: As this might not be true on all machines, suggest to use `compas.geometry.basic.allclose`
```suggestion
    >>> allclose(q, [0.707107, 0.707107, 0, 0], tol=1e-05)
    True
```

src/compas/geometry/transformations/quaternions.py
```diff
@@ -0,0 +1,324 @@
...
+.. http://mathworld.wolfram.com/Quaternion.html
+
+"""
+
+# ----------------------------------------------------------------------
+ATOL = 1e-16  # absolute tolerance
+RTOL = 1e-3  # relative tolerance
+
+
+def isclose(a, b, atol=ATOL, rtol=RTOL):
```
romanarust: Better use same order as numpy.isclose or math.isclose (new in 3.5)
Also, this function should move to compas.geometry.basic (to `allclose`)

```suggestion
def isclose(a, b, rtol=RTOL, atol=ATOL):
```

src/compas/geometry/transformations/quaternions.py
```diff
@@ -0,0 +1,324 @@
...
+    axis : list of float
+        Coordinates [x,y,z] of the rotation axis vector.
+    angle : float
+        Angle of rotation in radians.
+
+    Example
+    -------
+    >>> q = [1,1,0,0]
+    >>> axis,angle = axis_angle_from_quaternion(q)
+    >>> print("axis = ", axis, " angle = ", angle)
```
romanarust:
```suggestion
    >>> allclose(axis, [1., 0., 0.])
    True
    >>> isclose(angle, 1.57079, 1e-5)
```

src/compas/geometry/xforms/quaternion.py
romanarust: Similar to Point, Vector, Frame, ... `__str__` and `__repr__` should return something that looks like constructor, e.g.
```
return 'Quaternion({:.{prec}f}, {:.{prec}f}, {:.{prec}f}, {:.{prec}f})'.format(*self.wxyz, prec=3)
```

src/compas/geometry/xforms/quaternion.py
```diff
@@ -0,0 +1,118 @@
...
+        Quaternion
+            P = R * Q
+
+        Examples
+        --------
+        >>> R = Quaternion(1,-2,3,-4).unitized
+        >>> Q = Quaternion(0,1,-2,-3).unitized
+        >>> P = R * Q
+        >>> print(P)
+        Quaternion = [-0.19518001458970669, -0.7807200583588265, -0.5855400437691198, -0.09759000729485334]
```
romanarust: Similar here, rather use `allclose`

src/compas/geometry/xforms/quaternion.py
```diff
@@ -0,0 +1,118 @@
...
+        Multiplication of two quaternions R * Q can be interpreted as applying rotation R to an orientation Q,
+        provided both R and Q are unit-length.
+        The result is also unit-length.
+        Multiplication of quaternions is not commutative!
+
+        """
+        p = quaternion_multiply(list(self), list(other))
+        return Quaternion(*p)
+
+    @property
```
romanarust, line 166: add also property `wxyz`

src/compas/geometry/transformations/quaternions.py
```diff
@@ -0,0 +1,324 @@
...
+    quaternion : list
+        Quaternion [qw,qx,qy,qz] as a list of four real values.
+
+    Example
+    -------
+    >>> axis =  [1.0, 0.0, 0.0]
+    >>> angle = math.pi/2
+    >>> q = quaternion_from_axis_angle(axis,angle)
+    >>> print(q)
+    [0.7071067811865476, 0.7071067811865475, 0.0, 0.0]
```
romanarust:
```suggestion
```

romanarust (changes requested):
Great! 
Can you also change the quaternion() property of the Frame such that it returns no list, but a `Quaternion` and update the changelog.
Thanks!
Add the change to the CHANGELOG.md

src/compas/geometry/transformations/quaternions.py
tomvanmele: yes, the format is latex. check the numerical package for examples...

src/compas/geometry/transformations/quaternions.py
tomvanmele: agreed

src/compas/geometry/xforms/quaternion.py
tomvanmele: yes

src/compas/geometry/transformations/quaternions.py
tomvanmele: please also mind the syntax.

```python
q = [1, 1, 0, 0]
axis, angle = ...
```

src/compas/geometry/transformations/quaternions.py
```diff
@@ -310,8 +310,8 @@ def axis_angle_from_quaternion(q):
     -------
     >>> q = [1,1,0,0]
     >>> axis,angle = axis_angle_from_quaternion(q)
-    >>> print("axis = ", axis, " angle = ", angle, "(%s°)"%math.degrees(angle))
-    axis =  [1.0, 0.0, 0.0]  angle =  1.5707963267948966 (90.0°)
+    >>> print("axis = ", axis, " angle = ", angle)
+    axis =  [1.0, 0.0, 0.0]  angle =  1.5707963267948966
     """
```
tomvanmele, line 330: we should agree on a number of digits in examples and docstrings.
and have the `allclose`, `isclose`, ... functions check accordingly by default.
engineering precision, for example, is a fraction of a millimeter, but not more as it does not make sense at the building scale...

src/compas/geometry/transformations/quaternions.py
tomvanmele: please also mind the syntax (space after comma, spaces around operators, ...)

src/compas/geometry/transformations/quaternions.py
tomvanmele: it is a detail, but 

```
Note
-----
```
and 
```
Notes
------
```
render very differently.
please use `Notes`

src/compas/geometry/transformations/quaternions.py
```diff
@@ -0,0 +1,324 @@
...
+    n = quaternion_norm(q)
+    if isclose(n, 0.0):
+        raise ValueError("The given quaternion has zero length.")
+    else:
+        return [x/n for x in q]
+
+
+def quaternion_is_unit(q):
+    """Checks if a quaternion is unit-length.
+    Parameters
```
tomvanmele, line 124: please leave a blank line above the header (`Parameters`)

src/compas/geometry/transformations/quaternions.py
```diff
@@ -0,0 +1,324 @@
...
+RTOL = 1e-3  # relative tolerance
+
+
+def isclose(a, b, atol=ATOL, rtol=RTOL):
+    # https://docs.scipy.org/doc/numpy/reference/generated/numpy.isclose.html#numpy.isclose
+    # absolute(a - b) <= (atol + rtol * absolute(b))
+
+    return abs(a - b) <= (atol + rtol * abs(b))
+
+# ----------------------------------------------------------------------
```
tomvanmele: not a big fan of these.
either make a section comment with some information or leave it out

```
# --------------------------------------------------------------------------
# ...
# --------------------------------------------------------------------------
```

src/compas/geometry/transformations/quaternions.py
```diff
@@ -0,0 +1,324 @@
...
+        A quaternion of length 1.0.
+    """
+    n = quaternion_norm(q)
+    if isclose(n, 0.0):
+        raise ValueError("The given quaternion has zero length.")
+    else:
+        return [x/n for x in q]
+
+
+def quaternion_is_unit(q):
```
tomvanmele: more of a point for discussion rahter than a comment.
i think functions that return a boolean should be named such that they make sense in an `if` statement.
in that sense `quaternion_is_unit` is a good name: `if quaternion_is_unit: ...`.
the rest of the framework is named differently though with the `is` at the front: `if is_quaternion_unit: ...`

what do you guys think about this?

src/compas/geometry/transformations/quaternions.py
```diff
@@ -0,0 +1,324 @@
...
+    bool
+        True if the quaternion is unit-length, and False if otherwise.
+    """
+    n = quaternion_norm(q)
+    return isclose(n, 1.0)
+
+
+def quaternion_multiply(r, q):
+    """Multiplies two quaternions.
+    Parameters
```
tomvanmele, line 143: blank line

src/compas/geometry/transformations/quaternions.py
```diff
@@ -0,0 +1,324 @@
...
+    ----------
+    q : list
+        Quaternion as a tuple of four real values [qw,qx,qy,qz].
+
+    Returns
+    -------
+    list
+        A quaternion in a canonic form.
+
+    Note
```
tomvanmele: Notes

src/compas/geometry/transformations/quaternions.py
```diff
@@ -0,0 +1,324 @@
...
+        Quaternion as a tuple of four real values [qw,qx,qy,qz].
+
+    Returns
+    -------
+    list
+        A conjugate quaternion.
+
+    Note
+    ----
+    .. _mathworld quaternion conjugate
```
tomvanmele, line 215: technically this would go in a `References` section, and `Notes` would refer to it...

src/compas/geometry/transformations/quaternions.py
```diff
@@ -0,0 +1,324 @@
...
+        return [x for x in q]
+
+
+def quaternion_conjugate(q):
+    """Conjugate of a quaternion.
+
+    Parameters
+    ----------
+    q : list
+        Quaternion as a tuple of four real values [qw,qx,qy,qz].
```
tomvanmele: type is indicated as `list`, but you describe it as a tuple.
please also mind the syntax: `[qw, qx, qy, qz]`

tomvanmele: and add code ticks around code: ``[qw, qx, qy, qz]``.
otherwise they are rendered as plain text.

src/compas/geometry/transformations/quaternions.py
```diff
@@ -0,0 +1,324 @@
...
+
+def quaternion_from_euler_angles(e, static=True, axes='xyz'):
+    """Returns a quaternion from Euler angles.
+
+    Parameters
+    ----------
+    euler_angles : list of float
+        Three numbers that represent the angles of rotations about the specified axes.
+    static : bool, optional
+        If True, the rotations are applied to a static frame.
```
tomvanmele: you write `True` and not true.
so you write code.
please add code ticks.

src/compas/geometry/transformations/quaternions.py
```diff
@@ -0,0 +1,324 @@
...
+# matrix_from_quaternion(q) - already in compas/geometry/transformations/matrices.py
+# ----------------------------------------------------------------------
+
+
+def quaternion_from_euler_angles(e, static=True, axes='xyz'):
+    """Returns a quaternion from Euler angles.
+
+    Parameters
+    ----------
+    euler_angles : list of float
```
tomvanmele: again just a question for discussion.
many of the inputs that just require something to be "iterable" are marked as `list`.
the `numpy` doc guidelines suggest to be as specific as possible, which in this case would be something like `iterable(3) of float` or perhaps just `float[3]`.

also, in the sense of "iterable" or `list` an `iterator` and a `list` are really not the same.
the "iterable" can be used only once in the code block because it gets consumed.
a `list` or `tuple` can be reused multiple times...

tomvanmele: @romanarust 
@gonzalocasas

src/compas/geometry/transformations/quaternions.py
```diff
@@ -0,0 +1,324 @@
...
+        Three numbers that represent the angles of rotations about the specified axes.
+    static : bool, optional
+        If True, the rotations are applied to a static frame.
+        If False, the rotations are applied to a rotational frame.
+        Defaults to True.
```
tomvanmele: code ticks, or write true...

src/compas/geometry/transformations/quaternions.py
```diff
@@ -0,0 +1,324 @@
...
+    axes : str, optional
+        A three-character string specifying the order of the axes.
+        Defaults to 'xyz'.
```
tomvanmele: i would add code ticks: ``'xyz'``

src/compas/geometry/transformations/quaternions.py
```diff
@@ -0,0 +1,324 @@
...
+
+    Returns
+    -------
+    list of float
+        Quaternion [qw,qx,qy,qz] as a list of four real values.
```
tomvanmele: syntax

src/compas/geometry/transformations/quaternions.py
```diff
@@ -0,0 +1,324 @@
...
+    Returns
+    -------
+    list of float
+        Three Euler angles.
+    """
+    m = matrix_from_quaternion(q)
+    e = euler_angles_from_matrix(m, static, axes)
+    return e
+
+# ----------------------------------------------------------------------
```
tomvanmele: same comment as before

src/compas/geometry/transformations/quaternions.py
```diff
@@ -0,0 +1,324 @@
...
+
+
+def quaternion_from_axis_angle(axis, angle):
+    """Returns a quaternion describing a rotation around the given axis by the given angle.
+
+    Parameters
+    ----------
+    axis : list
+        Coordinates [x,y,z] of the rotation axis vector.
```
tomvanmele: code ticks

src/compas/geometry/transformations/quaternions.py
```diff
@@ -0,0 +1,324 @@
...
+    angle : float
+        Angle of rotation in radians.
+
+    Returns
+    -------
+    quaternion : list
+        Quaternion [qw,qx,qy,qz] as a list of four real values.
```
tomvanmele: syntax

src/compas/geometry/xforms/quaternion.py
tomvanmele: agreed

src/compas/geometry/xforms/quaternion.py
```diff
@@ -0,0 +1,118 @@
...
+from compas.geometry.transformations import quaternion_conjugate
+from compas.geometry.transformations import quaternion_unitize
+from compas.geometry.transformations import quaternion_canonic
+from compas.geometry.transformations import quaternion_norm
+from compas.geometry.transformations import quaternion_is_unit
+
+__all__ = ['Quaternion']
+
+class Quaternion(object):
+    def __init__(self, w,x,y,z):
```
tomvanmele: leave a blank line and add a docstring

src/compas/geometry/xforms/quaternion.py
tomvanmele: and mention the type of `other`

src/compas/geometry/xforms/quaternion.py
```diff
@@ -0,0 +1,118 @@
...
+            Second argument, a Quaternion object.
+
+        Returns
+        -------
+        Quaternion
+            P = R * Q
+
+        Examples
+        --------
+        >>> R = Quaternion(1,-2,3,-4).unitized
```
tomvanmele: syntax

src/compas/geometry/xforms/quaternion.py
tomvanmele: also, the `__rep__` should return something with reasonable precision, for example 4 digits...

src/compas/geometry/xforms/quaternion.py
```diff
@@ -0,0 +1,118 @@
...
+        >>> Q = Quaternion(0,1,-2,-3).unitized
+        >>> P = R * Q
+        >>> print(P)
+        Quaternion = [-0.19518001458970669, -0.7807200583588265, -0.5855400437691198, -0.09759000729485334]
+        >>> print(quaternion_is_unit(P))
+        True
+
+        Note
```
tomvanmele: Notes

src/compas/geometry/xforms/quaternion.py
```diff
@@ -0,0 +1,118 @@
...
+        return Quaternion(*qc)
+
+    @property
+    def norm(self):
+        """
+        Returns the length (euclidean norm) of the quaternion.
+        """
+        return quaternion_norm(list(self))
+
+if __name__ =="__main__":
```
tomvanmele: would add a comment section here with as comment `Main`.
see virtually any other COMPAS module for an example...

tomvanmele (changes requested):
many small things, but looks good overall

tomvanmele:
no progress here?

src/compas/geometry/transformations/quaternions.py
funkchaser: ok

src/compas/geometry/transformations/quaternions.py
funkchaser: ok

src/compas/geometry/xforms/quaternion.py
funkchaser: ok

src/compas/geometry/xforms/quaternion.py
funkchaser: ok.
what I wanted to stress using R, Q and P notation is the order P= R x Q and not Q x R.

src/compas/geometry/transformations/quaternions.py
funkchaser: ok

src/compas/geometry/xforms/quaternion.py
funkchaser: should  `__str__` and `__repr__` be the same?

src/compas/geometry/xforms/quaternion.py
funkchaser: ok, but I'd only see it for "external" use (output as a list). In the methods I'd leave it as list(self).

src/compas/geometry/transformations/quaternions.py
funkchaser: I believe these equations are already PEP8 formatted? [https://www.python.org/dev/peps/pep-0008/#whitespace-in-expressions-and-statements ](url)

src/compas/geometry/transformations/quaternions.py
funkchaser: where will this module docstring and other comments in the methods in this module apprear in the documentation? it seems the modules in `geometry.transformations` are hidden for now, right?

src/compas/geometry/transformations/quaternions.py
funkchaser: ok

src/compas/geometry/transformations/quaternions.py
funkchaser: ok

src/compas/geometry/xforms/quaternion.py
funkchaser: added. 
Since the `@property`  methods are not in the docs, where will the examples within them appear?

funkchaser:
> Great!
> Can you also change the quaternion() property of the Frame such that it returns no list, but a `Quaternion` and update the changelog.
> Thanks!
> Add the change to the CHANGELOG.md

Ok, should it be changed in `Rotation` accordingly as well? 
Also, then I'd expect that in all "axis+angle" methods, the axis is a `Vector` object, right?

> * It would be important to have a `from_quaternion` on the `Frame` class and a match `from_frame` in the `Quaternion` class, to make this all more integrated.

added `Quaternion.from_Frame`. 
`Frame.from_quaternion` already exists, however it takes a list as an input. 

I guess this opens up a bit a larger discussion whether we pass objects or list into constructors?

src/compas/geometry/transformations/quaternions.py
tomvanmele: all modules beyond the second module are "hidden" because functionality is meant to be pulled up and used from that second level. otherwise the user needs to be aware of the underlying structure of those modules further down. it also provides the opportunity to move things around without requiring users to change their code.

this means you should add the module level information to the docstring of `compas.geometry.__init__.py`, for example in a section "Quaternions".

it is true that the geometry module is getting quite big and perhaps a bit unmanageable in that sense. however, for now, we will stick to this convention...

src/compas/geometry/transformations/quaternions.py
```diff
@@ -0,0 +1,317 @@
...
+    'quaternion_is_unit',
+    'quaternion_multiply',
+    'quaternion_canonize',
+    'quaternion_conjugate',
+    'quaternion_from_euler_angles',
+    'euler_angles_from_quaternion',
+    'quaternion_from_axis_angle',
+    'axis_angle_from_quaternion']
+
+ATOL = 1e-8  # absolute tolerance
```
tomvanmele: should we add this to `compas`? => `compas.ATOL` like with the precision variable `compas.PRECISION`. it seems to make sense to control this "COMPAS-wide"...

tomvanmele: @gonzalocasas 
@romanarust

src/compas/geometry/transformations/quaternions.py
tomvanmele: you are right. there is indeed a recommendation about operator priority. leave it as it is...

src/compas/geometry/xforms/quaternion.py
tomvanmele: no. 
`__str__` can return whatever you want. `__repr__` should be a string representation that can be used in combination with `eval` to reproduce the object.

src/compas/geometry/xforms/quaternion.py
```diff
@@ -0,0 +1,138 @@
...
+
+    @property
+    def is_unit(self):
+        """
+        Returns ``True`` if the quaternion is unit-length or ``False`` if otherwise.
+        """
+        return quaternion_is_unit(list(self))
+
+    @property
+    def unitized(self):
```
tomvanmele, line 194: this should be a method, not a property...

src/compas/geometry/xforms/quaternion.py
```diff
@@ -0,0 +1,138 @@
...
+
+    def canonize(self):
+        """
+        Makes the quaternion canonic.
+        """
+        qc = quaternion_canonize(list(self))
+        self.w, self.x, self.y, self.z = qc
+
+    @property
+    def canonized(self):
```
tomvanmele, line 208: this should be a method, not a property...

src/compas/geometry/transformations/quaternions.py
funkchaser: ok. now the functions appear in the docs, but not the top module docstring, so I copied it into the Quaternion class dosctring, ok?
Still I wonder where will `@property` and operator functions be visible in the documentation?

src/compas/geometry/transformations/quaternions.py
funkchaser: on the one hand it would be handy, on the other - I doubt you one number will fit all cases. You might want a different precision (assuming SI units only anyway) for e.g. meters than radians than Newtons.. ?

src/compas/geometry/xforms/quaternion.py
funkchaser: ok

src/compas/geometry/xforms/quaternion.py
funkchaser: ok

src/compas/geometry/xforms/quaternion.py
```diff
@@ -0,0 +1,225 @@
...
+from compas.geometry.transformations import quaternion_is_unit
+
+from compas.geometry.basic import allclose
+
+__all__ = ['Quaternion']
+
+
+class Quaternion(object):
+
+    """Creates a ``Quaternion`` object.
```
tomvanmele, line 17: docstrings with lots of math and especially latex symbols can be marked as "raw" (r""") to avoid having to escape backslashes.

src/compas/geometry/xforms/quaternion.py
```diff
@@ -0,0 +1,225 @@
...
+    Notes
+    -----
+    The default convention to represent a quaternion :math:`q` in this module is by four real values **w**, **x**, **y**, **z**.
+    The first value **w** is the scalar (real) part, and **x**, **y**, **z** form the vector (complex, imaginary) part [1]_, so that:
+
+    :math:`q = w + xi + yj + zk`
+
+    where :math:`i, j, k` are basis components with following multiplication rules [2]_:
+
+    :math:`ii = jj = kk = ijk = -1`
```
tomvanmele, line 44: `:math:` is typically used for inline math. to add a block of math stuff use `.. math::`

src/compas/geometry/xforms/quaternion.py
```diff
@@ -0,0 +1,225 @@
...
+
+    **Quaternion as rotation.**
+    A rotation through an angle :math:`\\theta` around an axis defined by a euclidean unit vector :math:`u = u_{x}i + u_{y}j + u_{z}k`
+    can be represented as a quaternion:
+
+    :math:`q = cos(\\frac{\\theta}{2}) + sin(\\frac{\\theta}{2})  [u_{x}i + u_{y}j + u_{z}k]`
+
+    i.e.:
+
+    :math:`w = cos(\\frac{\\theta}{2})`
```
tomvanmele, line 63: `.. math::`

src/compas/geometry/xforms/quaternion.py
tomvanmele: is there a reason for providing `wxyz` and `xyzw`?

seems confusing because once the quaternion is converted to a list there is no way to distinguish between the `wxyz` and `xyzw` form...

src/compas/geometry/transformations/quaternions.py
```diff
@@ -0,0 +1,339 @@
+"""
```
tomvanmele, line 1: use the "raw" flag to avoid having to escape backslashes in math formulas: `r"""`

src/compas/geometry/transformations/quaternions.py
```diff
@@ -0,0 +1,339 @@
...
+Notes
+-----
+The default convention to represent a quaternion :math:`q` in this module is by four real values **w**, **x**, **y**, **z**.
+The first value **w** is the scalar (real) part, and **x**, **y**, **z** form the vector (complex, imaginary) part [1]_, so that:
+
+:math:`q = w + xi + yj + zk`
+
+where :math:`i, j, k` are basis components with following multiplication rules [2]_:
+
+:math:`ii = jj = kk = ijk = -1`
```
tomvanmele, line 13: use `.. math::` for blocks of math stuff. `:math:` is for inline math formulas...

romanarust:
> Ok, should it be changed in `Rotation` accordingly as well?

yes, correct.

> Also, then I'd expect that in all "axis+angle" methods, the axis is a `Vector` object, right?

yes. This has not been updated yet. Could you make the change?
 
> `Frame.from_quaternion` already exists, however it takes a list as an input.

It should still work if you pass the Quaternion object, since it is iterable. Sure, the docstrings have to be updated. 

> I guess this opens up a bit a larger discussion whether we pass objects or list into constructors?

No, for those objects where it is possible, we should provide both.

src/compas/geometry/transformations/quaternions.py
romanarust: I would go for `compas` wide control, but agree with @funkchaser: different tolerance values for different measures (`compas.ATOL_ANGLE`, ...)

src/compas/geometry/xforms/quaternion.py
```diff
@@ -0,0 +1,225 @@
+
+
```
tomvanmele, line 2: can we remove these?

src/compas/geometry/xforms/quaternion.py
```diff
@@ -0,0 +1,225 @@
...
+        The scalar (real) part of a quaternion.
+    x, y, z : float
+        Components of the vector (complex, imaginary) part of a quaternion.
+
+
+    Examples
+    --------
+    >>> Q = Quaternion(1.0, 1.0, 1.0, 1.0).unitized()
+    >>> R = Quaternion(0.0,-0.1, 0.2,-0.3).unitized()
+    >>> P = R*Q
```
tomvanmele, line 31: in this case there should be whitespace around the operator (it is not a case of operator precedence)

src/compas/geometry/xforms/quaternion.py
```diff
@@ -0,0 +1,225 @@
...
+from compas.geometry.transformations import quaternion_norm
+from compas.geometry.transformations import quaternion_is_unit
+
+from compas.geometry.basic import allclose
+
+__all__ = ['Quaternion']
+
+
+class Quaternion(object):
+
```
tomvanmele, line 16: personal preference, but can we remove the blank line?

src/compas/geometry/xforms/quaternion.py
```diff
@@ -0,0 +1,225 @@
...
+    :math:`y = sin(\\frac{\\theta}{2})  u_{y}`
+
+    :math:`z = sin(\\frac{\\theta}{2})  u_{z}`
+
+    For a quaternion to represent a rotation or orientation, it must be unit-length.
+    A quaternion representing a rotation :math:`p` resulting from applying a rotation :math:`r` to a rotation :math:`q`, i.e.:
+    :math:`p = rq`,
+    is also unit-length.
+
+
```
tomvanmele, line 76: remove blank line?

src/compas/geometry/xforms/quaternion.py
```diff
@@ -0,0 +1,225 @@
...
+    :math:`ij = k,\\qquad ji = -k`
+
+    :math:`jk = i,\\qquad kj = -i`
+
+    :math:`ki = j,\\qquad ik = -j`
+
+    Quaternions are associative but not commutative.
+
+
```
tomvanmele, line 54: remove blank line?

src/compas/geometry/xforms/quaternion.py
```diff
@@ -0,0 +1,225 @@
...
+    References
+    ----------
+    .. [1] http://mathworld.wolfram.com/Quaternion.html
+    .. [2] http://mathworld.wolfram.com/HamiltonsRules.html
+    .. [3] https://github.com/matthew-brett/transforms3d/blob/master/transforms3d/quaternions.py
+
+    """
+
+    def __init__(self, w, x, y, z):
+
```
tomvanmele, line 86: remove blank line?

src/compas/geometry/xforms/quaternion.py
```diff
@@ -0,0 +1,225 @@
...
+        Returns the quaternion as a list of float in the "xyzw" convention.
+        """
+        return [self.x, self.y, self.z, self.w]
+
+    @property
+    def norm(self):
+        """
+        Returns the length (euclidean norm) of the quaternion.
+        """
+        return quaternion_norm(list(self))
```
tomvanmele, line 178: the conversion to list seems unnecessary because the class implements the iterator protocol

src/compas/geometry/xforms/quaternion.py
```diff
@@ -0,0 +1,225 @@
...
+
+    @property
+    def is_unit(self):
+        """
+        Returns ``True`` if the quaternion is unit-length or ``False`` if otherwise.
+        """
+        return quaternion_is_unit(list(self))
```
tomvanmele, line 185: the conversion to list seems unnecessary because the class implements the iterator protocol

src/compas/geometry/xforms/quaternion.py
```diff
@@ -0,0 +1,225 @@
...
+
+    def unitize(self):
+        """
+        Scales the quaternion to make it unit-length.
+        """
+        qu = quaternion_unitize(list(self))
```
tomvanmele, line 191: the conversion to list seems unnecessary because the class implements the iterator protocol

src/compas/geometry/xforms/quaternion.py
```diff
@@ -0,0 +1,225 @@
...
+        self.w, self.x, self.y, self.z = qu
+
+    def unitized(self):
+        """
+        Returns a :obj:`Quaternion` with a unit-length.
+        """
+        qu = quaternion_unitize(list(self))
```
tomvanmele, line 198: the conversion to list seems unnecessary because the class implements the iterator protocol

src/compas/geometry/xforms/quaternion.py
```diff
@@ -0,0 +1,225 @@
...
+        return Quaternion(*qu)
+
+    def canonize(self):
+        """
+        Makes the quaternion canonic.
+        """
+        qc = quaternion_canonize(list(self))
```
tomvanmele, line 205: the conversion to list seems unnecessary because the class implements the iterator protocol

src/compas/geometry/xforms/quaternion.py
```diff
@@ -0,0 +1,225 @@
...
+        self.w, self.x, self.y, self.z = qc
+
+    def canonized(self):
+        """
+        Returns a :obj:`Quaternion` in a canonic form.
+        """
+        qc = quaternion_canonize(list(self))
```
tomvanmele, line 212: the conversion to list seems unnecessary because the class implements the iterator protocol

src/compas/geometry/xforms/quaternion.py
```diff
@@ -0,0 +1,225 @@
...
+        return Quaternion(*qc)
+
+    def conjugate(self):
+        """
+        Returns a conjugate :obj:`Quaternion`.
+        """
+        qc = quaternion_conjugate(list(self))
```
tomvanmele, line 219: the conversion to list seems unnecessary because the class implements the iterator protocol

tomvanmele:
the problems with the master repo have been fixed, so we can now merge.
@romanarust @gonzalocasas okay?
@funkchaser could you still make the last few corrections?

tomvanmele (approved):

PR #299, opened 2019-06-26
Fixes #297: Support install to Rhino with non-admin accounts

### LINKED ISSUE #297: [compas_rhino] import compas error in Rhino under non-admin Windows account
**Describe the bug**

I'm using compas with a **non-admin account** in Windows 10.  `compas` and `compas_rhino` are installed successfully in Anaconda prompt **with admin privilege**. But `compas` cannot be imported in Rhino 6.0.

**To Reproduce**

In Windows 10, I am using a non-admin account. The admin installed Rhino 6.0 for all users. Thus, both users have their own `C:\Users\<user name>\AppData\Roaming\McNeel\Rhinoceros\6.0\Plug-ins\IronPython (814d908a-e25c-493d-97e9-ee3861957f49)` directory. I have the password for the admin account, but I am trying to install and use compas in my own non-admin account.

I can install `compas_rhino` successfully by running the Anaconda prompt with admin privilege:
image

After restarting Rhino 6, I can't import `compas` in Rhino's python editor.
image

**Potential cause for the bug**

The snapshot below shows the different results by running `whoami` in the Anaconda prompt with admin privilege (upper) and without (lower), under my non-admin Windows account.
image

Installing `compas_rhino` requires admin privilege,  which will make the symlink set to the admin's Rhino IronPython but not the in the non-admin account's.

gonzalocasas:
Reworked the install scripts to Rhino to trigger privilege escalation requests from a non-admin user account during the installation itself (without having to start it from an administrator's prompt, which causes the install to always end up in the admin's Rhino user folder).

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Restructure halffaces as lists in `compas.datastructures.VolMesh`.
 - Correctly handle `python-net` module presence during IronPython imports.
 - Switched to `compas.IPY` check instead of `try-except` for preventing non IronPython friendly imports.
+- Changed installation of compas packages to Rhino to support non-admin user accounts on Windows.
 
 ### Removed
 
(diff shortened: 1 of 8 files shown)


gonzalocasas:
@yijiangh @brgcode could you please check this again? It should only pop the UAC (privilege escalation) dialog once and it should also not interfere with the Mac install. I have not tested the plugin install functionality, @brgcode maybe you can give that a shot, just in case?

gonzalocasas:
fixes #297

tomvanmele (approved):

PR #307, opened 2019-07-18
changes for compas_vol

worbit:
this time just a small change, add the new classes (`Cone`, `Cylinder` and `Torus`) as well as the new function (`distance_point_plane_signed`) to the `__all__` list and the `__init__.py` files respectively.

### DIFF
--- a/src/compas/geometry/_primitives/__init__.py
+++ b/src/compas/geometry/_primitives/__init__.py
@@ -16,6 +16,8 @@ from .curve import Bezier
 from .sphere import Sphere
 
 from .box import Box
-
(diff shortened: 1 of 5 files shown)


tomvanmele:
the failing builds can be seen here.

https://travis-ci.com/compas-dev/compas/jobs/217245428

something is wrong with the imports...

worbit:
Yes, I know. Cone.py fails to import Circle. But that's already the case with the current state, without this pull request. I didn't have the new classes in the imports yet, do you think this will fix it? The new class Cylinder also imports Circle, there it seems to work?

PR #322, opened 2019-09-12
first pull for testing

Licini:
1. fixed two test warnings because of python deprecations

2. moved the current test files under `test/compas` folder to mirror the file structure from `src/`, so later we can also have `test/compas_rhino`, `test/compas_blender` etc

1. completed test script for `compas.geometry.bbox` functions


gonzalocasas:
Most model tests are broken now

src/compas/files/urdf.py
```diff
@@ -217,7 +217,8 @@ def get_metadata(type):
     if hasattr(type, 'from_urdf'):
         metadata['from_urdf'] = getattr(type, 'from_urdf')
     else:
-        argspec = inspect.getargspec(type.__init__)
+        # argspec = inspect.getargspec(type.__init__) # this is deprecated in python3
+        argspec = inspect.getfullargspec(type.__init__)
```
gonzalocasas: I think this doesn't work on python 2

gonzalocasas (changes requested):
The change still needs some work, it breaks on python 2 (python death clock is ticking but is not dead yet) ~, and I'm not sure about the moved files, I don't see that you updated the loading of them so I would assume they are not working either (?) I didn't check this thou, I might be wrong.~

src/compas/files/urdf.py
Licini: Hi Gonzalo, thanks for the review. yep, was fully aware we will still use 2.7. Will make change regarding to this part

Licini: Also regarding to moved files,  would you mind to point out what/where is the loading you are referring to? I'm under the impression everything is being detected automatically when put under `test/` folder, and I was able to execute and pass these test after moving(only with 3.7 though). First day working on COMPAS code base, still getting familiar on everything : )

gonzalocasas: You are correct, sorry! I missed the fact that you moved the entirely of the test + fixtures one level deeper.

Licini: added a detector, now it will use old code for python2.7

tomvanmele:
@gonzalocasas all good now?

PR #323, opened 2019-09-14
Sorting vertices' coordinates in MeshViewer()

arpastrana:

### Checklist

### Why this PR?

There is a discordance between the ordering of the vertices coordinates (attribute `_self.xyz`) and the faces keys (attribute `self._faces`) whenever the `mesh` property of `MeshViewer()` is set. 

The most probable reason is that `mesh.get_vertices_attributes('xyz')` in line 91 yields values from the unordered dictionary `self.vertex' from`compas.datastructures.Mesh`.

Hence, it is proposed that the ordering of the vertices' keys is explicitly stated when querying for their `xyz` attributes: `mesh.get_vertices_attributes('xyz', keys=sorted(list(mesh.vertices())`

This bug was not evident when testing against default meshes (such as that derived from `compas.get('faces.obj')`).

### How to reproduce?

The following will work:
```
from compas_viewers import Viewer

viewer.mesh = Mesh.from_obj(compas.get('faces.obj)')
viewer.show() 
```
Screenshot 2019-09-14 at 13 08 23

The following will not work:

```
from compas_viewers import Viewer

viewer.mesh = Mesh.from_obj('my_favorite_mesh.obj')
viewer.show() 
```

Screenshot 2019-09-14 at 13 07 25

### Expected result

Screenshot 2019-09-14 at 13 06 59

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,10 +9,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- Added compas rhino installer for Rhino Mac 6.0 `compas.__init__`.
+- Added compas rhino installer for Rhino Mac 6.0 `compas_rhino.__init__`.
 - Added oriented bounding box for meshes `compas.datastructures.mesh_oriented_bounding_box_numpy`.
 
 ### Changed
+- Fixed unsorted mesh vertex coordinates `xyz` in `compas_viewers.viewer.MeshView`
 - Changed stderr parameter from STDOUT to PIPE in `compas.rpc.Proxy` for Rhino Mac 6.0.
 - Fixed import of `delaunay_from_points` in `Mesh.from_points`.
 - More control over drawing of text labels in Rhino.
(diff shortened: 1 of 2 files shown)


tomvanmele:
should this not be handled with a key_index map instead, because by just sorting the vertices things can still go wrong...

can you send me the obj of the mesh you use for testing?

arpastrana:
Please find it attached below. 

I have roughly tried the `key_index` mapping unsuccessfully in 38009b8. I think there is still an issue with an ordering mismatch between faces and xyz coordinates.

[test_mesh.zip](https://github.com/compas-dev/compas/files/3615355/test_mesh.zip)

arpastrana:
So actually, the `key_index` mapping did work to fix the faces display, but what I was noticing still as a problem was that the `edges` in the property of `MeshView()`were still pointing the _original_ vertex keys thus causing a discrepancy. 

In 838a0d3 a `key_index`mapping for the edges was implemented too. This solved the visualisation problem.

It should be noted that the problem is not evident with meshes created with the method `to_obj()`/ `from_obj`, or others which use `Mesh.to_vertices_and_faces()` / `Mesh.from_vertices_and_faces()`in the background, since the mesh topology is already reconstructed entirely (and indirectly) through a `key_index` dictionary.

PR #328, opened 2019-09-17
Expand test functions for Mesh Class

Licini:
1.expanded test functions to cover the majority of apis from Mesh class

2.Fixed few minor obvious bugs from _mesh.py. 

3.Applied OrderedDict in obj importer and one of mesh api to maintain key order accross py2/3 so they could pass few new testing functions.

(some of new tests will only pass with implementation of OrderedDict this time)
### Checklist


src/compas/datastructures/mesh/_mesh.py
```diff
@@ -1339,7 +1340,7 @@ def is_valid(self):
                 if self.halfedge[u][v] is None and self.halfedge[v][u] is None:
                     return False
                 fkey = self.halfedge[u][v]
-                if fkey:
+                if fkey or fkey == 0:
```
tomvanmele: should this be `if fkey is not None`

gonzalocasas (commented):
LGTM!

Only two questions: 
1. Do we want to leave the empty test functions (e.g. `def test_clear_halfedgedict`) as a reminder that we need to add them? or should we just get rid them them and only add when we really add the test?
2. You mention it's a breaking change, but I am not sure, I thought `OrderedDict` is mostly a drop-in replacement of `dict`, and then only thing that would break is if someone is doing explicit type-checking on this, which they shouldn't anyway.  Or is there's some other breaking behavior on the ordered dict?

src/compas/datastructures/mesh/_mesh.py
```diff
@@ -2617,7 +2618,7 @@ def vertex_curvature(self, vkey):
 
         """
 
-        return 2 * pi - sum([angle_points(mesh.vertex_coordinates(vkey), mesh.vertex_coordinates(u), mesh.vertex_coordinates(v)) for u, v in pairwise(self.vertex_neighbors(vkey, ordered = True) + self.vertex_neighbors(vkey, ordered = True)[:1])])
+        return 2 * pi - sum([angle_points(self.vertex_coordinates(vkey), self.vertex_coordinates(u), self.vertex_coordinates(v)) for u, v in pairwise(self.vertex_neighbors(vkey, ordered = True) + self.vertex_neighbors(vkey, ordered = True)[:1])])
```
tomvanmele, line 2626: we should split up this kind of unreadableness. i think i showed one too many times how great one-liners are and now some of our guys write everything they can that way :)

Licini:
Thanks for review, @gonzalocasas, my thoughts on your questions :
1. I am more intended to leave them as reminders so later people don't have to browse through the source code to find out what to add. (I'm also thinking to make some generator that can create draft test files and just fill in actual testing code for each functions)

2. I guess I got a bit semiological confusion there. It is some of new tests added in will not pass without OrderedDict. After thinking a bit more, it seems to be backwards-compatible instead. Gonna make a change to the note

tomvanmele (approved):
just a few notes

src/compas/datastructures/mesh/_mesh.py
Licini: good point! changed

src/compas/datastructures/mesh/_mesh.py
Licini: lol, this was an impressive one-liner!

gonzalocasas:
@brgcode could we merge this? or any pending remarks?

PR #331, opened 2019-09-18
Handling colinear segments in offsets for polylines and polygons

arpastrana:

### Checklist

### Why this PR?

The current version of `offset_polyline` and `offset_polygon` does not take into account cases where two consecutive segments are collinear. This is because the underlying `intersect_line_line`method will return `None` (as expected) when faced with such challenge. 

However, being able to properly handle these collinear situations would come in quite in handy as they are abundant (for example, to consistently offset the tri/quad/ngon faces of a `Mesh()`), and because, in sum, _colinearity_ does not diminish the _polygonality_ of a `polygon`.

A restructure of `offset_line`, `offset_polyline` and `offset_polygon` is therefore proposed. Shared functionality among them has been encapsulated into their own individual methods. 

Moreover, I have added a few testing functions to the `tests` folder 😄 

### How to reproduce?

The following will work:

```
from compas.geometry import offset_polygon
from compas.geometry import offset_polyline
from compas_plotters import Plotter


polygon = [
	[0.0, 0.0, 0.0],
	[1.0, 0.0, 0.0],
	[1.0, 1.0, 0.0],
	[0.0, 1.0, 0.0],	
]

offset = offset_polygon(polygon, 0.25)

polygons = [{'points': polygon}, {'points': offset, 'facecolor': (240, 240, 240)}]
lines = [{'start': a, 'end': b, 'color': '#00ff00'} for a, b in zip(polygon, offset)]
points_a = [{'pos': xyz, 'radius': 0.015} for xyz in polygon]
points_b = [{'pos': xyz, 'radius': 0.025, 'facecolor': (255, 0, 0)} for xyz in offset]

plotter = Plotter(figsize=(12, 9))
plotter.draw_polygons(polygons)
plotter.draw_lines(lines)
plotter.draw_points(points_a + points_b)
plotter.show()

```

Nevertheless, adding an intermediate vertex to the polygon will not when offsetting:

```
polygon = [
	[0.0, 0.0, 0.0],
	[1.0, 0.0, 0.0],
	[1.0, 1.0, 0.0],
	[0.0, 1.0, 0.0],
	[0.0, 0.5, 0.0]	
]
```

### Expected result

With the proposed PR, the following is successfully achieved:

### DIFF
--- a/src/compas/geometry/offset/offset.py
+++ b/src/compas/geometry/offset/offset.py
@@ -10,6 +10,7 @@ from compas.geometry import cross_vectors
 from compas.geometry import centroid_points
 from compas.geometry import intersection_line_line
 from compas.geometry import normal_polygon
+from compas.geometry import is_colinear
 
 from compas.utilities import pairwise
 
@@ -62,14 +63,11 @@ def offset_line(line, distance, normal=[0.0, 0.0, 1.0]):
         print(line_offset)
 
     """
+
     a, b = line
     ab = subtract_vectors(b, a)
     direction = normalize_vector(cross_vectors(normal, ab))
-
-    if isinstance(distance, (list, tuple)):
-        distances = distance
-    else:
-        distances = [distance, distance]
+    distances = match_value_sequence(distance, line)
 
     u = scale_vector(direction, distances[0])
     v = scale_vector(direction, distances[1])
@@ -78,7 +76,7 @@ def offset_line(line, distance, normal=[0.0, 0.0, 1.0]):
     return c, d
 
 
-def offset_polygon(polygon, distance):
+def offset_polygon(polygon, distance, tol=1e-6):
     """Offset a polygon (closed) by a distance.
 
     Parameters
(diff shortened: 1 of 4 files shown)


src/compas/geometry/offset/offset.py
```diff
@@ -185,33 +168,76 @@ def offset_polyline(polyline, distance, normal=[0.0, 0.0, 1.0]):
...
+        point = intersect(s1, s2, tol)
+        offset.append(point)
+    offset.append(segments[-1][1])
+
+    return offset
+
 
-    if isinstance(distance, (list, tuple)):
-        distances = distance
+def match_value_sequence(value, sequence):
```
tomvanmele, line 183: not sure about the name of this thing, otherwise it looks great.
perhaps this should also become generally available like `pairwise`, `flatten`, ...

src/compas/geometry/offset/offset.py
```diff
@@ -10,6 +10,7 @@
 from compas.geometry import centroid_points
 from compas.geometry import intersection_line_line
 from compas.geometry import normal_polygon
+from compas.geometry import is_colinear
```
tomvanmele, line 13: this is not necessarily related to this PR, but we need to reconsider the naming convention for this kind of functions. after i read the rule (in a PHP context) about `is_` for expected boolean returns, i read something about the need for boolean returns making sense in questions.
in this case `if points_colinear(...)` seems to make a lot more sense than `if is_colinear(...)`

tomvanmele: anyways...

gonzalocasas: > in this case if points_colinear(...) seems to make a lot more sense than if is_colinear(...)

Mmm... I find `if is_colinear()` more readable than the alternative.
C# also has the convention of `is` & `has` prefixing for boolean returns, and it makes a lot of sense, it creates a very clear visual distinction of what a method returns, so clear you don't need to open the docs.

src/compas/geometry/offset/offset.py
arpastrana: How do `match(value, sequence)` or `match_sequence(value, sequence)` in `utilities.itertools_.py` look to you?

tomvanmele: not sure but perhaps it should be thebtopic of a separate PR

PR #333, opened 2019-09-19
Fallback to python executable if pythonw is not present

gonzalocasas:
On a normal case, we use `pythonw` to start the RPC Proxy services and other tasks that also require a CPython interpreter to be executed, but on some Mac installations that executable is not present (it needs to be installed with the `python.app` package), so now we fallback to `python` standard executable if the windowless one is not there. 

### Checklist


Licini:
Looking good to me! Just tested on OS and Windows, all works great!

gonzalocasas:
@brgcode @worbit any comments? or can I go ahead and merge this?

PR #337, opened 2019-09-23
compas_rhino.helpers => compas_rhino.constructors

Licini:
Added compas_rhino.constructor module
moved non-constructor functions from helpers into separate modules like selectors and modifiers

### Checklist


Licini:
@brgcode where you'd like the rest of functions in helpers to go? Would it be something like:

compas_rhino.helpers.mesh.mesh_select_edges => compas_rhino.selectors.mesh_select_edges ?

Licini:
@brgcode another question is how you would like to put the draw functions, should it be called: 
compas_rhino.drawers.mesh_draw_vertices ?

tomvanmele:
> @brgcode where you'd like the rest of functions in helpers to go? Would it be something like:
> 
> compas_rhino.helpers.mesh.mesh_select_edges => compas_rhino.selectors.mesh_select_edges ?

`mesh_select_edges` is just a wrapper around `EdgeSelector.select_edges` so either we remove it or i would add it to the `edgeselector` module as a functional interface to the same functionality...

tomvanmele:
same goes for the draw functions, but then in relation to `MeshArtist`. we should however formalise and clarify how to use `MeshArtist` to draw the vertices of different meshes, and not just of the mesh assigned to the artist at construction time.

```
artist = MeshArtist(mesh1)
artist.draw_vertices()
artist.mesh = mesh2
artist.draw_vertices()
```
or even
```
MeshArtist.draw_vertices(mesh1)
MeshArtist.draw_vertices(mesh2)
```

This is now implemented explicitly in the selector, modifiers, ... by macing them static methods, but quoting Guido, static functions in Python are really silly so we could change all of that and explicitly document how to use the methods of all of these classes as "static" functions.

let me know if that did not make any form of sense to you :)

Licini:
Yep on large scope it all makes lot of sense,  will keep working on this branch and check with you later again

tomvanmele:
perhaps we should also stop requiring that a mesh is assigned at construction time. and simply throw an error if no mesh is availt when one of the drawing functions is called, either as a bound or an unvound method...

tomvanmele:
i will write up a few scenarios...

gonzalocasas:
@brgcode @Licini I think my preferred API would be changing artists to be uniquely linked to one mesh and disable the option to change it, so that there's a fixed relation of 1-artist to 1-mesh. This would remove the slightly tricky statefulness of the current implementation. And even more extreme option would be turn everything to `staticmethods` as you just suggested, but there then we lose a bit of value and the `MeshArtist` are only a namespace for drawing functions.

To summarize, I would propose: 
 - Enforce `mesh` as constructor param (and fail if not set)
 - Remove the setter of the datastructure/mesh, so that is not "possible" to change it
 - As syntactic sugar, add one `staticmethod` to `draw_mesh` (or `draw`) directly, to enable the simple use case of `MeshArtist.draw(mesh)`
 - Document drawing of multiple meshes by clearly showing you need multiple artists to do that.

tomvanmele:
perhaps we should then also split up the current base `Artist` into a true `(Base)Artist` and a `(Base)PrimitiveArtist`. the primitive artist can then be the base for the drawing functionality of all geometric primitives and shapes. currently the data structure artists inherit from `Artist` which also defines all sorts of functionality for drawing primitives. if the data structure artists are really only assigned to one specific data structure and don't draw anything else, we could keep track of the GUIDs of the elements that have been drawn in collections (similar to mpl and the plotters) and identify elements of the data structure by keeping track of the relationship between GUID and data structure element identifier...

tomvanmele:
furthermore, if we make a few small modifications to the `PrimitiveArtist` then primitive objects can also be drawn without any explicit conversion by the user. for example, if we define a line in the line dictionary as a "line" rather then two separate points ("start", "end"), it doesn't matter if you provide a `Line` or a list of two points.

Licini:
@tomvanmele Just making sure if I'm understanding this correctly. So we'll have two type of artists:

1. `PrimitiveArtist` which only used for static methods like `PrimitiveArtist.draw_circles()`,  `PrimitiveArtist.draw_sphere()` etc.

2. `Artist` that bounds to a datastructure in 1-to-1 fashion and cannot be changed, contains functions like `.draw_vertices()` and `.draw_edges()` which draw attributes of of bounded datastructure (assigned at construction time).

Licini:
Hey, @brgcode
Just adjusted again a bit on this part, as you suggested, the selector/modifier wrappers are now moved into corresponding modules instead of keeping as separate files. The discussion about drawing functions regarding to artist is now being worked on a separate PR: https://github.com/compas-dev/compas/pull/349

Licini:
> > @brgcode where you'd like the rest of functions in helpers to go? Would it be something like:
> > compas_rhino.helpers.mesh.mesh_select_edges => compas_rhino.selectors.mesh_select_edges ?
> 
> `mesh_select_edges` is just a wrapper around `EdgeSelector.select_edges` so either we remove it or i would add it to the `edgeselector` module as a functional interface to the same functionality...

Hey, @brgcode
Just adjusted again a bit on this part, as you suggested, the selector/modifier wrappers are now moved into corresponding modules instead of keeping as separate files. The discussion about drawing functions regarding to artist is now being worked on a separate PR: #349

PR #338, opened 2019-09-23
Merge xfroms into transformations

Licini:
create transofrmations.helper_numpy, moved all imports out of functions to top
merge compas.geometry.xforms into compas.geometry.transformations
**TESTs are to be added in this request**

### Checklist

### DIFF
--- a/src/compas/geometry/transformations/__init__.py
+++ b/src/compas/geometry/transformations/__init__.py
@@ -24,5 +24,12 @@ _NEXT_SPEC = [1, 2, 0, 1]
 from .helpers import *
 from .matrices import *
 from .transformations import *
+from .transformation import Transformation
+from .projection import Projection
+from .reflection import Reflection
+from .rotation import Rotation
+from .scale import Scale
+from .shear import Shear
+from .translation import Translation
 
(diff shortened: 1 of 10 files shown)


Licini:
Question for @brgcode :
For migrated functions do we delete them or add deprecation warnings and delete them in a future version?

gonzalocasas:
@Licini i know this is a draft, but I would suggest the files are not re-created, but really moved. Otherwise we lose the git history on them. And regarding the compatibility question, the easiest on the user would be to add aliases on the old location, and add a deprecation warning on the `xforms` package's `__init__` file.

Licini:
@gonzalocasas thanks, that indeed makes more sense. Just to make sure, you mean to move the files, and keep the `__init__` of xforms, just change it to import from new location (and add deprecation warning), right?

gonzalocasas:
> @gonzalocasas thanks, that indeed makes more sense. Just to make sure, you mean to move the files, and keep the `__init__` of xforms, just change it to import from new location (and add deprecation warning), right?

exactly ;)

Licini:
@gonzalocasas sounds good, gonna close this and make a new request

PR #339, opened 2019-09-24
merge xforms to transformations

Licini:
Migrated xforms into transformations
relinked __init__ of xforms to new locations
replaced imports from xforms to transformations project-wise
added deprecated warning to xforms
added full tests for transformations 

### Checklist


tests/compas/geometry/test_tranformations/test_helpers.py
```diff
@@ -0,0 +1,63 @@
+import pytest
```
tomvanmele: the name of this folder (`test_tranformations`) seems to be wrong => `test_transformations`

src/compas/robots/model/link.py
```diff
@@ -4,7 +4,7 @@
 
 from compas.files import URDF
 from compas.files import URDFParser
-from compas.geometry.xforms import Transformation
+from compas.geometry.transformations import Transformation
```
tomvanmele: the other packages in compas should always use the "second level import" convention to allow package developers to move stuff around without affecting the functionality of other packages.

this should have been `from compas.geomtry import Transformation` from the beginning...

tests/compas/geometry/test_tranformations/test_helpers.py
tomvanmele: true for all other test, but will not repeat it...

tomvanmele (commented):
there is a lot more stuff to do for transformations, but the PR does what it says, "merge xforms to transformations"

tests/compas/geometry/test_tranformations/test_helpers.py
Licini: my bad, thats embarrassing...

src/compas/robots/model/link.py
Licini: got it, updated

PR #347, opened 2019-10-03
box properties

worbit:

some more docstrings on the properties of shapes.box.Box and two new properties (area and volume)


gonzalocasas (approved):
LGTM!

PR #348, opened 2019-10-06
File reading function moved to new parent class & pathlib support

tetov:
`BaseReader.read_from_location` returns an iterable for lines in file.

| File format | ASCII | Binary | Skeleton? | BaseReader implemented    |
| ----------- | ----- | ------ | --------- | ------------------------- |
| `amf`       | xml   |        | x         | x                         |
| `dxf`       | x     | x      | x         | x                         |
| `las`       |       | x      | x         | x                         |
| `obj`       | x     |        |           | x                         |
| `off`       | x     |        |           | x                         |
| `ply`       | x     | x      |           | For ASCII                 |
| `stl`       | x     | x      |           | Not yet                   |
| `urdf`     | xml   |        |           | Not yet                   |
| `xml`       | x     |        |           | Needs whole file to parse?|
| `xml_cli`   | x     |        |           | Needs whole file to parse?|

### Checklist


tetov:
I will look into adding the same functionality for binary files, either keep this PR open until that is done or merge this and I'll submit a new PR later on.

**Questions**
* Would you like me to move the tests in `tests/compas/files/test_stl.py` to `tests/compas/datastructures/test_mesh.py`? Could be fitting to do it in this PR.
* Would you like me to put the `BaseReader` class in a file with another name, e.g. `_utils.py`?
* I'm guessing that iterating over lines won't work for the XML reader, am I right in that assumption?

tetov:
I updated questions in above comment. :)

Licini (approved):
LGTM

Licini:
> I will look into adding the same functionality for binary files, either keep this PR open until that is done or merge this and I'll submit a new PR later on.
> 
> **Questions**
> 
> * Would you like me to move the tests in `tests/compas/files/test_stl.py` to `tests/compas/datastructures/test_mesh.py`? Could be fitting to do it in this PR.
> * Would you like me to put the `BaseReader` class in a file with another name, e.g. `_utils.py`?
> * I'm guessing that iterating over lines won't work for the XML reader, am I right in that assumption?

My thoughts on questions : )

- Yep, let's merge the content of test_stl.py into test_from_stl in test_mesh.py.
- It's better to keep as it is for clarity.
- agree, since it's marked-up language.  It would be nice to have an mechanism to throw an error if user tries to load xml in this fashion

tetov:
> * agree, since it's marked-up language.  It would be nice to have an mechanism to throw an error if user tries to load xml in this fashion

There's usually checks in the different modules, e.g. `ply.py` will throw an error if first line is not `'ply'`. It could of course be added to the BaseReader as a general check if you want :).

Licini:
> > * agree, since it's marked-up language.  It would be nice to have an mechanism to throw an error if user tries to load xml in this fashion
> 
> There's usually checks in the different modules, e.g. `ply.py` will throw an error if first line is not `'ply'`. It could of course be added to the BaseReader as a general check if you want :).

That's should be fine then, as long as user be informed the reason of breaking

src/compas/files/off.py
```diff
@@ -128,11 +96,11 @@ def read(self):
...
         """
         if not self.content:
-            return
+            raise RuntimeError('Import failed')
 
         header = next(self.content)
-        if not header.lower() == 'off':
-            return
+        if header.lower() != 'off':
+            raise RuntimeError('Not a valid OFF file')
```
tomvanmele: just out of curiosity, why `RuntimeError`

tetov: > just out of curiosity, why `RuntimeError`

I reread the builtin exception list and I see now that I doesn't really fit.

TypeError instead?

src/compas/files/_BaseReader.py
```diff
@@ -0,0 +1,69 @@
...
+    Attributes
+    ----------
+    location : str or pathlib object
+        Path or URL to the file
+    """
+
+    def __init__(self, location):
+        self.location = location
+
+    def read_from_location(self):
```
gonzalocasas: Maybe this names should be `def read(mode)` instead of `read_from_location` and `read_binary_from_location`...?

src/compas/files/_BaseReader.py
```diff
@@ -0,0 +1,69 @@
...
+
+        Returns
+        -------
+        iterable
+            Iterable returning file line by line
+        """
+
+        if isinstance(self.location, Path):
+            with self.location.open(mode='r') as fh:
+                content = iter(fh.readlines())
```
gonzalocasas: Instead of reading the whole content and creating an iterator around it, we should do something like:
```python
with self.location.open(mode='r') as fh:
    for line in fh:
        yield line
```

src/compas/files/_BaseReader.py
```diff
@@ -0,0 +1,69 @@
...
+        elif self.location.startswith('http'):
+            resp = urlopen(self.location)
+            content = iter(resp.read().decode('utf-8').split('\n'))
+        else:
+            with open(self.location, 'r') as fh:
+                content = iter(fh.readlines())
```
gonzalocasas: same comment as above... ;)

src/compas/files/_BaseReader.py
```diff
@@ -0,0 +1,69 @@
...
+        Returns
+        -------
+        iterable
+            Iterable reading file
+
+        """
+
+        if self.location.isinstance(Path):
+            with self.location.open(mode='r') as fh:
+                content = iter(partial(fh.read, 64), b'')
```
gonzalocasas: is `64` the block size? perhaps it should be an argument, with a default value.

src/compas/files/off.py
tomvanmele: `TypeError` or `ValueError`

tomvanmele: i think there are quite a few other occurrences...

tetov: I know, I might have used the wrong one but at least I was consistently wrong ;).

tetov:
I will change `RuntimeError` to `TypeError` or `ValueError` depending on occurence in all files and move the STL tests to `test_mesh.py` tonight.

I think this should be merged without a new binary reader, and I'll open another PR to start work on that.

src/compas/files/off.py
gonzalocasas: Just `Exception` if you're in doubt. TypeError is not the right one either, because it's not related to the type system.

src/compas/files/ply.py
```diff
@@ -138,106 +141,98 @@ def read_header(self):
...
-
-            line = file.readline().rstrip()
-
-            if line != 'ply':
-                raise Exception('not a valid ply file')
-
-            self.start_header = file.tell()
-
-            element_type = None
+        if next(self.content).rstrip() != 'ply':
```
gonzalocasas:
```suggestion
        line = next(self.content)

        if line.rstrip() != 'ply':
```

It's better not to mix a conditional with a call that has side effects (such as `next()`) because it might cause bugs when code is refactored at a later time.

gonzalocasas (changes requested):
I like this change! It definitely goes in the right direction of removing code duplication.
Regarding the XML question, it won't really work, we would need a event-based parser like sax or similar, but it's a bit meh....

The main thing I would ask to fix before merging is that now the files are loaded in its entirety and then iterated over lines. This has much higher memory consumption than iterating over the lines directly off of the file object.

You could test this pretty quickly by loading a very large file and printing doing something like, print each line. If the behavior is an iterator, the loading will be almost no time, and then it will start printing one by one; but if the behavior is to load everything up front, you will see a delay before it starts printing.  The alternative is to profile this code when loading large files.

But overall, good stuff!! Keep up the good work!

src/compas/files/_BaseReader.py
tetov: Ok!

src/compas/files/_BaseReader.py
tetov: Hm, I thought this would do that.. I'll change it!

src/compas/files/_BaseReader.py
tetov: Yup, it should. This part is not done and should maybe not be in this PR. I'll merge the functions, add an if check regarding mode and raise NotImplementedError for binary.

src/compas/files/_BaseReader.py
tetov: All the file type readers have a read function. Would you still like to use `read` for `BaseReader` and use `read_EXT` (where EXT=file extension)? (See `ply.py` or `obj.py` for good examples)

src/compas/files/_BaseReader.py
tomvanmele: no, it converts the list to an iterator, but first makes the list...

src/compas/files/_BaseReader.py
tomvanmele: does this work if the file is on a remote location? or does the comment only apply to the `else` part?

src/compas/files/_BaseReader.py
tetov: Got it now, sorry for the mistake :).

tomvanmele:
also, perhaps for `dxf` we could build upon `ezdxf`...

tetov:
I'll wait to commit new files when questions raised in @gonzalocasas review and following comments are figured out.

The BaseReader should as pointed out above just yield lines, but I'll also do some profiling to see how much and when the specific modules use up memory.

src/compas/files/ply.py
```diff
@@ -401,33 +414,32 @@ def read_faces_binary(self):
...
+        self.faces = None
         self.parse()
 
     def parse(self):
-        self.vertices = [(vertex['x'], vertex['y'], vertex['z']) for vertex in self.reader.vertices]
-        self.faces = [face['vertex_indices'] for face in self.reader.faces]
-
+        self.vertices = [(vertex['x'], vertex['y'], vertex['z'])
+                         for vertex in self.reader.vertices]
+        self.faces = [face[self.reader.face_properties[0][0]] for face in self.reader.faces]
```
tetov, line 429: This change is because while the definition says `vertex_indices ` Rhino uses `vertex_index ` instead.

tetov:
Per IRL discussion with @brgcode, binary ply has not been implemented, so that test will be commented out.

The code needs to be cleaned up, travis errors fixed but it's slowly getting there.

src/compas/files/_BaseReader.py
tetov: @gonzalocasas: what Tom said ☝️

tetov:
Oh and sorry about the amount of commits. Didn’t know a rebase would show up like that on GitHub. I’ll merge next time.

tomvanmele:
@gonzalocasas can we merge this?

tetov:
@gonzalocasas, @brgcode 
https://github.com/compas-dev/compas/pull/348#discussion_r335163659

.bumpversion.cfg
```diff
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 0.8.1
+current_version = 0.9.0
```
gonzalocasas, line 2: Why is it show this weird diff? Maybe you need to merge master back into your branch?

CHANGELOG.md
```diff
@@ -7,10 +7,34 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Changed
```
gonzalocasas, line 10: Watch out here, this one also needs to be updated.

docs/conf.py
```diff
@@ -19,7 +19,7 @@
 copyright        = 'Block Research Group - ETH Zurich'
 author           = 'Tom Van Mele'
 
-release = '0.8.1'
+release = '0.9.0'
```
gonzalocasas, line 22: same

setup.py
```diff
@@ -28,7 +28,7 @@ def read(*names, **kwargs):
 
 setup(
     name='COMPAS',
-    version='0.8.1',
+    version='0.9.0',
```
gonzalocasas, line 31: same

src/compas/__init__.py
```diff
@@ -35,7 +35,7 @@
 __license__   = 'MIT License'
 __email__     = 'vanmelet@ethz.ch'
 
-__version__ = '0.8.1'
+__version__ = '0.9.0'
```
gonzalocasas, line 38: and same here

src/compas/__init__.py
```diff
@@ -192,7 +184,7 @@ def get(filename):
     This folder is only locally available if you installed **COMPAS** from source,
     or if you are working directly with the source.
     In all other cases, the function will get the corresponding files direcly from
-    the GitHub repo, at https://raw.githubusercontent.com/compas-dev/compas/master/data
+    the GitHub repo, at https://github.com/compas-dev/compas/tree/master/data
```
gonzalocasas, line 187: Why is this changed? I think we're still pull from `raw.githubusercontent.com`...

src/compas/files/_BaseReader.py
```diff
@@ -0,0 +1,104 @@
...
+class BaseReader(object):
+    """Base class containing file reading functions for file extension specific readers
+
+    Attributes
+    ----------
+    location : str or pathlib object
+        Path or URL to the file
+    """
+    def __init__(self, location):
+        self.is_URL = False
```
gonzalocasas, line 23: this should be all lowercase:

```suggestion
        self.is_url = False
```

src/compas/files/_BaseReader.py
```diff
@@ -0,0 +1,104 @@
...
+        self.location = self.check_location(location)
+
+    def read(self, mode='text'):
+        """Returns iterable reading file line by line when given a location,
+           either as pathlib object, string containing path or string containing url
+           TODO: Update docstring
```
gonzalocasas, line 29: Don't leave this TODO in the docstring, it will be shown in the pubic docs.

src/compas/files/_BaseReader.py
```diff
@@ -0,0 +1,104 @@
...
+                            yield line
+
+        elif mode == 'binary':
+            raise NotImplementedError
+
+        else:
+            raise ValueError('mode {} not recognized'.format(mode))
+
+    def check_location(self, location):
+        """ TODO: Docstring
```
gonzalocasas, line 66: same note here about `TODO` in docstring.

src/compas/files/_BaseReader.py
```diff
@@ -0,0 +1,104 @@
+from __future__ import absolute_import
```
gonzalocasas, line 1: module names should be lower case, and in doing so, you can remove the leading underscore that was probably causing a name collision with your class.

src/compas/files/_BaseReader.py
```diff
@@ -0,0 +1,104 @@
...
+            pathobj = Path(location)
+        else:
+            pathobj = location
+
+        if pathobj.exists():
+            return pathobj
+        else:
+            raise IOError('File not found.')
+
+    def check_file_signature(self):
```
gonzalocasas, line 81: This method is great stuff, but I would make a few changes for clarity, because it took me a while to understand what you're doing here:
1. `self.file_signature` is a constant per-class, so I would make that uppercase, eg `self.FILE_SIGNATURE`.
2. Docs! This is a pretty significant addition to the public API, so, it really needs to be well documented.
3. The first check cannot be passed, there are 3 options: 1) the support for urls is implemented 2) not implemented and raises exception, 3) not implemented, but add a `return` immediately.
4. Try to avoid nesting checks. The `else` that contains everything is not needed, because the previous code paths end in error (at least, they should, see previous point)

Then, two questions:
1. Does this work with ascii files (since you open the file in binary mode)?
2. Would it be possible to use `urlretrieve` and actually remove al the remote stuff to be a one-time download into temp and handle everything as files?

src/compas/files/amf.py
```diff
@@ -2,30 +2,30 @@
 from __future__ import absolute_import
 from __future__ import division
 
+from compas.files._BaseReader import BaseReader
```
gonzalocasas, line 5: After the previously suggested rename:
```suggestion
from compas.files.base_reader import BaseReader
```

src/compas/files/obj.py
```diff
@@ -273,8 +242,8 @@ def _read_freeform_geometry(self, name, data):
             if self.deg[0] == 1:
                 if len(data) == 4:
                     self.lines.append((int(data[2]) - 1, int(data[3]) - 1))
-                    if self.group:
-                        self.groups[self.group].append(('l', len(self.lines) - 1))
+                if self.group:
```
gonzalocasas, line 245: I don't understand this one... is it really correct?

src/compas/files/off.py
```diff
@@ -114,7 +87,7 @@ def pre(self):
     def post(self):
         pass
 
-    def read(self):
+    def read_off(self):
```
gonzalocasas, line 90: I'm not very convinced about calling these `read_off()`, `read_obj()` and so on. I think it should be the same name across the board. I see that now `read()` is used, so, either rename this level of reads to `read_content` or `read_file`, or rename the `BaseReader.read` to `BaseReader.read_bytes` (or `BaseReader.read_line` or `BaseReader.read_chunk`)

src/compas/files/off.py
```diff
@@ -128,11 +101,7 @@ def read(self):
 
         """
         if not self.content:
-            return
-
-        header = next(self.content)
-        if not header.lower() == 'off':
-            return
+            raise Exception('Import failed')
```
gonzalocasas, line 104: `Import failed` is not very clear, we're not really importing, we're opening files (which will then might be imported into something), so I would be more explicit and limited to this module's functionality in the error message.

src/compas/files/ply.py
```diff
@@ -122,12 +130,14 @@ def is_ascii(self):
...
 
-    def read(self):
+    def read_ply(self):
         self.read_header()
         if self.format == 'ascii':
             self.read_data()
         else:
-            self.read_data_binary()
+            # Until we have an implentation with or without numpy
+            raise NotImplementedError
```
gonzalocasas, line 139: Is that really a good option to do now? If I understand correctly, binary works before this change, but without numpy so it might be slow; but with this change, it doesn't work anymore. Am I correct?

src/compas/files/ply.py
```diff
@@ -406,33 +415,31 @@ def read_faces_binary(self):
...
     import compas
-
     from compas.datastructures import Mesh
 
     ply = PLY(compas.get_bunny())
 
     mesh = Mesh.from_vertices_and_faces(ply.parser.vertices, ply.parser.faces)
 
-    print(mesh.summary())
+    print(mesh.summry())
```
gonzalocasas, line 445: hehe.. is this method really still with a typo on compas Mesh?

src/compas/geometry/basic.py
```diff
@@ -59,6 +60,27 @@
 ]
 
 
+def close(value1, value2, tol=1e-05):
```
gonzalocasas, line 63: This commit actually comes from the rebasing things, right? It's not really part of your branch..?

tests/compas/files/test_stl.py
```diff
@@ -1,36 +0,0 @@
-import os
```
gonzalocasas, line 1: Is this test now done somewhere else?

gonzalocasas (changes requested):
I tried to review this in detail, but the rebase thing gets a lot in the way. There are a few things that are not clear to me, some naming things that should change before merging, the removal of tests, and other remarks; but really, the length of the current PR makes it tricky to review.  If merging master back into this branch does not get rid of all the spurious diffs, then I would suggest you create a new clean branch, and cherry pick your changes on top of a clean status.

tetov:
I’ll open a fresh one (and go through the comments).

src/compas/__init__.py
tetov: I’ll have a think about why I did that. I agree but I think I had a reason.

src/compas/files/_BaseReader.py
tetov: Leading underscore to only make it available inside the module. But maybe not needed with our __init__.py?

src/compas/files/_BaseReader.py
tetov: Questions:
1. Yes!
2. This goes against what I thought you wanted, to not load a big file into memory?

src/compas/files/obj.py
tetov: Doublecheck.

src/compas/files/ply.py
tetov: Binary ply doesn’t work at all. Double checked with Tom.

tetov: Would be implemented hopefully in my binary PR.

src/compas/geometry/basic.py
tetov: Yup.

tests/compas/files/test_stl.py
tetov: Yes. Moved to test_mesh.py under tests/compas/datastructures.

tetov:
I replied to some. The ones I didn’t reply to I agree to completly. 

I’ll move over your comments to new PR. Thanks for reviewing.

tests/compas/files/test_stl.py
gonzalocasas: ok! I actually think it is more proper to keep this in a `compas/files/test_stl` because it's supposed to be unit test, and by testing the mesh_from_stl, you're actually already integrating a lot of other things; but in this case, it's ok if you want to keep it in the new location, it's not such a big test anyway.. the only thing I would suggest thou is to split the `test_from_stl` method into different ones, each describing what they are actually testing (now it's only a comment but if it fails, you dont see the comment text in the test failure)

src/compas/files/ply.py
gonzalocasas: Ah! alright! yes, now I remember we talked about it at some point (like 3 lifetimes ago!).

src/compas/files/_BaseReader.py
gonzalocasas: ah ok.. yeah, I think it won't be automatically pulled by the `__init__`, so you could get rid of the leading underscore, but it's also not a big thing if you leave it.

src/compas/__init__.py
gonzalocasas: Ah.. I think I see why. If a user reads this and tries to open that URL, he/she gets a 404, while with your change, would get the directory listing, which is what one would expect, so it's good, leave it as you have.

src/compas/files/_BaseReader.py
tetov: Moved to #418

src/compas/files/obj.py
tetov: Moved to #418.

src/compas/files/off.py
tetov: Moved to #418.

src/compas/files/off.py
tetov: Moved to #418.

src/compas/files/ply.py
tetov: Hm, we might need to implement mesh.summry().

```
def summry():
    print('No it's rather winntry at the moment')
```

tests/compas/files/test_stl.py
tetov: Moved to #418.

tetov (commented):
All unresolved issues moved to #418.

PR #349, opened 2019-10-07
Artist for everything

Licini:
1. Moved all primitive drawing functions from `compas_rhino.artists.Artist` to a newly created `compas_rhino.artists.PrimitiveArtist` as static functions, new APIs would be like:
```
from compas_rhino.artists import PrimitiveArtist

PrimitiveArtist.draw_lines(...)
PrimitiveArtist.draw_circles(...)
```

2. Added datastructure type check when constructing Mesh/Volmesh/Network artists

### Checklist

   1. [x] Add documentations


Licini:
Will add corresponding docs and tests if this looks correct to you guys @tomvanmele @gonzalocasas

tomvanmele:
we should discuss this further during the next meeting and map out a clear plan.
* artists dedicated to specific objects
* what about selectors and modifiers? perhaps we should get rid of all the staticness and just use functions? perhaps this should be a general rule for the framework?
* what about the relation to the base CAD object?
* ...

Licini:
> we should discuss this further during the next meeting and map out a clear plan.
> 
> * artists dedicated to specific objects
> * what about selectors and modifiers? perhaps we should get rid of all the staticness and just use functions? perhaps this should be a general rule for the framework?
> * what about the relation to the base CAD object?
> * ...

Sounds good. have lot of thoughts on this

tomvanmele:
was working on the plotters and i am more and more convinced (also after reading some more about the new OO approach of mpl) that we should go for something along the lines of what you proposed already a while ago.

for example, for a line

* `LineObject` as the container for functionality
* `LineObject.geometry` is basically an instance of `compas.geometry.Line`
* `LineObject.attributes` a dict with visualisation settings and whatever else makes sense
* `LineObject.artist` an instance of the relevant artist.

in  case of the plotters, `LineObject.artist` would be an instance of `compas_plotters.artists.LineArtist`.

to let the artist know where to plot, an instance of a `Plotter` would have to be assigned to the artist at some point so it can use the correct set of axes to add stuff to (well or actually it can just use `gca`, but a plotter would still have to be created to control the setup of those axes).

```python
from compas.objects import LineObject
from compas_plotters import Plotter
from compas_plotters.artists import LineArtist

plotter = Plotter()
line = LineObject(Line(start, end), LineArtist())

...
```

gonzalocasas:
But this means we need to create additional "sibling" classes for all classes that are supported, right? Why not embed this into the existing geometry object (which already *are* themselves and already contain `attributes`) and artists already are bundling a geometry object with a rendering object (the artist).

Having to create a separate layer as "transport for visualization" does not feel very correct, especially because this would not be an internal implementation detail, but it pushes accidental complexity onto the users.

tomvanmele:
true that actually makes more sense :)
so something like this

```python
from compas_plotters import Plotter

from compas_plotter.artists import LineArtist

plotter = Plotter()

line = Line(start, end)
artist = LineArtist(line)
artist.draw()
```

```python
from compas_rhino.artists import LineArtist

line = Line(start, end)
artist = LineArtist(line)
artist.draw()
```

tomvanmele:
`MeshPlotter` and `NetworkPlotter` would become `MeshArtist` and `NetworkArtist` in `compas_plotters`

tomvanmele:
and perhaps `compas_plotters` should actually be `compas_plotter`

tomvanmele:
perhaps it would also make sense to differentiate between artists for individual elements and artists for element collections

`PointArtist` vs `PointCollectionArtist`
`LineArtist` vs `LineCollectionArtist`

...

tomvanmele:
all of these could derive from a base artist class that is defined in `compas` (in `compas.base`, `compas.abc`, `compas.xxx`?) and that has to be implemented in the various viz packages `compas_plotter`, `compas_viewer`, `compas_rhino`, ...

tomvanmele:
there is first proposal for the new plotters on the `new-plotters` branch!
let me know what you think...

tomvanmele:
```python
from compas_plotters import Plotter2
from compas_plotters import PointArtist

plotter = Plotter2(view=([0, 16], [0, 10]), size=(8, 5), bgcolor='#cccccc')

PointArtist.plotter = plotter

a = Point(1.0, 1.0)
b = Point(9.0, 5.0)
c = Point(9.0, 1.0)

a_artist = PointArtist(a)
b_artist = PointArtist(b)
c_artist = PointArtist(c)

plotter.add_artist(a_artist)
plotter.add_artist(b_artist)
plotter.add_artist(c_artist)

plotter.draw(pause=1.0)

for i in range(10):
    a[0] += 0.5
    plotter.redraw(pause=0.1)

plotter.show()
```

gonzalocasas:
That snippet opens up several questions (disclaimer: just based on this, I haven't checked your branch's code yet):

- `PointArtist.plotter = plotter`: assigning a plotter instance to a class-level variable of artist is not a good idea. Mixes class and instance life-cycles and makes it impossible to do threaded plotting (if that would eventually be possible)
- Adding plotter to artist (`PointArtist.plotter = plotter`)  and also artists to plotter (`plotter.add_artist(a_artist)`) points to a circular dependency.
- Should `Plotter` (or `Plotter2`) actually be called `Plot`? Or `Canvas`? It looks like now it is an individual drawing surface, not an api to the underlying plotter tool (although, that's a fine semantic line)

tomvanmele:
understood. will make a few changes, because most of that just grew that way but is not strictly necessary anymore...

tomvanmele:
```python
from compas_plotters import Plotter2
from compas_plotters import PointArtist

plotter = Plotter2(view=([0, 16], [0, 10]), size=(8, 5), bgcolor='#cccccc')

a = Point(1.0, 1.0)
b = Point(9.0, 5.0)
c = Point(9.0, 1.0)

a_artist = PointArtist(a)
b_artist = PointArtist(b)
c_artist = PointArtist(c)

plotter.add_artists([a_artist, b_artist, c_artist])

plotter.draw(pause=1.0)

for i in range(10):
    a[0] += 0.5
    plotter.redraw(pause=0.01)

plotter.show()

```

tomvanmele:
```python
from compas_plotters import Plotter2
from compas_plotters import PointArtist
from compas_plotters import LineArtist

plotter = Plotter2(view=([0, 16], [0, 10]), size=(8, 5), bgcolor='#cccccc')

a = Point(1.0, 0.0)
b = Point(3.0, 2.0)
line = Line(a, b)

a_artist = PointArtist(a)
b_artist = PointArtist(b)
line_artist = LineArtist(line)

plotter.add_artists([a_artist, b_artist, line_artist])

plotter.draw(pause=1.0)
for i in range(10):
    a[0] += 0.5
    line.start[0] += 0.5
    plotter.redraw(pause=0.01)

plotter.show()
```

tomvanmele:
it is still an API to the underlying mpl functionality, though. albeit one that is a bit more friendly to our primitives and shapes and is more in line with the CAD artists.

just addressed the individual base primitives. there should be artists for point collections, line collections, ... and mesh and network artists as well.

gonzalocasas:
looks better, feels more natural. If there's a 1-to-1 mapping between classes and artists, then wouldn't theoretically be redundant to specify the artist class when creating it, i.e. the construction of artists could be replaced by a factory pattern of some kind (and perhaps even done for the user behind the scenes?)

tomvanmele:
yes indeed. something that check for the type or  an `add_point`, `add_line`, ... method?

i think it would be a good idea to write out a few of these snippets for the CAD packages as well and look at them side by side to see where we can find a natural common API...

gonzalocasas:
I'm "writing out loud" here, but how about:

```python
from compas_plotters import Plotter2
from compas_plotters import PointArtist
from compas_plotters import LineArtist

plotter = Plotter2(view=([0, 16], [0, 10]), size=(8, 5), bgcolor='#cccccc')

a = Point(1.0, 0.0)
b = Point(3.0, 2.0)
line = Line(a, b)

plotter.add(a)
plotter.add(b)
plotter.add(line)

plotter.draw(pause=1.0)
for i in range(10):
    a[0] += 0.5
    line.start[0] += 0.5
    plotter.redraw(pause=0.01)

plotter.show()
```

`Plotter2` would have:

```python
class Plotter2(object):
    # ... other stuff ...
    def add(self, item, artist=None):
        # the artist arg would allow to override default behavior
        # in case one wants to use a custom artist
        if not artist:
            artist = Artists.build(item, type(self))

        artist.draw(item)
        return artist
```

and `Artists` would do something like:

```python
_ARTIST_TYPE_PER_FRONTEND = {}

class Artists(object):
    @staticmethod
    def register(frontend_type, item_type, artist_type):
        _ARTIST_TYPE_PER_FRONTEND[frontend_type][item_type] = artist_type

    @staticmethod
    def build(item, frontend_type, **kwargs):
        artists = _ARTIST_TYPE_PER_FRONTEND[frontend_type]
        artist_type = artists[type(item)]
        artist = artist_type(**kwargs)
        return artist
```

finally, the root `__init__` of each CAD (or plotters) package could take care of registering the known artists in that package.

tomvanmele:
yup that could work, but then we would make some kind of `Rhino` object equivalent to `Plotter`?

tomvanmele:
would the factory method not be part of the artist base class?

gonzalocasas:
> would the factory method not be part of the artist base class?

yep, it could be there

CHANGELOG.md
```diff
@@ -9,10 +9,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 ### Added
 
+- Added `compas_rhino.artists.PrimitiveArtist` for visualising primitives
+- Added datastructure type check when constructing Mesh/Volmesh/Network artists
+
 ### Changed
 
 ### Removed
 
+- Deleted primitive drawing functions from `compas_rhino.artists.Artist`
```
tomvanmele, line 31: we need to bump the version for this because it is a backwards incompatible change.
always wondered, do we do this in advance, afterwards or as part of the PR?

src/compas_rhino/artists/artist.py
```diff
@@ -18,7 +18,7 @@
 
 
 class Artist(object):
-    """The base ``Artist`` defines functionality for drawing geometric primitives in Rhino.
+    """The base ``Artist`` defines functionality for drawing a bounded datastructure in Rhino.
```
tomvanmele, line 21: should we then not create a base artist called `Artist` and then do

* `Artist` > `PrimitiveArtist` > `LineArtist`, ...
* `Artist` > `DataStructureArtist` > `MeshArtist`, ...

Licini:
Hi @gonzalocasas, Just had a discussion with @brgcode , since this branch is a bit old and the discussion presents few possible directions( I also have some uncommitted ideas, very similar to you guys' snippets but not settled in every details),
we are gonna merge the initial changes of this branch which only involves migrating static drawing functions to the new PrimitiveArtist class and some type checks for mesh/volmesh/network artist. 
And continue the discussion with two new PRs on Plotter and compas_rhino artists until we con-solidify on a uniformed API set across platforms

src/compas_rhino/artists/artist.py
Licini: mmmm.....agree, to be honest I feel that's a much better idea

tomvanmele:
perhaps the static methods can always be used to draw collections of things rather than the specific object bound to the artist.

but then would it be better that all these collection methods are made available through the `PrimitiveArtist` or by the respective dedicated artists?

* `PointArtist` has a static method to draw a collection of points rather than just the one point that was bound to it.
* `LineArtist` has a static method for drawing a collection of lines rather than just the one line that was bound to it.
* ...

for its one dedicated primitive (or data structure) each artist provides a lot of additional visualisation functionality. for example, for a line you can choose to visualise just the line but also  the start and end point, the segment between the two, and so on.

alternatively, there could be dedicated artists for collections.

* `PointArtist` vs `PointCollectionArtist`
* `LineArtist` vs `LineCollectionArtist`
* ...
* `MeshArtist` vs `MeshCollectionArtist`
* ...

Licini:
Well, after more discussion this part will be reworked in a more atomic fashion into something like:

- Artist > PrimitiveArtist > LineArtist, ...
- Artist > DataStructureArtist > MeshArtist, ...

And static drawing functions will be put into corresponding artists, current drawing functions under artist will be assigned deprecation warning before a new version

PR #364, opened 2019-10-20
Mesh additions

worbit:

Adds some functionality to the mesh class (`to_obj_with_attribute`, attributes as obj groups) and to ghpython mesh display (`draw_mesh`: compute normals if none are given). Optionally, this method could call `mesh.Compact()` before returning, see [here](https://developer.rhino3d.com/api/RhinoCommon/html/M_Rhino_Geometry_Mesh_Compact.htm):

> Removes any unreferenced objects from arrays, reindexes as needed and shrinks arrays to minimum required size.

### DIFF
--- a/src/compas/datastructures/mesh/_mesh.py
+++ b/src/compas/datastructures/mesh/_mesh.py
@@ -824,6 +824,40 @@ class Mesh(FromToPickle,
                 vertices = [key_index[key] + 1 for key in vertices]
                 fh.write(' '.join(['f'] + [str(index) for index in vertices]) + '\n')
 
+    def to_obj_with_attribute(self, filepath, attribute):
+        """
+        Write the mesh to an OBJ file grouping the faces by attribute.
+
+        Parameters
+        ----------
+        filepath: str
+            Full path of the file.
+        attribute: str
+            The name of the faces' attribute to group by.
+        """
+        # if none of the faces has the attribute, export normally
+        if not any(self.get_faces_attribute(self.faces(), attribute)):
+            self.to_obj(filepath)
+            return
+
+        fkeys = list(self.faces())
+        fkeys.sort(key=lambda x: str(self.get_face_attribute(x, attribute)))
+
+        key_index = self.key_index()
+        current_group = 'na'
+        with open(filepath, 'w+') as fh:
+            for key, attr in self.vertices(True):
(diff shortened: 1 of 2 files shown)


src/compas/datastructures/mesh/_mesh.py
```diff
@@ -824,6 +824,40 @@ def to_obj(self, filepath):
                 vertices = [key_index[key] + 1 for key in vertices]
                 fh.write(' '.join(['f'] + [str(index) for index in vertices]) + '\n')
 
+    def to_obj_with_attribute(self, filepath, attribute):
```
tomvanmele, line 827: name doesn't really says what it does in my opinion.
also, should this not be part of the `OBJ` API?
then there could just be a flag in the `to_obj` function of the mesh to request "grouping by attribute"...

worbit: i considered adding this functionality to the existing `to_obj` function, but i thought this should maybe stay simple and lightweight. with OBJ API, you mean `compas.files.obj.OBJ`?

tomvanmele: i more meant that the entire implementation should go to `compas.files.OBJ`, which is actually also true for the current `to_obj` functionality. the mesh just needs to provide the `OBJ` with data it understands after which it an do its thing with it...

now the mesh also takes care of the actual writing of the OBJ, which should not be the case in either of the two functions...

worbit: ok I see. I didn't know `to_obj` was supposed to be moved anyway and therefore built on top of it. but having an `OBJ` object take care of file writing makes sense, optionally with a flag for face grouping! Currently, there are the classes `OBJ`, `OBJReader` and `OBJParser`. Would there be a new class `OBJWriter` then?

tetov: This could of course also be setup for all the different `to_*` methods, preferably using a common superclass named e.g. BaseWriter..

tomvanmele: agreed. it was always the goal to have writers as well. i just never got around to it...

tetov: I'll add that to my list of things to do. (After current PR and refactor of binary read)

tomvanmele:
can i close this? seems this will get resolved during the general files overhaul...

PR #365, opened 2019-10-20
Meshfactory: mesh from shape

worbit:

### Checklist

As discussed in a previous dev meeting: the `Mesh` class gets a new method `from_shape`, similar to `from_polyhedron`. This method calls the `to_vertices_and_faces` method on the `Shape` instance in question.

### DIFF
--- a/src/compas/datastructures/mesh/_mesh.py
+++ b/src/compas/datastructures/mesh/_mesh.py
@@ -40,9 +40,6 @@ from compas.geometry import angle_points
 from compas.geometry import bestfit_plane
 from compas.geometry import distance_point_plane
 from compas.geometry import distance_point_point
-from compas.geometry import transform_points
-from compas.geometry import Frame
-from compas.geometry import matrix_from_frame
 
 from compas.datastructures import Datastructure
 
@@ -743,7 +740,7 @@ class Mesh(FromToPickle,
         u : int
             The resolution in u direction (Torus, Sphere, Cylinder, Cone).
         v : int
-            The resolution in v direction (Sphere only).
+            The resolution in v direction (Torus and Sphere only).
 
         Returns
         -------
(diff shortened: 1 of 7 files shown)


src/compas/datastructures/mesh/_mesh.py
```diff
@@ -729,6 +730,43 @@ def from_polyhedron(cls, f):
         p = Polyhedron.generate(f)
         return cls.from_vertices_and_faces(p.vertices, p.faces)
 
+    @classmethod
+    def from_shape(cls, s, **kwargs):
```
tomvanmele: why not call it `shape` instead of `s`?

src/compas/datastructures/mesh/_mesh.py
```diff
@@ -729,6 +730,43 @@ def from_polyhedron(cls, f):
...
+
+            import compas
+            from compas.geometry import Torus
+            from compas.datastructures import Mesh
+
+            t = Torus(((0, 0, 0), (0, 0, 1)), 2, 1)
+            mesh = Mesh.from_shape(t)
+
+        """
+        if not isinstance(s, Shape):
```
tomvanmele: perhaps this is not even necessary. you could just do
```python
vertices, faces = shape.to_vertices_and_faces(**kwargs)
```
as long as that doesn't fail, who cares if the input was actually a `Shape`...

src/compas/geometry/primitives/shapes/box.py
```diff
@@ -378,6 +375,10 @@ def to_data(self):
         """
         return self.data
 
+    def to_vertices_and_faces(self, **kwargs):
+        """Returns a list of vertices and faces, called by `Mesh.from_shape()`."""
```
tomvanmele: would not specify that this will be "called by `Mesh.from_shape()`", as it might not be, nor does it need to be to be useful...

src/compas/geometry/primitives/shapes/cone.py
```diff
@@ -167,6 +171,35 @@ def volume(self):
         """Float: The volume of the cone."""
         return pi * self.circle.radius**2 * (self.height / 3)
 
+    def to_vertices_and_faces(self, **kwargs):
+        """Returns a list of vertices and faces, called by `Mesh.from_shape()`."""
```
tomvanmele: same as above

src/compas/geometry/primitives/shapes/cone.py
```diff
@@ -167,6 +171,35 @@ def volume(self):
...
+        if 'u' in kwargs:
```
gonzalocasas: you could simplify these defaults with one line:
```
    u = kwargs.get('u', 10)
```

src/compas/geometry/primitives/shapes/cone.py
gonzalocasas:
```suggestion
        """Returns a list of vertices and faces, called by ``Mesh.from_shape()``."""
```

src/compas/geometry/primitives/shapes/cylinder.py
```diff
@@ -165,6 +169,37 @@ def volume(self):
         """Float: The volume of the cylinder."""
         return self.circle.area * self.height
 
+    def to_vertices_and_faces(self, **kwargs):
+        if 'u' in kwargs:
```
gonzalocasas: same here: `u = kwargs.get('u', 10)`

src/compas/geometry/primitives/shapes/sphere.py
```diff
@@ -145,6 +147,57 @@ def to_data(self):
 
         """
         return self.data
+    
+    def to_vertices_and_faces(self, **kwargs):
+        """Returns a list of vertices and faces, called by `Mesh.from_shape()`."""
+        if 'u' in kwargs:
```
gonzalocasas: and here:
```
u = kwargs.get('u', 10)
v = kwargs.get('v', 10)
```

gonzalocasas (approved):
There's some minor formatting thing with white space around arithmetic operators, perhaps run autopep8 on these files; but other than that, it LGTM!

src/compas/geometry/primitives/shapes/cone.py
worbit: great suggestion! will do

src/compas/geometry/primitives/shapes/cone.py
worbit: good to know `code` formatting needs double "`"s, but this specific part will probably removed anyway as suggested by @tomvanmele

src/compas/datastructures/mesh/_mesh.py
worbit: bold!

worbit:
thanks, Gonzalo and Tom, for the suggestions. I will make these changes and push again.

src/compas/geometry/primitives/shapes/box.py
worbit: removed

src/compas/geometry/primitives/shapes/cylinder.py
tomvanmele: there is a small issue to consider here though, which i find a bit an annoying side effect of the `.get()` functionality.

if a user provides `u=None` as a parameter to the function, this will still count as a valid option for `u`, and will break your code. if zero is also unwanted as a value (for example if it should be at least 1) you could do

```python
u = kwargs.get('u') or 10
```

worbit: I see the problem. actually, for the function to "make sense", `u` should even be at least 3. The question now is, what default in what case?
`u` not in `kwargs` -> 10
`u in [None, <0, 0, 1, 2]` -> 3
?

tomvanmele: i think @gonzalocasas was suggesting the `.get` function because it is shorthand for what you wrote in a more verbose way. but now it is a different problem. perhaps just
```python
u = kwargs.get('u') or 0
if u < 3:
    raise ValueError('The value for u should be u > 3.')
```

worbit: that sounds good, i would still do `u = kwargs.get('u') or 10` as a default (corrected typo)

src/compas/geometry/primitives/shapes/sphere.py
```diff
@@ -146,6 +148,54 @@ def to_data(self):
         """
         return self.data
 
+    def to_vertices_and_faces(self, **kwargs):
+        """Returns a list of vertices and faces"""
+
+        u = kwargs.get('u') or 10
+        v = kwargs.get('v') or 10
+        if u < 3 or v < 3:
```
tomvanmele, line 156: i would split this into two checks and two separate, more specific messages...

src/compas/geometry/primitives/shapes/torus.py
```diff
@@ -159,6 +164,41 @@ def to_data(self):
         """
         return self.data
 
+    def to_vertices_and_faces(self, **kwargs):
+        """Returns a list of vertices and faces"""
+
+        u = kwargs.get('u') or 10
+        v = kwargs.get('v') or 10
+        if u < 3 or v < 3:
```
tomvanmele, line 172: same here

gonzalocasas:
@worbit could you pls add a line about this addition on the CHANGELOG (directly on `master`)?

worbit:
Yes done, see [here](https://github.com/compas-dev/compas/blob/master/CHANGELOG.md#added-1). I also changed the separate checks for `u` and `v` as suggested above, but in the PR draft for the `Capsule`.

gonzalocasas:
👍

PR #366, opened 2019-10-22
compas_rhino: Object3D > Artist > PrimitiveArtist > PointArtist ...

Licini:
Added Object3D class which allows parent/children hierarchy and can be shared as base of Artists classes across platforms.

Adjust compas_rhino artists inheritance order into:
 Object3D > Artist > PrimitiveArtist > PointArtist/LineArtist/CircleArtist ...     and
Object3D > Artist > DatastructureArtist > MeshArtist/NetworkArtist/VolmeshArtist

The new API would be like:
```
from compas_rhino.artists import PointArtist
from compas_rhino.artists import Artist
import math

# create a artist as container to draw everything inside it
artist = Artist()

# add point directly
artist.add(Point(10, 0, 0))

# add point with attributes
artist.add(Point(0, 10, 0), attributes={'color': (255, 0, 0)})

# add point through PointArtist
pa1 = PointArtist(Point(-10, 0, 0))
artist.add(pa1)

# add point with attributes through PointArtist
pa2 = PointArtist(Point(0, -10, 0), attributes={'color': (0, 255, 0)})
artist.add(pa2)

R = Rotation.from_axis_and_angle([0, 0, 1], math.pi / 40)

for i in range(0, 20):
    artist.apply_transformation(R)
    artist.draw(0.1)

print('finished!')
```

### Checklist

### DIFF
--- a/src/compas_rhino/artists/artist.py
+++ b/src/compas_rhino/artists/artist.py
@@ -6,6 +6,7 @@ import time
 
 import compas
 import compas_rhino
+from compas.geometry import Object3D
 
 try:
     import rhinoscriptsyntax as rs
@@ -17,7 +18,10 @@ except ImportError:
 __all__ = ['Artist']
 
 
-class Artist(object):
+_ITEM_ARTIST = {}
+
+
+class Artist(Object3D):
     """The base ``Artist`` defines functionality for drawing geometric primitives in Rhino.
 
     Parameters
@@ -55,7 +59,7 @@ class Artist(object):
 
     __module__ = "compas_rhino.artists"
 
-    def __init__(self, layer=None):
+    def __init__(self, geometry=None, attributes=None, layer=None):
         self._layer = None
         self.layer = layer
         self.defaults = {
@@ -63,6 +67,9 @@ class Artist(object):
             'color.line'    : (0, 0, 0),
             'color.polygon' : (210, 210, 210),
         }
+
+        self._last_transformation = None
+        super(Artist, self).__init__(geometry, attributes)
 
     @property
     def layer(self):
@@ -95,6 +102,48 @@ class Artist(object):
             compas_rhino.clear_layer(self.layer)
         else:
             compas_rhino.clear_current_layer()
+
+    @staticmethod
+    def register(item_type, artist_type):
+        _ITEM_ARTIST[item_type] = artist_type
+
+    @staticmethod
+    def build(item, **kwargs):
+        artist_type = _ITEM_ARTIST[type(item)]
+        artist = artist_type(item, **kwargs)
+        return artist
+
+    def add(self, item, **kwargs):
+        if isinstance(item, Artist):
+            artist = item
+        else:
+            artist = self.build(item, **kwargs)
+        self.add_child(artist)
+        return artist
+
+    def remove(self, item):
+        self.remove_child(item)
+
+    def _update_to_rhino(self):
+        if hasattr(self, 'GUID'):
+            T = self.transformation_world
+            if self._last_transformation is not None:
+                T = self._last_transformation.inverse() * T
+            rs.TransformObject(self.GUID, T.matrix)
+            self._last_transformation = self.transformation_world.copy()
+
+    def draw(self, timeout=None):
+        """Update and draw the bounded compas geometry in rhino"""
+        self._update_to_rhino()
+        for c in self.children:
+            c._update_to_rhino()
+        self.redraw(timeout)
+
+    def hide(self):
+        rs.HideObject(self.GUID)
+
+    def show(self):
+        rs.ShowObject(self.GUID)
 
     # ==========================================================================
     # save image, video, gif, ...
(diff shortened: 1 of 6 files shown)


src/compas_rhino/artists/artist.py
```diff
@@ -371,4 +420,37 @@ def draw_circles(self, circles, layer=None, clear_layer=False, redraw=False):
 
 if __name__ == "__main__":
 
-    pass
+    from compas_rhino.utilities import unload_modules
+    unload_modules("compas")
+    unload_modules("compas_rhino")
```
gonzalocasas, line 425: this line is redundant, because the one above (`unload_modules("compas")`) unloads all modules that start with `"compas"`

src/compas_rhino/artists/artist.py
```diff
@@ -55,7 +59,7 @@ class Artist(object):
 
     __module__ = "compas_rhino.artists"
 
-    def __init__(self, layer=None):
+    def __init__(self, geometry=None, attributes=None, layer=None):
```
gonzalocasas, line 62: maybe `layer` is an attribute too?

src/compas_rhino/artists/artist.py
```diff
@@ -55,7 +59,7 @@ class Artist(object):
...
         self._layer = None
```
gonzalocasas, line 63: I know it's not your code, but I always find these two lines redundant. Setting `self._layer = None` has no effect because the next line will set it again to whatever value `layer` has.

src/compas_rhino/artists/artist.py
tomvanmele: agreed. everything that is relevant for serialisation should go in attributes.
but i would add the option of `**kwargs` that allows for overwriting some of the values in the attribute dict using name-value pairs. same as for `add_vertex`, for example...

src/compas_rhino/artists/artist.py
```diff
@@ -17,7 +18,10 @@
 __all__ = ['Artist']
 
 
-class Artist(object):
+_ITEM_ARTIST = {}
+
+
+class Artist(Object3D):
```
gonzalocasas, line 24: This feels very confusing. Why is `Artist` inheriting from `Object3D`? Maybe is correct, but I just find it rather unnatural...

src/compas_rhino/artists/artist.py
tomvanmele:
```python
self._layer = None
self.layer = layer
```
in combination with
```python
@property
def layer(self):
    return self._layer
@layer.setter
def layer(self, layer):
    self._layer = layer
```
is there to allow for processing the value assigned to layer before assigning it.
however, if i am not mistaken, the original version of this used the `property` decorators to allow the user to directly assign to the layer attribute but have the attribute stored in the attributes dict.
```python
self.layer = layer
```
```python
@property
def layer(self):
    return self.attributes['layer']
@layer.setter
def layer(self, layer):
    self.attributes['layer'] = layer
```
since it would make sense to make the layer part of the serialised data. the latter would make sense if we still want the user to have access to the attribute directly...

src/compas_rhino/artists/artist.py
Licini: i see, thx for hint!

src/compas_rhino/artists/artist.py
tomvanmele: shouldn't keeping track of parent-child relationships, transformation stack, ... be a general responsibility of COMPAS objects rather than of the visualisation objects only?

Licini: It is a bit... Another way is to have Object3D under an artist its attribute, but that would create a double layer wrapper situation when adding a geometry directly under an artist as children:  it has to firstly wrap a Object3D around geometry, then a child artist around Object3D, and then add it to the parent artist.  At same time, to access local geometry info from artist would also become quite deep like `artist.object3d.line.start.x`

src/compas_rhino/artists/artist.py
Licini: yep was thinking the same, let me see if there is a way to adjust this without breaking the old api

src/compas_rhino/artists/artist.py
tomvanmele: ideally you don't go through that hierarchy at all (see the new plotters). you just update the geometry directly and this will affect the drawing in the next drawing cycle. the artist only provides access to all attributes related to representation...

src/compas_rhino/artists/artist.py
gonzalocasas: > is there to allow for processing the value assigned to layer before assigning it.

but the thing is always assigned, it's the immediate next line (done via the property setter)
Or am I missing something?

src/compas_rhino/artists/artist.py
gonzalocasas: > you just update the geometry directly and this will affect the drawing in the next drawing cycle

makes more sense to me

tomvanmele: > makes no sense to me

as in "this is a really bad idea!", or "how would that work?!"

gonzalocasas: hehe.. it's the other way around, it's *more*, not *no*! ;)  It does make sense to me

tomvanmele: ha sorry, misread :)

tomvanmele: i think we should first look at a few usage scenarios in some (dummy) snippets across the various interfaces rather than immediately dive in with the implementations like this.

perhaps a few gists to compare?

Licini: haha agree, i m writing up some

gonzalocasas: 👍  API-first design! ;)

Licini: @gonzalocasas @brgcode 
https://gist.github.com/Licini/3c8ca96a68705414b00f81f06b458fe2

Licini: After writing few snippets, starting to agree having artists inheriting object3d was indeed too weird. Actually artist as attribute of object3d makes more sense to me at this point. Also I agree a lot people should never access geometry through an artist which is basically a utility only for visualisation purpose.

Licini: I'm just reposting those snippets here again from slack in case anyone cannot find them:

example of auto-generating artist from object3d: https://gist.github.com/Licini/ddaba6afe47b0b56d09d118399347d21
some ideas about “Collection” class and “Group” class
https://gist.github.com/Licini/f9d8a6aa7bc9d906d1e9a066e2935a84
some possible simplified drawing apis:
https://gist.github.com/Licini/f7b0c3a5c9e39db850ae18f0d0d42fdb
create and update object3ds from CAD:
https://gist.github.com/Licini/258f3a6e655ab27ccca05ddceaee0752

PR #367, opened 2019-10-22
Added iterable_like() to compas.utilities

arpastrana:

### Checklist

### PR Description

Following up on PR #331, it is proposed to add extract data pairing functionality from `offset_polygon()` and to make it generally available like `pairwise()`, `flatten()`, etc. 

Should it be the case, suggestions for improving the name of `match_length()` are very welcome!


src/compas/utilities/itertools_.py
```diff
@@ -273,14 +274,60 @@ def random_combination_with_replacement(iterable, r):
     return tuple(pool[i] for i in indices)
 
 
+def match_length(value, iterable):
```
tomvanmele: regardless of the name, perhaps it would be a good idea to allow the user to provide a "padding" value rather than always using the last value in the list

src/compas/utilities/itertools_.py
```diff
@@ -273,14 +274,60 @@ def random_combination_with_replacement(iterable, r):
...
 # ==============================================================================
 # Main
 # ==============================================================================
 
 if __name__ == "__main__":
 
-    s = range(5)
+    s = list(range(5))
 
     for u, v, w in window(s + s[0:2], 3):
```
tomvanmele, line 333: this reminds me that we should add an optional parameter that does this for the user. it is annoying to have to manually extend the list all the time...

gonzalocasas (changes requested):
Almost all of these things can be done in one line with `itertools`: 
```python
zip_longest(value, target, fillvalue=fill_value)
```
e.g. 
- `match_length(0, "hello")` can be done with `zip_longest([0], "hello", fillvalue=0)`
- `match_length("a", [0, 1, 2])` can be done with `zip_longest("a", [0, 1, 2], fillvalue="a")`
- `match_length(list(range(2)), range(3))` can be done with `zip_longest(range(2), range(3), fillvalue=1)`

zip returns a tuple, so the extra element can easily be discarded, e.g. `[u for u, _ in zip_longest("a", range(3), fillvalue="a")]`

So I don't think this should go into `compas.utilities`.

src/compas/utilities/itertools_.py
```diff
@@ -48,7 +48,8 @@
     'first_true',
     'random_permutation',
     'random_combination',
-    'random_combination_with_replacement'
+    'random_combination_with_replacement',
+    'match_length'
```
tomvanmele: perhaps a good name would be something more in line with how a similar function is named in `numpy` (https://docs.scipy.org/doc/numpy/reference/generated/numpy.pad.html). so something like `to_padded_list`?

in rhino there are a lot of `coerce` functions. perhaps `coerce_list`?

in `numpy` there is also something like `array_like` and `empty_like` => `list_like`?

tomvanmele: https://docs.scipy.org/doc/numpy/reference/generated/numpy.empty_like.html

src/compas/utilities/itertools_.py
arpastrana: I guess a parameter like `wrap=True` can take care of this in the background. But would this be exclusive to the `window()` method?

arpastrana:
Thanks for the feedback, @gonzalocasas! I hadn't seen this one from `itertools' in action before. 
I have the impression though that the one-liner becomes less _readable_. 

Therefore, and following up on what @brgcode suggested, perhaps a wrapper function can be a friendlier alternative. I opted for a numpy-like name. Additionally, it could automatically check whether the supplied `value` is a `list` or an `iterable`, and if not, convert it into one for convenience.

```
def list_like(target, value, fillvalue=None):
    # check for type prior to passing it to zip_longest
    # e.g. 1
    #  if not isinstance(value, list):
            # value = [value]
    # e.g. 2
    # if not is_item_iterable(value):
        # value = iter(value)
    '''
    return [u for _ , u in zip_longest(target, value, fillvalue=fillvalue)]
```

Then one could say:

```
target = "hello"
base = 0.50
new_list = list_like(target, base, 0.0)
# [0.5, 0.0, 0.0, 0.0, 0.0]
```

or 

```
from compas.geometry import pointcloud

target = pointcloud(4, (0, 10))
base = pointcloud(1, (5, 6))
new_list = list_like(target, base, target[2])

```

Would this be an option?

gonzalocasas:
> Would this be an option?

yup! looks good! 👍

tomvanmele:
i think this is a lot better indeed. what about the order of the arguments? does it make sense that the list providing the target length comes first? (just wondering...)

tomvanmele:
`zip_longest` can't be imported in Rhino :)

gonzalocasas:
in py 2.x it's `izip_longest` iirc

gonzalocasas (commented):
At the moment, this works only if target is not a  generator, but it would be possible to change this, and not only support iterators but turn it into an iterator as well. If instead of forcing the `len` part at first and build a list and then slice only up until length of target, you do a `for` on the `zip_longest`, yield each item and then stop iteration based on exhausting the target iterator.

arpastrana:
In the last commit, I have changed the name of the function to `list_like` as discussed 😄.

To handle the discrepancy between python 2 and 3, I included a `try`/ `except ImportError` at the top of the file. I think this is more transparent than having something other exotic import selector internal to the function. 

Target length comes first as I believe that it is more _intuitive_ to say: 
`list_like(this, from_this, with_this)`, 
rather than e.g.
`list_like(from_this, based_on_this, with_this)`.

In the end, didn't include the `isinstance(of_list)` and `is_item_iterable()` checks, as now the function expects two iterables as arguments by definition. Not entirely sure about this move though, but I felt we could go on (almost forever) with the `if` verifications. 

To tackle down the cases where, for some reason, the target `iterable` is shorter than the base one, the `matched` list is sliced before being returned.

arpastrana:
> At the moment, this works only if target is not a generator, but it would be possible to change this, and not only support iterators but turn it into an iterator as well. If instead of forcing the `len` part at first and build a list and then slice only up until length of target, you do a `for` on the `zip_longest`, yield each item and then stop iteration based on exhausting the target iterator.

But would we want to support both `iterators` and `generators` as targets? 😄 

If I got it right, does this look something like:

```
def list_like(target, base, fillvalue=None):
     for _, u in zip_longest(target, base, fillvalue=fillvalue):
        yield u
```

gonzalocasas:
> But would we want to support both iterators and generators as targets? 😄

Yes, totally! :) And actually, that's what I find most problematic about the current state of this PR: it mentions support for iterables, but generators are iterables too, so it is at least misleading. Now the dilemma is that with the current name (`list_like`), it makes no sense to return a generator, because that's really not "list-like". 

-- 
EDIT: We always want to support generators, they are generally more efficient than lists for almost everything except small lists.

tomvanmele:
in addition to all of this above there is the problem of the value being an iterable itself, which happens for example when you try to create a list of colors. `zip_longest` will not yield the expected result when this is the case. for example...
```python
colors = iterable_like(keys, (255, 0, 0))
```

arpastrana:
Since an `iterable` could be a `generator` as well as a `list`, `tuple`, `string`, etc; would it make sense to be more specific and rename the function to `generator_like()` instead of `iterable_like()`?

arpastrana:
The function now yields one value at a time until the target iterable is exhausted.

Added parameter `as_single` to leave to the user the responsibility of declaring whether the `reference` argument passed to `iterable_like()` should be treated as a single entry or not. 

Following up on Tom's example, there would be three likely outcomes:
```
keys = [0, 1, 2]
color = (255, 0, 0)

colors = list(iterable_like(keys, color))
# returns [255, 0, 0]

colors = list(iterable_like(keys, color, color, True))
# returns [(255, 0, 0), (255, 0, 0), (255, 0, 0)]

colors = list(iterable_like(keys, [color], color))
# returns [(255, 0, 0), (255, 0, 0), (255, 0, 0)]
``` 

When matching two iterables:
```
keys = [0, 1, 2, 3, 4, 5]
attributes = ['x', 'y', 'z']

colors = list(iterable_like(keys, attributes))
# returns ['x', 'y', 'z', None, None]

colors = list(iterable_like(keys, attributes, 'u'))
# returns ['x', 'y', 'z', 'u', 'u']

colors = list(iterable_like(keys, attributes, 'u', True))
# returns [['x', 'y', 'z'], ['x', 'y', 'z'], ['x', 'y', 'z'], 'u', 'u']
```

I wonder though if for the sake of simplicity, the function should always take `reference` as a single entry, so that indeed:

```
keys = [0, 1, 2]
color = (255, 0, 0)

colors = list(iterable_like(keys, color))
# returns  [(255, 0, 0), (255, 0, 0), (255, 0, 0)]
```

src/compas/utilities/itertools_.py
```diff
@@ -273,14 +278,60 @@ def random_combination_with_replacement(iterable, r):
...
+
+    Parameters
+    ----------
+    target : iterable
+        An iterable to be matched in size.
+    reference: object
+        Object taken as departure point.
+    fillvalue : object, optional
+        Fill value. Defaults to `None`.
+    is_single : bool, optional
```
tomvanmele: minor type: `is_single` => `as_single`

src/compas/utilities/itertools_.py
```diff
@@ -273,14 +278,60 @@ def random_combination_with_replacement(iterable, r):
...
+    [(255, 0, 0), None, None]
+    >>> [_ for _ in iterable_like(keys, color, color, as_single=True)]
+    [(255, 0, 0), (255, 0, 0), (255, 0, 0)]
+    """
+    if as_single:
+        reference = [reference]
+
+    zipped = zip_longest(target, reference, fillvalue=fillvalue)
+    for _ in target:
+        yield next(zipped)[1]
```
tomvanmele, line 321: why not do
```python
for t, r in zip_longest(target, reference, fillvalue=fillvalue):
    yield r
```

arpastrana: Because, if for some reason `target` is shorter than `reference`, would instead keep returning values until `reference` is exhausted. 

For example:

```
target = range(2)
reference = ['a', 'b', 'c']

for t, r in zip_longest(target, reference):
    yield r

# yields 'a', 'b', 'c'
```

Thus, I find the current proposed way useful not only to extend but also to _cap_ to a target iterable.

src/compas/utilities/itertools_.py
arpastrana: Changed.

PR #369, opened 2019-10-23
Frames subdivision on Meshes

arpastrana:

### Checklist

### Description

This algorithm is (draft) proposed to extend the menu of subdivision methods for compas Meshes. Its functionality, in a simplified version, stems from the work we carried out on the slab project in order to create _ribbed_ meshes. The method naming is reminiscent of a similar method part or the weaverbird plugin in Grasshopper.

It's output, after plotting, looks as follows:

mesh_frames

However, I am not entirely sure about a) whether it would be preferable to encapsulate the code blocks I precede with an enumeration comment (lines 502, 507, 513, 518, 525, 531) as inner functions; particularly for the sake of readability and maintainability, or to convert the method into a subdivision `object`. 

On the other hand, having a look at the other subdivision methods, seems like there is not yet a consensus on using `mesh_fast_copy()` or `SubdMesh()` to handle the creation of a new output `Mesh`. What is the overarching plan for it?

Furthermore, are the face/vertex/edge attributes of the input `Mesh()` intended to be copied over to the _subdivided_ output one?

Feedback is very appreciated! 😄

### DIFF
--- a/src/compas/datastructures/mesh/subdivision.py
+++ b/src/compas/datastructures/mesh/subdivision.py
@@ -7,6 +7,9 @@ from math import pi
 from copy import deepcopy
 
 from compas.geometry import centroid_points
+from compas.geometry import offset_polygon
+
+from compas.utilities import pairwise
 
 from compas.datastructures.mesh._mesh import Mesh
 from compas.datastructures.mesh.operations import mesh_split_edge
@@ -19,6 +22,7 @@ __all__ = [
     'mesh_subdivide_quad',
     'mesh_subdivide_catmullclark',
     'mesh_subdivide_doosabin',
+    'mesh_subdivide_window',
     'trimesh_subdivide_loop',
 ]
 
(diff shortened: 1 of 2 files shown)


src/compas/datastructures/mesh/subdivision.py
```diff
@@ -7,6 +7,10 @@
 from copy import deepcopy
 
 from compas.geometry import centroid_points
+from compas.geometry import offset_polygon
+
+from compas.utilities import pairwise
+from compas.utilities import match_length
```
Licini, line 13: This `match_length` function does not exist in current verions. And the test was failed due to it. Is this something you are working on to add?

Licini: nevermind, just saw it in your other PR

arpastrana: Yes, exactly. It is because of the other PR. Thanks for reviewing it btw!

arpastrana:
Closing this PR for now as it is contingent upon resolving #367.

PR #370, opened 2019-10-24
Transform collections

tomvanmele:

### Checklist


tomvanmele:
just want to confirm that this is what we were talking about before i do the others...

tomvanmele:
was also wondering if it makes sense to make a `PrimitiveCollection`.

and if it is still necessary to convert coordinates to float in the background. py3 does float division anyway. and perhaps it should be anyway up to the user to take care of this...

will shave quite bit of the instantiation time of large collections of objects.

gonzalocasas:
Looks good to me. 

About the name of these functions, are we set on `_collection`? There's a bit of room for misinterpretation because of the [`collections` module](https://docs.python.org/3.8/library/collections.html). Not sure of an alternative thou, maybe `transform_many`?

tomvanmele:
borh shapely and matplotlib refer to "collections" of points. that is where i got rhis from. so it is not that odd ...

romanarust:
Great! I am not sure about PrimitiveCollection though, seems a bit too much...

tomvanmele:
left the name as is for now. let's see what happens...

PR #371, opened 2019-10-24
Install compas_ghpython for Rhino 6.0 for Mac by default

tetov:
I think this was actually mentioned at the ITA19 course by @brgcode. Troubleshot a friend's compas install on Mac and realized that compas_ghpython wasn't installed, hence the PR.

### Checklist

1. ~~Add the change to the `CHANGELOG.md` file in the `Unreleased` section under the most fitting heading: `Added`, `Changed`, `Removed`.~~
(Too small of a change to add to the changelog)


src/compas_rhino/install.py
```diff
@@ -73,6 +68,14 @@ def install(version=None, packages=None):
 
     if not packages:
         packages = INSTALLABLE_PACKAGES
+        if system == 'darwin' and version == '5.0':
+            packages.remove('compas_ghpython')
+    elif system == 'darwin' and 'compas_ghpython' in packages:
```
tomvanmele: this seems to remove `compas_ghpython` from the packages list for both versions of Rhino for Mac. no?

tomvanmele (changes requested):

src/compas_rhino/install.py
tetov: Yes. Should include ’and version == ”5.0”’. Will fix!

tetov: Or do a check once for is_compatible if mac and 5.0 to make it cleaner.

tetov:
```python
if system == 'win32':
        print('Installing COMPAS packages to Rhino {0} IronPython lib:'.format(version))
    elif system == 'darwin':
        print('Installing COMPAS packages to Rhino IronPython lib.')
```

The os shouldn't matter for this print statement, so I removed the check.

gonzalocasas:
looking good! Could you please remove the `.pytest_cache` stuff and add the folder to the `.gitignore` as well?

tetov:
> looking good! Could you please remove the `.pytest_cache` stuff and add the folder to the `.gitignore` as well?
That's embarassing :/. Yeah it's in the pathlib fork already, I'll add it to the same line here as well.

PR #372, opened 2019-10-27
transpose, transposed, invert, inverse, inverted, determinant, concat…

tomvanmele:
transpose, transposed, invert, inverse/inverted, concatenate, concatenated => method naming consistent with intended conventions.
and added determinant...

### Checklist

### DIFF
--- a/src/compas/geometry/transformations/transformation.py
+++ b/src/compas/geometry/transformations/transformation.py
@@ -13,8 +13,10 @@ Ippoliti for providing code and documentation.
 import math
 
 from compas.geometry.basic import multiply_matrices
+from compas.geometry.basic import transpose_matrix
 
 from compas.geometry.transformations import matrix_inverse
+from compas.geometry.transformations import matrix_determinant
 from compas.geometry.transformations import identity_matrix
 from compas.geometry.transformations import matrix_from_frame
 from compas.geometry.transformations import matrix_from_euler_angles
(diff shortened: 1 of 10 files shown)


src/compas/geometry/transformations/transformation.py
```diff
@@ -264,11 +323,27 @@ def inverse(self):
...
         """
-        cls = type(self)
-        return cls(matrix_inverse(self.matrix))
+        # cls = type(self)
+        # return cls(matrix_inverse(self.matrix))
+        T = self.copy()
+        T.invert()
+        return T
+
+    inverted = inverse
```
romanarust, line 332: Does `inverted` make it to the Documentation like that as well?

romanarust (commented):
lgtm!

tomvanmele:
should i merge this and bump up the version or do we want to wait with that for more changes?

worbit:
in terms of naming i think the clearest would be if we consistently use the present/imperative verb form if it is an in-place operation, and the past tense, if a copy is returned.
`transform` > object is transformed
`transformed` > a transformed copy of object is returned, object unaffected

gonzalocasas (approved):
LGTM 👍

tomvanmele:
so but merge and bump up or wait?

gonzalocasas:
> so but merge and bump up or wait?

are there more of these changes coming? if so, wait, maybe create more PRs to merge onto this one, instead of all to master... otherwise, merge and bump...

PR #374, opened 2019-10-29
add capsule, some typos, separate checks u and v

worbit:

As discussed, add the `Capsule` class, will add unimplemented methods (eg `to_vertices_and_faces`)

### DIFF
--- a/src/compas/geometry/primitives/line.py
+++ b/src/compas/geometry/primitives/line.py
@@ -168,6 +168,24 @@ class Line(Primitive):
         v = self.direction * (0.5 * self.length)
         return self.start + v
 
+    @property
+    def data(self):
+        """Returns the data dictionary that represents the line.
+
+        Returns
+        -------
+        dict
+            The line data.
+
+        """
+        return {'start': list(self.start),
+                'end': list(self.end)}
+
+    @classmethod
+    def from_data(cls, data):
+        line = Line(data['start'], data['end'])
+        return line
+
     # ==========================================================================
     # representation
     # ==========================================================================
--- a/src/compas/geometry/primitives/line.py
+++ b/src/compas/geometry/primitives/line.py
@@ -260,7 +260,7 @@ class Line(Primitive):
     # ==========================================================================
 
     def transform(self, matrix):
-        """Transform this ``Plane`` using a given transformation matrix.
+        """Transform this ``Line`` using a given transformation matrix.
 
         Parameters
         ----------
--- a/src/compas/geometry/primitives/line.py
+++ b/src/compas/geometry/primitives/line.py
@@ -259,34 +259,33 @@ class Line(Primitive):
     # transformations
     # ==========================================================================
 
-    def transform(self, matrix):
-        """Transform this ``Line`` using a given transformation matrix.
+    def transform(self, transformation):
+        """Transform the line.
 
         Parameters
         ----------
-        matrix : list of list
-            The transformation matrix.
-
+        transformation : :class:`Transformation`
+            The transformation used to transform the line.
         """
-        self.start.transform(matrix)
-        self.end.transform(matrix)
+        self.start.transform(transformation)
+        self.end.transform(transformation)
 
-    def transformed(self, matrix):
-        """Return a transformed copy of this ``Line`` using a given transformation matrix.
+    def transformed(self, transformation):
+        """Returns a transformed copy of the current line.
 
         Parameters
         ----------
-        matrix : list of list
-            The transformation matrix.
+        transformation : :class:`Transformation`
+            The transformation used to transform the line.
 
         Returns
         -------
-        Line
-            The transformed copy.
+        :class: `Line`
+            The transformed line.
 
         """
         line = self.copy()
-        line.transform(matrix)
+        line.transform(transformation)
         return line
 
 
--- a/src/compas/geometry/primitives/shapes/capsule.py
+++ b/src/compas/geometry/primitives/shapes/capsule.py
@@ -112,6 +112,24 @@ class Capsule(Shape):
         """
         return self.data
 
+    @classmethod
+    def from_data(cls, data):
+        """Construct a capsule from its data representation.
+
+        Parameters
+        ----------
+        data : :obj:`dict`
+            The data dictionary.
+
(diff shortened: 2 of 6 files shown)


src/compas/geometry/primitives/shapes/capsule.py
```diff
@@ -0,0 +1,199 @@
...
+    @property
+    def line(self):
+        return self._line
+    
+    @line.setter
+    def line(self, line):
+        # add check if segment is
+        # ((x,y,z),(x,y,z)) or
+        # [[x,y,z],[x,y,z]] or
+        # [point,point] or ...
```
tomvanmele: what is all this for?

src/compas/geometry/primitives/shapes/capsule.py
```diff
@@ -0,0 +1,199 @@
...
+        capsule.line.transform(matrix)
+        return capsule
+
+
+# ==============================================================================
+# Main
+# ==============================================================================
+
+if __name__ == "__main__":
+    pass
```
tomvanmele: end with newline...

tomvanmele (commented):
i guess the only thing left to do is the caps?

src/compas/geometry/primitives/shapes/capsule.py
worbit: thanks for reminding me, this is a leftover from before i used a `Line` object for the `Capsule` axis

src/compas/geometry/primitives/shapes/capsule.py
```diff
@@ -0,0 +1,199 @@
...
+            The transformation matrix.
+        """
+        self.line.transform(matrix)
+    
+    def transformed(self, matrix):
+        """Return a transformed copy of this ``Capsule`` using a given transformation matrix.
+
+        Parameters
+        ----------
+        matrix : list of list
```
romanarust: Shouldn't this be rather a :class:`Transformation` than the pure matrix?

src/compas/geometry/primitives/shapes/capsule.py
```diff
@@ -0,0 +1,199 @@
...
+            The transformation matrix.
+
+        Returns
+        -------
+        Capsule
```
romanarust:
```suggestion
:class:`Capsule`
```

src/compas/geometry/primitives/line.py
```diff
@@ -260,7 +260,7 @@ def copy(self):
     # ==========================================================================
 
     def transform(self, matrix):
```
romanarust: shouldn't this be rather a :class:`Transformation`?

src/compas/geometry/primitives/shapes/capsule.py
```diff
@@ -0,0 +1,228 @@
...
+            The copy.
+        """
+        cls = type(self)
+        return cls(self.line.copy(), self.height)
+
+    # ==========================================================================
+    # transformations
+    # ==========================================================================
+
+    def transform(self, matrix):
```
romanarust:
```suggestion
    def transform(self, transformation):
```

src/compas/geometry/primitives/shapes/capsule.py
```diff
@@ -0,0 +1,228 @@
...
+    Attributes
+    ----------
+    segment: :class:`compas.geometry.Line`
+        The axis line of the capsule.
+    radius: float
+        The radius of the capsule.
+
+    Examples
+    --------
+    >>> TODO
```
romanarust:
```suggestion
    >>> line = Line((1, 2, 3), (5, 4, 3))
    >>> capsule = Capsule(line, 1.2)
```

src/compas/geometry/primitives/shapes/capsule.py
```diff
@@ -0,0 +1,228 @@
...
+        """Returns the data dictionary that represents the capsule.
+
+        Returns
+        -------
+        dict
+            The capsule data.
+
+        """
+        return self.data
+
```
romanarust, line 115: what about a `__repr__` that returns a string that can be used to construct a Capsule again, like:
```suggestion
    def __repr__(self):
        return "Capsule({0}, {1})".format(self.line, self.radius)
```

romanarust (changes requested):
docstrings and examples missing here and there

src/compas/geometry/primitives/line.py
tomvanmele: yes indeed

src/compas/geometry/primitives/shapes/capsule.py
tomvanmele: does `repr` get applied recursively? i mean, will this apply `repr` to `self.line` as well?

src/compas/geometry/primitives/shapes/capsule.py
tomvanmele: agreed

src/compas/geometry/primitives/shapes/capsule.py
tomvanmele: yes indeed

src/compas/geometry/primitives/shapes/capsule.py
```diff
@@ -0,0 +1,228 @@
...
+# Main
+# ==============================================================================
+
+if __name__ == "__main__":
+    from compas.datastructures import Mesh
+    line = Line((1, 2, 3), (5, 4, 3))
+    capsule = Capsule(line, 1.2)
+
+    mesh = Mesh.from_shape(capsule)
+    mesh.to_obj('/Users/bernham/Desktop/capsule.obj')
```
tomvanmele: should this not be a call to `doctest`?

src/compas/geometry/primitives/shapes/capsule.py
worbit: Agree. I copied this from `primitives.Line` where it is also a matrix and this same docstring. I will go through all the `Shape` classes and update it according to your suggestion.

tomvanmele:
@worbit is this thing intentionally still a draft?

worbit:
no, i think it's good to get merged!

PR #375, opened 2019-10-29
Start compas.cad package

Licini:
This is to layout the basic structure of compas.cad package.

1. added object3d class with functionality of storing local/world transformation, hosting and traversing it children/parent nodes.
2. added scene class as container of object3ds to be visualised conveniently as a whole

### Checklist

### DIFF
--- a/src/compas/cad/scene.py
+++ b/src/compas/cad/scene.py
@@ -1,0 +1,30 @@
+from compas.cad import Object3D
+
+
+__all__ = ["Scene"]
+
+
+class Scene(Object3D):
+
+    """
+    Scence class serves as a container to store object3ds for visualization
+    """
+
+    __module__ = "compas.cad"
+
+    def __init__(self):
+        super(Scene, self).__init__()
+
+    @property
+    def parent(self):
+        return None
+
+
+if __name__ == "__main__":
+
+    from compas.geometry import Point
+    from compas.cad import Object3D
+
+    scene = Scene()
+    scene.add(Object3D(geometry=Point(0, 0, 0)))
+    print(scene.get_children())
--- a/src/compas/cad/scene.py
+++ b/src/compas/cad/scene.py
@@ -1,10 +1,7 @@
-from compas.cad import Object3D
-
-
 __all__ = ["Scene"]
 
 
-class Scene(Object3D):
+class Scene(object):
 
     """
     Scence class serves as a container to store object3ds for visualization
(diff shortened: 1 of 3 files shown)


src/compas/cad/scene.py
```diff
@@ -0,0 +1,30 @@
+from compas.cad import Object3D
+
+
+__all__ = ["Scene"]
+
+
+class Scene(Object3D):
```
tomvanmele: somehow this feels wrong to me. i would think that a scene and the objects in it are profoundly different. is this coming from ThreeJs?

gonzalocasas:
I think `cad` is really not the correct name. Sorry to complain about naming again thou, but there's nothing in this package that relates to CAD. And actually, CAD is such a generic term that we should not use it to indicate that something is a sort of base for all CADs, because it is likely that we can reuse this for non-CAD tooling, like Unity/Unreal and other gaming engines.

Also, as a sort of convention remark, I believe any PR that proposes an entirely new package should include much more documentation at the module level. This one does not even link the new module to the API reference index, so no docs will be displayed at all.

Licini:
> I think `cad` is really not the correct name. Sorry to complain about naming again thou, but there's nothing in this package that relates to CAD. And actually, CAD is such a generic term that we should not use it to indicate that something is a sort of base for all CADs, because it is likely that we can reuse this for non-CAD tooling, like Unity/Unreal and other gaming engines.
> 
> Also, as a sort of convention remark, I believe any PR that proposes an entirely new package should include much more documentation at the module level. This one does not even link the new module to the API reference index, so no docs will be displayed at all.

My bad for the doc, was intending to get some feedback from you guys first and finish docs at very end, maybe I should have put this as draft PR.

You have very fair points about the name `cad`, we are so used to migrate familiar terms and mis-apply them unconsciously everywhere....  
How about something like `canvas` or `canvas3d` ?   Or any other suggestions?

gonzalocasas:
`canvas` sounds good to me, but my background with JS might be a bias here, so, we probably need to sit together and go through a few options. Could you please add a card about this for the next dev meeting?

Licini:
> `canvas` sounds good to me, but my background with JS might be a bias here, so, we probably need to sit together and go through a few options. Could you please add a card about this for the next dev meeting?

sounds good, yep let's do that!

src/compas/cad/scene.py
Licini: Embarrassingly yes, https://threejs.org/docs/index.html#api/en/scenes/Scene
The main reason is to use the get_children() function from object3d to traverse everything held under the scene. However other than that, these is not much needed to inherent from object3d.  We could also have them as two entirely separate classes, which seems more right to you?

tomvanmele:
i don't think it should be `compas.canvas`. it will not be clear what the package is for. furthermore, it should provide a basis for more than just the artists. it should also formalise picking, selecting, and finding objects, and an interface for, for example, updating attributes, and so on. so `compas.cad` might not be that bad after all...

tomvanmele:
https://docs.blender.org/manual/en/latest/scene_layout/index.html

PR #377, opened 2019-10-29
make matrix math specific to 4x4 matrices

Licini:
Added 4x4 matrix specific functions (skipping dimension checks) to make transformation faster

### Checklist

### DIFF
--- a/src/compas/geometry/basic.py
+++ b/src/compas/geometry/basic.py
@@ -1101,12 +1101,39 @@ def multiply_matrices4(A, B):
 
     Examples
     --------
-    >>> TODO
+    >>> A = Translation([5.0, 5.0, 5.0])
+    >>> B = Translation([5.0, 5.0, 5.0])
+    >>> multiply_matrices(A, B)
+    [[1.0, 0.0, 0.0, 10.0], [0.0, 1.0, 0.0, 10.0], [0.0, 0.0, 1.0, 10.0], [0.0, 0.0, 0.0, 1.0]]
     """
     A = list(A)
     B = list(B)
-    B = list(zip(* list(B)))
-    return [[dot_vectors(row, col) for col in B] for row in A]
+    return [
+            [
+                A[0][0]*B[0][0] + A[0][1]*B[1][0] + A[0][2]*B[2][0] + A[0][3]*B[3][0],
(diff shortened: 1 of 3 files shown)


Licini:
Questions:

1. Should the 4x4 specific function added to basics (like how it is now in commit), or resides somewhere under transformations folder
2. Does the function need to be written down even more explicitly (right now only skipping shape check)

gonzalocasas:
I am not sure if I understood what @brgcode meant on this one, my impression was that were were going to change all existing functions to work with 4x4 matrices, not add a new variant that works only with 4x4; but I might be wrong.

tomvanmele:
no i meant to add versions of matrix functions that work specifically/only for 4x4 matrices. but not just remove the shape check. also write out the math to remove all loops and make them as fast as possible in pure python...

tomvanmele:
`transpose_matrix4`
`multiply_matrices4`
`invert_matrix4`
`determinant_matrix4`

Licini:
> `transpose_matrix4`
> `multiply_matrices4`
> `invert_matrix4`
> `determinant_matrix4`

That's sounds more right to me.....However I'm a tiny bit doubting on how much performance it could improve since those are very small loops. But let me write that out and do a bit benchmark, probably will also compare with numpy. Optimisation tasks are always so fascinating

Licini:
So after a little benchmark:
1 million original matrix multiplications : ~13s
1 million new explicit matrix multiplication : ~5s
1 million Numpy matrix multiplication : ~0.8s

tomvanmele:
we could add a C++ version using ctypes in the mix to see if that makes a big difference...

gonzalocasas:
> we could add a C++ version using ctypes in the mix to see if that makes a big difference...

you mean, as a baseline benchmark? because I don't expect we'll do better than numpy on that..

In the meantime, I tried a few things:
1. Using a tuple instead of list of list for the pure python version
1. Running the same tests inside Rhino
1. Added a test using Rhino SDK as backend

My benchmark values are as follows:
- 1 million original matrix multiplications : ~25s
- 1 million new explicit matrix multiplication : ~8s
- 1 million Numpy matrix multiplication : ~1.2s
- 1 million new explicit matrix multiplication using *tuples*: ~5s
- 1 million Rhino SDK matrix multiplication: ~1.1s

So, imo, we should focus on numpy and C# as backends, and transparently select the right one based on the environment.

PR #386, opened 2019-11-06
RPC Command-line interface

gonzalocasas:
The idea of implicitly starting the RPC server from the proxy side is very problematic, so I propose we officialize a more straightforward approach, where the RPC server is explicitly started from the command line via a little cli utility.

The functionality of automatically trying to start it is still there, but I have now documented the new cli-based option, and I would suggest we focus on using that one more often in courses, workshops, etc. It will also make errors and output much more explicit (now it's very often the case that a failure is completely obscure and it ends up in something like `Fault` exception on the proxy side).

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,14 +10,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Added `compas.geometry.icp_numpy` for pointcloud alignment using ICP.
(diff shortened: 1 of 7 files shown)


yck011522:
I really liked the idea where the proxy can be started automatically in the background. It felt like magic and it felt elegant to end-user. With an explicit start, an end user will need to keep another opened console, get to the correct environment manually, start the server and keep that console alive.

Why is the previous background proxy problematic?

Regarding "Fault" errors, can the error messages that happen within the proxy be properly thrown back to the host?

PR #387, opened 2019-11-07
closes #359: Added is_line_line_colinear()

### LINKED ISSUE #359: is_line_line_colinear
# Feature Request

As a **COMPAS user**, I want **compas.geometry.queries to have a function for lines co-linear check** and so that **we can all use it**.

## Details

**Describe the solution you'd like**
I would like the following (or something similar) to happen:
```
line1 = Line([0, 0, 0], [1, 1, 1])
line2 = Line([2, 2, 2], [3, 3,3])
assert is_line_line_colinear(line1, line2) == true
"Another Direction should work too"
line3 = Line([-2,- 2, -2], [3, 3,3])
assert is_line_line_colinear(line1, line3) == true
```

**Describe alternatives you've considered**
Currently there is a function to check if 3 points in colinear. 
https://github.com/compas-dev/compas/blob/16f3f6ef3b49a6a2191f41feb5da79d7bdac3f5c/src/compas/geometry/queries.py#L117

There is also feature to do intersection between 2 lines.  
https://github.com/compas-dev/compas/blob/16f3f6ef3b49a6a2191f41feb5da79d7bdac3f5c/src/compas/geometry/queries.py#L792

But not a simple 2 lines colinear check, or did I miss something? Would be useful to have this feature.

similarly, `is_line_line_overlap(l1, l2, epsilon)` might also be useful

yck011522:

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added `__version__` to `compas_plotters`.
 - Added `compas_plotters` to `.bumpversion.cfg`.
 - Added `Colormap` to `compas.utilities`.
+- Added `is_line_line_colinear()` to `compas.geometry`
 - Added link to Github wiki for devguide.
 - Added pointcloud alignment example to docs.
 - Show git hash on `compas.__version__` if installed from git.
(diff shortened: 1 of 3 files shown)


tomvanmele:
nice!

PR #390, opened 2019-11-08
fixes LayerVisible different argument naming between Rhino5 and Rhino6

Achillx:
Argument names have changed between Rhino5 and Rhino6 for `LayerVisible` command. This fix switches from named arguments to positional ones. Also changes the behaviour of setting back the visibility state of the parent layer, to always return to visible. 

### Checklist

2. [x] Run all tests on your computer (i.e. `invoke test`).
3. [ ] If you add new functions/classes, check that:


tomvanmele:
cool thanks!

PR #395, opened 2019-11-12
point collection and collection with numpy backend

tomvanmele:
Check the timings at the bottom of the modules.
The Numpy based collection is factor 100 faster than the Python one...

### Checklist

### DIFF
--- a/src/compas/geometry/primitives/__init__.py
+++ b/src/compas/geometry/primitives/__init__.py
@@ -15,7 +15,9 @@ from .polygon import Polygon
 from .circle import Circle
 from .curve import Bezier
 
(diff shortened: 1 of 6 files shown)


tomvanmele:
a few notes:
* i would change the names of the static methods of the primitives that transform multiple objects at once to `_multiple` instead of `_collection` (as originally suggested by @gonzalocasas) because collections are now there own separate thin.
* perhaps further improvements are possible by using `numpy`'s structured arrays
* numpy based collections could automatically expose a proxy function and provide convenience methods for the JSON encoders and decoders to ensure transparent transfer of data across the wire...

Issue #341, opened 2019-09-29
Documentation tied to version of compas pkg.


tetov:
# Documentation tied to version of compas pkg.

## Problem description

I'd guess that the majority of users are installing compas from pypi or conda which means that they are installing the repo at a [tagged commit](https://github.com/compas-dev/compas/tags). If they want to view the [documentation online](https://compas-dev.github.io/main/) however they will see documentation from the most recent commit, which could cause confusion.

So the correct (version specific) documentation will be available in the installed package but not online.

## Proposed solution

I think it would be good to consider a system that allows documentation tied to tagged commits.

Two examples with both works with Sphinx would be:
* [sphinxcontrib-versioning](https://github.com/sphinx-contrib/sphinxcontrib-versioning)
* [ReadTheDocs](https://readthedocs.org/)

Sphinxcontrib-versioning could probably work with our current `Invoke` solution while a solution like ReadTheDocs would mean having the docs built and hosted by another service.

If there's interest I'd happily help making this change.

tomvanmele:
hi anton,

yes, that is indeed an unsolved problem that causes confusion. especially when we make backwards incompatible changes to the API.

as far as i know, it is not possible to use your own theme with RTD, so i would prefer the `sphinxcontrib-versioning` solution.

would be helpful indeed if you could propose a solution.

thanks,
tom

tetov:
Hi Tom!

I see. Yeah, RTD only allows its own theme or the standard Sphinx theme.

I will try to get the versioning-extension to work in the coming week. There's is some questions regarding the maintenance of the project, see <https://github.com/sphinx-contrib/sphinxcontrib-versioning/issues/59>, but there seems to be people interested and keen to keep it going.

**EDIT**
Here's an [example of sphinxcontrib-versioning in use](http://www.clawpack.com/), [source](https://github.com/clawpack/doc). Adding it here to give an example, and as a note to self.

tetov:
![image](https://user-images.githubusercontent.com/14882117/65876928-bf8a1380-e38a-11e9-91b7-2d84d8643a11.png)

Easiest would be to add the version selection like mockup above. Would that be a good solution?

And make the current version bold or in a different color.

tomvanmele:
if clicking on the version number switches the docs to the corresponding version, then that is great :)

worbit:
pasting here from a parallel discussion:

[...]

PR #396, opened 2019-11-16
Deploy docs only for tagged commits (releases)

tetov:

### Checklist

### Description
 
I've tried to implement sphinxcontrib-versioning to have docs generated for all releases of Compas, as described in #341. I've gotten it to work now after some patching of the sphinx extension (it's in maintenance limbo at the moment, again see #341 ).

However, Sphinx itself fails for a few of the tagged releases and fixing all those errors release by release doesn't seem feasible.

I suggest instead that we change the Travis job to only deploy after building tagged commits as described in [Travis docs](https://docs.travis-ci.com/user/deployment/releases/#deploying-only-on-tagged-builds).

This means that the docs will reflect the latest version of Compas which is most likely the one the user has installed. Travis will still run `invoke docs` and fail if there are errors in the Sphinx build. Contributors can run `invoke docs` to preview changes to the docs.

I've ran a few tests:
* [Tagged commit](https://travis-ci.org/tetov/compas/jobs/612837967#L3820) (tagged using `git tag -a 123)
* [Normal commit](https://travis-ci.org/tetov/compas/jobs/612840492#L3976)

This will only work if the commit is tagged before it's pushed, not if commits are tagged retroactively.

(Tags are never included in pull requests, at least according to [person on Stack Overflow](https://stackoverflow.com/a/45553305)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,5 +27,5 @@ deploy:
   target-branch: master
   local_dir: dist/docs
   on:
-    branch: master
+    tags: true
     condition: $TRAVIS_PYTHON_VERSION = 3.6
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Docs are only deployed to github pages for tagged commits.
 - Fixing printing issue with `compas.geometry.Quarternion` in ironPython.
 - Fixed a missing import in `compas.geometry.Polygon`.
 - Removed unused imports in `compas.geometry.Polyline`.


tetov:
Sidenote: Given the amount of time I've spent on this I find it quite funny that in the end the PR only has one edited line. Even the edit to the changelog is longer.

gonzalocasas:
lgtm!

gonzalocasas (approved):

tetov:
Reposting parts of slack conversation with @gonzalocasas.

Reverting to an older commit, committing the revert and tagging it would make Travis deploy. That could be useful in case a rollback ever needs to be done.

The bumpversion workflow would change a little bit, tagging the commit needs to happen before the push.

PR #403, opened 2019-11-25
Fix target generator exhaustion in iterable_like()

arpastrana:

### Checklist

### Description

This is a bug fix to #367.  
Zipped values from `zip_longest` are computed and yielded one by one until `target` is exhausted.

Passing a `generator` as `target` will raise a `StopIteration` exception prematurely since `zip_longest` would have already consumed it. Therefore, `itertools.tee()` is introduced to create a `counter` upon which zipped values will be released.

We iterate over `counter`, (previously `target`), to also cap `reference` to the size of `target` in case the former is target than the latter. 

Default parameters have also been adjusted.

### How to reproduce?

This works (list and tuple):

```
import compas
from compas.utilities import iterable_like

keys = [0, 1, 2, 3]
color = (255, 0, 0)
f = list(iterable_like(keys, color))
print(len(keys) == len(f))  # True
```

This will not (generator and generator):

```
import compas

from compas.utilities import iterable_like

faces = Mesh.from_obj(compas.get("faces.obj"))
hypar = Mesh.from_obj(compas.get("hypar.obj"))

f = list(iterable_like(faces.faces(), hypar.faces(), as_single=False))
print(len(list(faces.faces())) == len(f))  # False
```


tomvanmele:
would it make sense to drop the `as_single` parameter and leave it up to the user to wrap the reference value if it is supposed to be seen as a "single" value?

```python
colors = iterable_like(keys, [(255, 0, 0)])
```

arpastrana:
It definitely makes it more succinct.

I have the impression though that `fillvalue` would need to always be explicitly defined since `reference` could thus be any iterable.

```
color = (255, 0, 0)
colors = iterable_like(keys, [color], color)
```

Would this look reasonable to you?

tomvanmele:
it makes everything more explicit, which is a good idea i think...

arpastrana:
`as_single` has been removed; dependant functions modified accordingly 😄

PR #404, opened 2019-11-29
box is now at origin

tomvanmele:
The box shape is inconsistent with other shapes. If it is defined by its properties, it is not located at the origin, and the local origin is not at the center of the box. This PR fixes that...

### Checklist

### DIFF
--- a/src/compas/geometry/primitives/shapes/box.py
+++ b/src/compas/geometry/primitives/shapes/box.py
@@ -4,6 +4,7 @@ from __future__ import print_function
 
 from compas.geometry import subtract_vectors
 from compas.geometry.primitives import Frame
+from compas.geometry.primitives import Point
 from compas.geometry.primitives import Vector
 from compas.geometry.primitives.shapes import Shape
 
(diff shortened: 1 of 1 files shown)


romanarust:
Yes, correct. I would go for centering in x and y, but not in z.

romanarust:
Correct, it very much feels strange that the frame is in the center of box....

tomvanmele:
i understand, but it is inconsistent with other shapes and that is arguably the most confusing. note that if you create the box from spatial geometric inputs it will still be located in relation to those inputs, but the origin of its frame is always at the centroid. this seems to make the most sense for applying transformations...

worbit:
hi all
i can understand the reservations, but i still think this makes sense.

Screenshot 2019-11-29 at 15 05 42

note that also grasshopper (left) and rhino (right) put the gumball / transformation handle / plane in the centre of the box (after creation). creating a box from two opposite corner points also still works in compas!

tomvanmele:
yes, so the methods `from_diagonal`, `from_corner_corner_height`, ... still place the box at the specified locations, but the local coordinate system is at the centroid such that transformations can be applied transparently...

romanarust:
true, then let's leave it like that. but we should we then provide methods like `Box.bottom/top_frame` and `Cylinder.bottom/top_frame` for convenience?

Issue #398, opened 2019-11-19
Polyhedron class is missing to/from data


gonzalocasas:
The `Polyhedron` class is the only `Shape` class that does not implement from/to data operations. Since all other shapes implement this, I categorize this as a bug rather than feature request.

**Steps to reproduce**
```
>>> t = Polyhedron.from_platonicsolid(4)
>>> t.data
Traceback (most recent call last):
    raise NotImplementedError
```

* compas version: `0.11.0`

PR #405, opened 2019-12-01
Added data methods to Polyhedron()

arpastrana:

### Checklist

### PR Description

This first commit of the PR intends to solve issue #398 by implementing `to_data`, `from_data` methods; as well as the `data` property.

@gonzalocasas I was wondering though if we should check if `len(faces)` is one of the supported face count values: (4, 6, 8, 12, 20)?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Added inherited methods to class docs.
 - Added data structure mixins to the docs.
-- Added `data`, `to_data`, and `from_data` to `compas.geometry.Polyhedron`
+- Added `data` and `from_data` to `compas.geometry.Polyhedron`
 
 ### Changed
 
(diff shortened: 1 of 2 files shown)


arpastrana:
I think we could do that with `cls.generate()`...

tomvanmele:
could you remove the `to_data` function? want to get rid of them everywhere...

arpastrana:
Sure! Just for curiosity, why would you like to get them removed everywhere?

tomvanmele:
just a wrapper for `data` and i would like to go towards "1 correct/preferred way of doing things"

gonzalocasas:
>>> could you remove the to_data function? want to get rid of them everywhere...

>> Sure! Just for curiosity, why would you like to get them removed everywhere?

> just a wrapper for data and i would like to go towards "1 correct/preferred way of doing things"

I would prefer to keep `to_data` for two reasons. 1) If there is a `from_data`, the "principle of least surprise" says there should be a complementary `to_data`. 2) consistency: the convention of `to_data`/ `from_data` is extremely spread, not only within compas, but on extension packages, getting rid of it partially will create inconsistency in the ecosystem.

So even if this is already merged, I would really suggest to re-add it.

PR #406, opened 2019-12-05
Removing autodoc for removed module

tetov:

- [x] Doc fix.

Since compas.remote has been removed in favor of compas.rpc autodoc can't import compas.remote.


gonzalocasas (approved):
😎👍

PR #418, opened 2019-12-05
Base reader, refactor and pathlib support

tetov:
This is a new PR with the changes from #348 but cleaned up.

@gonzalocasas latest code review will be copied over to this PR.

### Checklist


src/compas/files/base_reader.py
```diff
@@ -0,0 +1,103 @@
...
+            pathobj = Path(location)
+        else:
+            pathobj = location
+
+        if pathobj.exists():
+            return pathobj
+        else:
+            raise IOError('File not found.')
+
+    def check_file_signature(self):
```
tetov: @gonzalocasas:
> This method is great stuff, but I would make a few changes for clarity, because it took me a while to understand what you're doing here:
> 
> 1. `self.file_signature` is a constant per-class, so I would make that uppercase, eg `self.FILE_SIGNATURE`.
> 2. Docs! This is a pretty significant addition to the public API, so, it really needs to be well documented.
> 3. The first check cannot be passed, there are 3 options: 1) the support for urls is implemented 2) not implemented and raises exception, 3) not implemented, but add a `return` immediately.
> 4. Try to avoid nesting checks. The `else` that contains everything is not needed, because the previous code paths end in error (at least, they should, see previous point)
> 
> Then, two questions:
> 
> 1. Does this work with ascii files (since you open the file in binary mode)?
> 2. Would it be possible to use `urlretrieve` and actually remove al the remote stuff to be a one-time download into temp and handle everything as files?

@tetov:
> Questions:
> 
> 1. Yes!
> 2. This goes against what I thought you wanted, to not load a big file into memory?

src/compas/files/obj.py
```diff
@@ -276,8 +242,8 @@ def _read_freeform_geometry(self, name, data):
             if self.deg[0] == 1:
                 if len(data) == 4:
                     self.lines.append((int(data[2]) - 1, int(data[3]) - 1))
-                    if self.group:
-                        self.groups[self.group].append(('l', len(self.lines) - 1))
+                if self.group:
```
tetov: @gonzalocasas:
> I don't understand this one... is it really correct?

src/compas/files/off.py
```diff
@@ -113,7 +87,7 @@ def pre(self):
     def post(self):
         pass
 
-    def read(self):
+    def read_off(self):
```
tetov: @gonzalocasas:
> I'm not very convinced about calling these `read_off()`, `read_obj()` and so on. I think it should be the same name across the board. I see that now `read()` is used, so, either rename this level of reads to `read_content` or `read_file`, or rename the `BaseReader.read` to `BaseReader.read_bytes` (or `BaseReader.read_line` or `BaseReader.read_chunk`)

src/compas/files/off.py
```diff
@@ -127,11 +101,7 @@ def read(self):
 
         """
         if not self.content:
-            return
-
-        header = next(self.content)
-        if not header.lower() == 'off':
-            return
+            raise Exception('Import failed')
```
tetov: @gonzalocasas:
> `Import failed` is not very clear, we're not really importing, we're opening files (which will then might be imported into something), so I would be more explicit and limited to this module's functionality in the error message.

tests/compas/files/test_stl.py
```diff
@@ -1,36 +0,0 @@
-import os
```
tetov: @gonzalocasas:
> Is this test now done somewhere else?

@tetov:
> Yes. Moved to test_mesh.py under tests/compas/datastructures.

@gonzalocasas:
> ok! I actually think it is more proper to keep this in a `compas/files/test_stl` because it's supposed to be unit test, and by testing the mesh_from_stl, you're actually already integrating a lot of other things; but in this case, it's ok if you want to keep it in the new location, it's not such a big test anyway.. the only thing I would suggest thou is to split the `test_from_stl` method into different ones, each describing what they are actually testing (now it's only a comment but if it fails, you dont see the comment text in the test failure)

tetov: @gonzalocasas:
Ok, I think we need to be consistent here. Would you prefer I set up test_obj, test_ply etc as well? That wouldn't be a problem for me.

tetov:
[Checks fail](https://travis-ci.com/compas-dev/compas/jobs/263651836#L547) because I test using a file that is not yet present in compas/data..

src/compas/files/ply.py
```diff
@@ -137,154 +146,149 @@ def read(self):
...
                 elif section == 'edge':
-                    self.read_edges_binary_wo_numpy()
+                    self.read_edges_binary()
                 elif section == 'face':
-                    self.read_faces_binary_wo_numpy()
+                    self.read_faces_binary()
                 else:
-                    print('user-defined elements are not supported: {0}'.format(section))
-                    pass
+                    raise Exception(
```
tomvanmele: is the entire process supposed to fail because the file contains some non-standard elements as well?

tests/compas/datastructures/test_mesh.py
```diff
@@ -459,7 +521,9 @@ def test_centroid():
     assert mesh.centroid() == [0.0, 0.0, 0.5]
 
     mesh = Mesh.from_obj(compas.get('quadmesh.obj'))
-    assert mesh.centroid() == [2.508081952064351, 2.554046390557884, 1.2687133268242006]
+    assert mesh.centroid() == [
```
tomvanmele: why is this split? max-line-length for compas flake8 linting is set to 180...

tests/compas/datastructures/test_mesh.py
```diff
@@ -499,7 +566,9 @@ def test_vertex_laplacian():
 def test_vertex_neighborhood_centroid():
     mesh = Mesh.from_obj(compas.get('faces.obj'))
     assert mesh.vertex_neighborhood_centroid(0) == [1.0, 1.0, 0.0]
-    assert mesh.vertex_neighborhood_centroid(1) == [2.0, 0.6666666666666666, 0.0]
+    assert mesh.vertex_neighborhood_centroid(1) == [
```
tomvanmele: split?

tests/compas/datastructures/test_mesh.py
```diff
@@ -470,13 +534,16 @@ def test_normal():
     assert mesh.normal() == [0.0, 0.0, 0.0]
 
     mesh = Mesh.from_obj(compas.get('quadmesh.obj'))
-    assert mesh.normal() == [-2.380849234996509e-06, 4.1056122145028854e-05, 0.8077953732329284]
+    assert mesh.normal() == [
```
tomvanmele: same?

src/compas/files/base_reader.py
```diff
@@ -0,0 +1,103 @@
...
+        Path or URL to the file
+    """
+    def __init__(self, location):
+        self.is_url = False
+        self.location = self.check_location(location)
+
+    def read(self, mode='text'):
+        """Returns iterable reading file line by line when given a location,
+           either as pathlib object, string containing path or string containing url
+        Arguments
```
tomvanmele: since we shoot for a docstring style that is as readable as possible in code as well i would advocate for a newline here

tomvanmele: i personally also prefer "Parameters" over "Arguments"

src/compas/files/base_reader.py
```diff
@@ -0,0 +1,103 @@
...
+        ---------
+        mode : string
+            Treat file as ascii or binary
+        Yields
+        -------
+        string
+            yields next line
```
tomvanmele: "yields next line" => "next line"

src/compas/files/base_reader.py
```diff
@@ -0,0 +1,103 @@
...
+                resp = urlopen(self.location)
+                for line in resp.read().decode('utf-8').split('\n'):
+                    yield line
+            else:
+                try:
+                    with self.location.open(mode='r') as fh:
+                        for line in fh:
+                            yield line
+                except UnicodeDecodeError:
+                    with self.location.open(mode='r',
```
tomvanmele: max-line-length flake8 is set to 180

src/compas/files/base_reader.py
```diff
@@ -0,0 +1,103 @@
...
+    """Base class containing file reading functions for file extension specific readers
+
+    Attributes
+    ----------
+    location : str or pathlib object
+        Path or URL to the file
+    """
+    def __init__(self, location):
+        self.is_url = False
+        self.location = self.check_location(location)
```
tomvanmele: would this not be a perfect use case for `@property`?

src/compas/files/base_reader.py
tomvanmele: where is this file signature defined? it doesn't seem to have a default on the base reader and obj reader (for example) doesn't seem to define it either?

src/compas/files/off.py
tomvanmele: in general i think it is confusing that `read` does something else then what the corresponding method of a standard file object does 

https://docs.python.org/3/tutorial/inputoutput.html#methods-of-file-objects

src/compas/files/base_reader.py
```diff
@@ -0,0 +1,103 @@
...
+
+    def read(self, mode='text'):
```
tomvanmele: i find it confusing that the `read` method does something else than the default method of a file object

src/compas/files/obj.py
```diff
@@ -1,15 +1,18 @@
...
 from collections import OrderedDict
-
 try:
     from urllib.request import urlopen
 except ImportError:
     from urllib2 import urlopen
+try:
+    from pathlib import Path
+except ImportError:
+    from pathlib2 import Path
```
tomvanmele: is this necessary?
should path related functionality not be handled by the base reader?

src/compas/files/off.py
```diff
@@ -15,88 +11,66 @@
...
-           Available at: http://paulbourke.net/dataformats/obj/.
+    vertex_count : int
+        Vertex count stated in beginning of file
+    face_count : int
+        Face count stated in beginning of file
+    edge_count : int
+        Edge count stated in beginning of file
 
     """
+    file_signature = {
```
tomvanmele: only reader where this is defined as far as i can tell

src/compas/files/off.py
```diff
@@ -142,7 +112,8 @@ def read(self):
                 continue
 
             if len(parts) == 3:
-                self.v, self.f, self.e = int(parts[0]), int(parts[1]), int(parts[2])
+                self.vertex_count, self.face_count, self.edge_count = \
```
tomvanmele: in data structures these are called `number_of_...`

src/compas/files/ply.py
```diff
@@ -20,16 +21,22 @@ class PLY(object):
...
 
-class PLYReader(object):
+class PLYReader(BaseReader):
     """"""
 
-    keywords = ['ply', 'format', 'comment', 'element', 'property', 'end_header']
+    file_signature = {
+                      'content': b'ply',
+                      'offset': 0,
+    }
```
tomvanmele: not a big fan of this indentation style

tomvanmele (commented):
there is a lot going on in this PR.
perhaps we should focus on the base reader first?

src/compas/files/off.py
tetov: EDIT: Do you have any name suggestions?

src/compas/files/ply.py
tetov: It did before, but I think this should be changed to just ignore user-defined elements.

tests/compas/datastructures/test_mesh.py
tetov: Somethings up with my config, will fix.

tests/compas/datastructures/test_mesh.py
tetov: Somethings up with my config, will fix.

src/compas/files/base_reader.py
tetov: File signatures:
* ply: `b'ply'`
* stl: ascii has`b'SOLID'`, binary has a header..
* off: `b'OFF'`

Not all formats have them but for the ones that do it could be a good check before trying to load a file.

Needs more testing as well.

tests/compas/files/test_stl.py
gonzalocasas: Yes, that would make more sense, keep the test as small in scope as possible (if a test needs to construct other stuff only to trigger an underlying behavior, something is fishy ;) )

src/compas/files/ply.py
gonzalocasas: +1

tests/compas/datastructures/test_mesh.py
```diff
@@ -42,32 +46,78 @@ def test_from_obj():
...
     assert mesh.number_of_vertices() == 8
     assert mesh.number_of_edges() == 18
 
+    # testing ascii stl and pathlib integration
+    mesh = Mesh.from_stl(compas.get('cube_ascii.stl'))
+    assert mesh.number_of_faces() == 8016
+    assert mesh.number_of_vertices() == 4020
+    assert mesh.number_of_edges() == 11368
+
+    mesh = Mesh.from_stl(compas.get('binary_w_ascii_header.stl'))
```
gonzalocasas: I see now this thing... test files should be fixtures, not `data`, so I would revert the file moving, and load them using the fixture feature of `pytest`

src/compas/files/base_reader.py
gonzalocasas: Technically, the summary should be one-line only. Then you can have a new-line and expand on it with further detail.

src/compas/files/base_reader.py
gonzalocasas: if this is literally read line-by-line, then why not call this `read_line` and keep the rest of the names as `read` ?

gonzalocasas (commented):
Added some early feedback, but I assume you're still working on it because it's marked as draft, right?

src/compas/files/obj.py
tetov: True. Fixed :).

src/compas/files/off.py
tetov: `raise Exception('Failed to parse file as an OFF-file')`?

src/compas/files/off.py
tetov: Might be nice to rename them for consistency. On the other hand, after set they are not used again so maybe remove them all together?

src/compas/files/ply.py
tetov: Fixed.

src/compas/files/base_reader.py
tetov: Like this?

https://github.com/compas-dev/compas/blob/d27f338942d4d5109ad888c80534bd1ca8398193/src/compas/files/base_reader.py#L40

src/compas/files/base_reader.py
tetov: Renamed it read_line_or_chunk but that feels quite clunky too. iterate_over_file?

src/compas/files/base_reader.py
tetov: Ok, good points, fixed.

https://github.com/compas-dev/compas/blob/d27f338942d4d5109ad888c80534bd1ca8398193/src/compas/files/base_reader.py#L61

src/compas/files/base_reader.py
tetov: Reworked method a bit now.
https://github.com/compas-dev/compas/blob/d27f338942d4d5109ad888c80534bd1ca8398193/src/compas/files/base_reader.py#L101

And I think I understand what you meant now @gonzalocasas. Downloading file to tmp and then iterating over it would make the whole function more streamlined.

Do you agree with my reasoning in method docstring? Useful to keep the check for file formats with file signatures?

src/compas/files/obj.py
tetov: No, must have tabbed by mistake. Haven't really touched this part.

src/compas/files/off.py
tetov: https://github.com/compas-dev/compas/pull/418/files/0682dcf5876f08b75e93f31d36a8e50292a30a8b#r355006802

src/compas/files/ply.py
tetov: Changed. Should probably set up a test file to make sure it passes.

tetov:
@gonzalocasas 
> Added some early feedback, but I assume you're still working on it because it's marked as draft, right?

I'll flag you when another review is needed, I need to move the tests and and resolve your comments above.

### There are some open questions above I'd like answered as well :).

- [x] [Keep file signature check?](https://github.com/compas-dev/compas/pull/418#discussion_r355008896)
- [x] ~~[Rename or remove count of vertices, faces, edges?](https://github.com/compas-dev/compas/pull/418#discussion_r354950429)~~
- [x] [New name for BaseReader.read()](https://github.com/compas-dev/compas/pull/418#discussion_r355006802)
- [ ] ~~[Good rewording of error message?](https://github.com/compas-dev/compas/pull/418#discussion_r354948649)~~

### Todo list for me:

- [ ] ~~Move and rework tests in test_mesh that really belong to the file readers~~
- [ ] ~~Test if ply.py will ignore user-defined elements~~
- [x] ~~[Follow up weird edit, (just check git blame)](https://github.com/compas-dev/compas/pull/418#discussion_r355009197)~~
- [x] [Download files to tmp when given URL and iterate over contents](https://github.com/compas-dev/compas/pull/418#discussion_r355008896)
- [x] Docstrings and comments

src/compas/files/base_reader.py
tomvanmele: `iterlines`?

gonzalocasas: Instead of having one method that does two things and receives a `mode` to `if` on it, why not have two methods properly named: `read_line()` and `read_bytes()` and ditch the mode parameter?

tetov:
Yeah, totally agree. iterlines and iterchunks?

tetov:
Or read_line & read_chunks. Sorry missed last part of the comment.

tomvanmele:
i think this PR is too big. perhaps we should split off the base reader and get that sorted first. also, perhaps this is useful for binary/not binary https://github.com/audreyr/binaryornot

tetov:
I've moved my other changes to another branch on my fork to make the scope of this PR more limited.

I'll work through the remaining pointers for base_reader and update the PR (tonight hopefully).

### To add to this PR:
- [x] Download file to tmp and start iterating over lines
- [x] Tests
- [x] Docstrings
- [x] Binary detection using library @brgcode suggested (https://github.com/audreyr/binaryornot)

### Future PR's:
* Update file reader modules to use base_reader (one PR per module)

src/compas/files/__init__.py
```diff
@@ -13,9 +13,12 @@
     :toctree: generated/
     :nosignatures:
 
+    BaseReader
     OBJ
     OBJReader
     OBJParser
+    OFF
+    OFFReader
```
tomvanmele: i guess these need to be removed for now

src/compas/files/base_reader.py
```diff
@@ -0,0 +1,141 @@
...
+                pathobj = Path(location)
+            else:
+                pathobj = location
+
+            if pathobj.exists():
+                self._location = pathobj
+            else:
+                raise IOError('File not found.')
+
+    def read_line_or_chunk(self, mode='ascii'):
```
tomvanmele: i think something like `iter_lines` would make more sense.
```python
for line in reader.iter_lines():
    # do something with line
```

src/compas/files/base_reader.py
```diff
@@ -0,0 +1,141 @@
...
+        Yields
+        -------
+        string
+            Next line or chunk of file
+        """
+        # TODO: Handle url's to binary files
+        # TODO: Handle continuing lines (as in OFF files)
+        # TODO: Read binary chunks
+
+        self.check_file_signature()
```
tomvanmele: `binaryornot` checks for binary files in a completely different way. perhaps it might help solve the problems with the PLY files.

src/compas/files/base_reader.py
```diff
@@ -0,0 +1,141 @@
...
+
+class BaseReader(object):
+    """Base class containing file reading functions for file extension specific readers
+
+    Attributes
+    ----------
+    location : str or pathlib object
+        Path or URL to the file
+    """
+
```
tomvanmele, line 24: should the base reader not define the interface for default methods of the other readers as well?
* `read`
* ...

tomvanmele: with a `NotImplementedError`

src/compas/files/base_reader.py
```diff
@@ -0,0 +1,141 @@
...
+                resp = urlopen(self.location)
+                for line in resp.read().decode('utf-8').split('\n'):
+                    yield line
+            else:
+                try:
+                    with self.location.open(mode='r') as fh:
+                        for line in fh:
+                            yield line
+                except UnicodeDecodeError:
+                    with self.location.open(mode='r', errors='replace', newline='\r') as fh:
```
tomvanmele: should this not be moved to `open`?

tomvanmele (changes requested):

src/compas/files/__init__.py
tetov: Ok!

src/compas/files/base_reader.py
tetov: Do you think we should remove the file_signature check?

src/compas/files/base_reader.py
tetov: Sure!

src/compas/files/base_reader.py
tomvanmele: no because otherwise there is no way to choose between the ascii and binary reading functionality. unless we leave that up to the user.

i would just look into their way of checking and see if it can be used in the background here.

also, perhaps signature reading should not be directly responsible for deciding between the two reading methods. perhaps this should be something like `is_binary` which uses information from the signature.

tetov: Ok! I'll leave this and will flesh it out while working on the specific readers.

src/compas/files/base_reader.py
tetov: `read` added. 

`is_valid` should be added too. 

Planning to write a `is_binary` that file readers can call.

src/compas/files/base_reader.py
tetov: `_open` so there's no conflict. Moved it out of `iter_lines`.

src/compas/files/base_reader.py
tetov: Changed.

setup.cfg
```diff
@@ -9,6 +9,9 @@ exclude = */migrations/*
 max-line-length = 180
 ignore = D001
 
+[pydocstyle]
+convention = numpy
```
tomvanmele, line 13: nice

src/compas/files/base_reader.py
```diff
@@ -18,85 +18,164 @@ class BaseReader(object):
     ----------
     location : str or pathlib object
         Path or URL to the file
-    """
+    is_url
```
tomvanmele: ?

src/compas/files/base_reader.py
```diff
@@ -18,85 +18,164 @@ class BaseReader(object):
...
+
+    Methods
+    -------
```
tomvanmele: this is not necessary because sphinx does this automatically

src/compas/files/base_reader.py
```diff
@@ -18,85 +18,164 @@ class BaseReader(object):
...
+            True if recognized as an URL
+        """
+        self._is_url = str(self._location).startswith('http')
+        return self._is_url
+
+    @property
+    def is_binary(self):
+        pass
+
+    @property
```
tomvanmele, line 82: you have a function and a property with the same name

src/compas/files/base_reader.py
```diff
@@ -18,85 +18,164 @@ class BaseReader(object):
...
+        if pathobj.exists():
+            return pathobj
 
-    def read_line_or_chunk(self, mode='ascii'):
-        """Returns iterable reading file line by line
+        raise IOError('File not found.')
 
-        Should accept both local files as well as URLs, in binary or ascii format.
+    @property
+    def is_url(self):
```
tomvanmele: `is_url` is not a property of the base reader. as mentioned in the docstring, it is a property of the location...

tests/compas/files/test_base_reader.py
```diff
@@ -0,0 +1,21 @@
+import os
+try:
+    from pathlib import Path
+except ImportError:
+    from pathlib2 import Path
```
tomvanmele, line 5: is this necessary?

tomvanmele (changes requested):

src/compas/files/base_reader.py
tetov: Ok!

tests/compas/files/test_base_reader.py
tetov: Not yet but I planned to give it path objects in tests.

src/compas/files/base_reader.py
tetov: Created open_ascii and open_binary.

src/compas/files/base_reader.py
tetov: Changed to normal method.

tests/compas/files/test_base_reader.py
tetov: Planning to add tests using path objects.

tomvanmele:
test fails on number of lines check of ply_ascii

tetov:
> test fails on number of lines check of ply_ascii

Working on it :)

src/compas/files/base_reader.py
tomvanmele: that is not what i meant:
```python
reader.is_url()
```
doesn't make any sense...

tomvanmele: better would be
```python
reader.is_location_url()
```
or 
```python
reader.location.is_url()
```

tetov: Ok, sorry, with you now.

Do you think it would make sense to create a new class for location?

src/compas/files/base_reader.py
```diff
@@ -0,0 +1,188 @@
...
+    def is_location_url(self):
+        """Checks if given location is a string containing an URL
+
+        Returns
+        -------
+        bool
+            True if recognized as an URL
+        """
+        self._is_url = str(self._location).startswith('http')
+        return self._is_url
```
tomvanmele: i think this should be 
```python
return str(self._location).startswith('http')
```
that said, i think it is a bad idea to use `self.location` and `self._location` in this way, because it is something you need to know to do it correctly.
perhaps it would be better to differentiate between the original address provided by the user, which could be a url, a relative path, an absolute path, and a `Path` object. the reader processes this into a new variable that always refers to an absolute local file path in the form of a `Path` object. assuming the original address provided by the user is stored as `self._address` this function would just do
```python
return str(self._address).startswith('http')
```

src/compas/files/base_reader.py
```diff
@@ -0,0 +1,188 @@
...
+
+        Returns
+        -------
+        location : Pathlib object
+            Path to tempfile
+        """
+
+        location, _ = urlretrieve(url)
+
+        return Path(location)
```
tomvanmele, line 113: after differentiating between "address" and "location" (or whatever the naming convention ends up being) this could be
```python
location, _ = urlretrieve(self._address)
return Path(location)
```

src/compas/files/base_reader.py
```diff
@@ -0,0 +1,188 @@
...
+        # TODO: Integrate with is_binary checks in file readers
+        # TODO. Look into binaryornot library to help with above
+
+        try:
+            file_signature = self.FILE_SIGNATURE['content']
+            signature_offset = self.FILE_SIGNATURE['offset']
+
+        # if file type has no associated file signature
+        except AttributeError:
+            return
```
tomvanmele: the `BaseReader` should define defaults for both parameters.
```python
class BaseReader(object):
    FILE_SIGNATURE = {
         'content': None,
         'offset': None,
    }
```
after which the above becaomes a simple if statement.

src/compas/files/base_reader.py
tomvanmele: i would also argue that `download` should be "underscored"

src/compas/files/base_reader.py
```diff
@@ -20,8 +22,14 @@ class BaseReader(object):
...
-    def __init__(self, location):
-        self._location = location
+
+    FILE_SIGNATURE = {
+         'content': None,
+         'offset': None,
+    }
+
+    def __init__(self, adress):
+        self._adress = adress
```
tomvanmele: address has two d's

src/compas/files/base_reader.py
```diff
@@ -30,48 +38,53 @@ def location(self):
...
+        Returns
         ------
-        IOError
-            If file is not found
+        Path object
+            Path object pointing to file location
         """
-        if self.is_location_url():
-            pathobj = self.download(self._location)
+        if self.is_adress_url():
```
tomvanmele: two d's

src/compas/files/base_reader.py
```diff
@@ -30,48 +38,53 @@ def location(self):
...
+            True if binary else false.
+        """
+        return binaryornot.check.is_binary(str(self.location))
 
     @property
     def is_valid(self):
         return NotImplementedError
 
-    def is_location_url(self):
+    def is_adress_url(self):
```
tomvanmele: two d's

src/compas/files/base_reader.py
```diff
@@ -10,6 +10,8 @@
 except ImportError:
     from urllib import urlretrieve
 
+import binaryornot.check
```
tomvanmele, line 13: i think this is a good idea, but we should check when it identifies a file as binary.
from reading through the code quickly, it seems that if more than 5 % of the code is non-ascii it is considered binary. while probably sufficient in real-world examples, ply headers (ascii) can be quite long even if the described object is very simple. not sure if this then would pass the test. but perhaps we should cross that bridge when we get there... :)

src/compas/files/base_reader.py
```diff
@@ -129,8 +139,8 @@ def iter_lines(self):
             for line in fo:
                 yield line.rstrip()
 
-    def iter_chunks(self, chunk_size=1024):
-        """Yields lines from local binary files
+    def iter_chunks(self, chunk_size=4096):
```
tomvanmele, line 151: does the chunk size in any way depend on the type of binary encoding?

src/compas/files/base_reader.py
```diff
@@ -158,31 +170,33 @@ def check_file_signature(self):
...
-        # TODO. Look into binaryornot library to help with above
-
-        try:
-            file_signature = self.FILE_SIGNATURE['content']
-            signature_offset = self.FILE_SIGNATURE['offset']
 
-        # if file type has no associated file signature
-        except AttributeError:
+        if self.FILE_SIGNATURE['content'] is None:
             return
```
tomvanmele: should this not be `False`

src/compas/files/base_reader.py
tetov: The standard value should probably be removed.

I'm not sure of how this should work, I'm not experienced in reading binary files. But what I was going for was that the file reader specifies a reasonable chunk size and then reads amount of bytes corresponding to format spec.

src/compas/files/base_reader.py
tetov: Yes, thanks!

src/compas/files/base_reader.py
tetov: Yeah, I think that we can have general checks in BaseReader while some more specific should be in the file reader. 

A well formatted binary PLY should have `format binary_little_endian 1.0` or 
`format binary_big_endian 1.0` in the next line. :)

tetov: > i think this is a good idea [...]

It was your idea ;).

src/compas/files/base_reader.py
tetov: Not in Swedish, haha! Fixed.

src/compas/files/base_reader.py
tomvanmele: fair enough

tetov: Do you have a chunk size you would like to have there instead or just change it to an arg?

tomvanmele: no leave it as it is

tetov:
Is it time to squash the commits and remove the WIP status?

src/compas/files/base_reader.py
```diff
@@ -0,0 +1,202 @@
...
+        file, a missing file signatures might be a sign of a malformed file.
+
+        More information about file signatures can be found on Wikipedia[1]_
+        as well as examples of file signatures[2]_
+
+        Returns
+        ------
+        bool
+            True file signature for file type is found in file or if file type
+            lacks file signature.
```
tomvanmele: True if correct file signature for file type is found in file, or if file type has no signature.

tomvanmele (commented):
nearly there

tomvanmele:
last changes and then we can merge

tetov:
> last changes and then we can merge

I got a warning from autosummary because it couldn't find BaseReader. I guess it's because it's not imported in __init__.py. I removed the entry.

Also, I tried interactive rebase but since I merged with master earlier I'm worried about squashing other commits as well. Maybe you could use [Github's solution](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/about-pull-request-merges#squash-and-merge-your-pull-request-commits) with my last commit msg?

PR #421, opened 2019-12-11
edited _mesh.set_edge_attributes()

mpopescu:
edited _mesh.set_edge_attributes() to add (v, u) entry to edgedata if it doesn't exist

### Checklist


tomvanmele:
can you explain what exactly the problem is? where does this fail or produce an unexpected result? because, having looked at it more closely, the code seems fine to me. there is no absolute need for setting the attributes in both directions because the `get_` functions perform that check...

PR #422, opened 2019-12-11
Added Polygon.from_sides_and_radius_xy()

arpastrana:

### Checklist

### Description

`classmethod` to create a regular (equilateral and equiangular) `Polygon`.

```
    from compas_plotters import MeshPlotter

    polygons = []
    points = []
    for i in range(3, 10):
        polygon = Polygon.from_sides_and_radius(i, 1)
        _p = polygon.points 
        polydict = {'points': _p + _p[:1], 'facecolor': (255, 0, 0), 'radius':0.03}
        polygons.append(polydict)
        
        for point in polygon.points:
            points.append({'pos': point, 'facecolor': (255, 0, 0), 'radius':0.03})
        
    # plot
    plotter = MeshPlotter(Mesh(), figsize=(12, 9))
    plotter.draw_polylines(polygons)
    plotter.draw_points(points)
    plotter.show()
```

Screenshot 2019-12-11 at 15 54 03


tomvanmele:
i like the method in general. have a few suggestions:

* rename to `Polygon.from_circle`
* change signature to `from_circle(circle, n)`

by basing the polygon on a circle it can be positioned anywhere in space...

arpastrana:
Made the suggested changes. The only _disadvantage_ I see, is that since the `Circle` orientation depends on a `Plane` and not a `Frame`, the polygon's rotation on the circle's plane is not always explicit.

tomvanmele:
true. should we add a frame instead?

arpastrana:
a. Changing `Circle` so that it is defined instead by a `Frame` and a `radius`

```
circle = Circle(Frame.worldXY(), 0.5)
```

b. In Polygon.from_circle we do a frame to frame transformation based on `Frame.from_plane()`
```
T = Transformation.from_frame(Frame.from_plane(circle.plane))
return cls(transform_points(points, T))
           
```

Would you favor any of the these options?

tomvanmele:
option b

arpastrana:
Reworked it along the lines of option b. 

However, `Frame.from_plane()` still selects arbitrarily the x an y axes. Therefore, an optional parameter `vector` is introduced to explicitly choose the in-plane alignment.

src/compas/geometry/primitives/polygon.py
```diff
@@ -83,6 +88,70 @@ def from_data(cls, data):
...
+        True
+        >>> dot_vectors(pentagon.normal, circle.normal) == 1
+        True
+        >>> centertofirst = subtract_vectors(pentagon.points[0], circle.center)
+        >>> dot_vectors(centertofirst, vec) == circle.radius
+        True
+        """
+        def check_number_sides():
+            if n < 3:
+                raise ValueError("Supplied number of sides must be at least 3!")
```
tomvanmele: why not just add this as normal statements? don't see an advantage of putting this into a function def. you could make it a one-liner with `assert`...

arpastrana: True. It's also more readable. Will change it.

arpastrana:
Following up, do I revert this PR to d1eaab3  so that the user is responsible for the orientation of the polygon , or we leave it as it is (plus the `assert` one-liner)?

tomvanmele:
revert

arpastrana:
Now reverted and included the `assert`.

PR #446, opened 2019-12-17
Add gltf import support

beverlylytle:

Adds functionality to import the tree structure and data of a glTF file, including the attached meshes.


src/compas/files/gltf.py
```diff
@@ -0,0 +1,162 @@
...
+    for name in meshes:
+        meshes[name].name = name
+
+    scene_tree = Network.from_vertices_and_edges(gltf.parser.scene_data['vertices'], gltf.parser.scene_data['edges'])
+    scene_tree.plot()
+
+    # Usage of MultiMeshViewer unclear
+    # viewer = MultiMeshViewer()
+    # viewer.meshes = meshes.values()
+    # viewer.show()
```
tomvanmele: ?

beverlylytle: I could clarify with "Usage of MultiMeshViewer unclear _to me_".  This seems to follow the pattern established in stl.py, but when I tried to run stl.py and gltf.py, it didn't work.  I left this in here so that some one might explain viewers to me (but would be removed/fixed before merging).

tomvanmele: okay, nor problem. do you have `compas_viewers` installed?

beverlylytle: yes

tomvanmele: also, in py3, `keys` and `values` are iterators by default so you may want to convert the output to an actual list
```python
viewer.meshes = list(meshes.values())
```

tomvanmele: https://compas-dev.github.io/examples/mesh-subd-comparison/mesh-subd-comparsion.html

beverlylytle: It seems that the import statement in stl.py was outdated and should be replaced with 
    `from compas_viewers.meshviewer import MeshViewer`
This is what was causing me confusion.

beverlylytle:
This PR will be updated substantially some time tomorrow.  Please hold off reviewing until then.

src/compas/files/gltf.py
```diff
@@ -0,0 +1,629 @@
...
+    'GLTFReader',
+    'GLTFParser',
+]
+
+import base64
+import io
+import json
+import os
+import struct
+from math import fsum
```
gonzalocasas: Totally minor comment, but perhaps best to just `import math` and use `math.fsum` (I didn't recognize it as a math module function when I saw it used)

src/compas/files/gltf.py
```diff
@@ -0,0 +1,629 @@
...
+        Name of the node, if any.
+    children : list
+        Child nodes referenced by node_key.
+    matrix : list of lists
+        Matrix representing the displacement from node's parent to the node.
+    mesh_index : int or str double check this
+        Index of the associated mesh within the JSON.
+    weights : list of ints
+        Weights used for computing morph targets in the attached mesh, if any.
+
```
gonzalocasas: not sure what sphinx will do with this blank line, it might think the list of attributes is over and the rest is description.

src/compas/files/gltf.py
```diff
@@ -0,0 +1,629 @@
...
+    """Object representing a root node of a scene.
+
+    Attributes
+    ----------
+    name : str
+        Name of the scene, if any.
+    edges : list
+        List of 2-tuples referencing nodes by key.
+    nodes : dict
+        Dictionary of (node_key, GLTFNode) pairs.
```
gonzalocasas:
```suggestion
        Dictionary of (node_key, :class:`GLTFNode`) pairs.
```
Here we can use the class directive thingy to have these linked/clickable on the doc.

src/compas/files/gltf.py
```diff
@@ -0,0 +1,629 @@
...
+
+        self._content = None
+        self._glb_buffer = None
+        self._buffers = {}
+
+        self.read()
+
+    def read(self):
+        with open(self.filepath, 'rb') as f:
+            self._content = memoryview(f.read())
```
gonzalocasas, line 195: Maybe wrap all this block until line ~206 in a `try..finally`, and call `self._content.release()` before nullifying the variable in the `finally` block to make sure we don't keep any dangling reference.

src/compas/files/gltf.py
```diff
@@ -0,0 +1,629 @@
...
+                self.image_data.append(image_data)
+
+    def get_image_data(self, image):
+        if 'bufferView' not in image:
+            return None
+        buffer_view_index = image['bufferView']
+        buffer_view = self.json['buffersViews'][buffer_view_index]
+        buffer = self.get_buffer(buffer_view['buffer'])
+        offset = buffer_view.get('byteOffset', 0)
+        length_ = buffer_view['byteLength']
```
gonzalocasas: Why is length suffixed with an underscore? Some kind of reserved keyword?

src/compas/files/gltf.py
```diff
@@ -0,0 +1,629 @@
...
+
+        # load json
+        type_, length_, json_bytes, offset = self.load_chunk(offset)
+        if type_ != b'JSON':
+            raise Exception('Bad glTF.  First chunk not in JSON format')
+        json_str = json_bytes.tobytes().decode('utf-8')
+        self.json = json.loads(json_str)
+
+        # load binary buffer
+        if offset < len(self._content):
```
gonzalocasas, line 239: If that is not true, should we raise an exception or is it ok to just ignore?

src/compas/files/gltf.py
```diff
@@ -0,0 +1,629 @@
...
+        chunk_header = struct.unpack_from('<I4s', self._content, offset)
+        length_ = chunk_header[0]
+        type_ = chunk_header[1]
+        data = self._content[offset + 8: offset + 8 + length_]
+
+        return type_, length_, data, offset + 8 + length_
+
+    def check_version(self):
+        version = self.json['asset']['version']
+        if float(version) != 2.0:
```
gonzalocasas: Wouldn't it be safer to keep this check as a string, because in the future, a theoretical patch version (eg. `2.0.1`) would break the float cast.

src/compas/files/gltf.py
```diff
@@ -0,0 +1,629 @@
...
+            )
+
+            for index, data_index in enumerate(sparse_indices):
+                data[data_index] = sparse_values[index]
+
+            if accessor.get('normalized', False):
+                for index, tuple_ in enumerate(data):
+                    new_tuple = ()
+                    for i in tuple_:
+                        if component_type == 5120:
```
gonzalocasas: These component types (`5120, 5121, 5122, 5123`) would be a better if we declare as constants at the top, maybe changing the `COMPONENT_TYPE_ENUM` to use it, something like:
```python
COMPONENT_TYPE_BYTE = 5120
COMPONENT_TYPE_UBYTE = 5121
..
COMPONENT_TYPE_ENUM = {
    COMPONENT_TYPE_BYTE: 'b',  # BYTE
    COMPONENT_TYPE_UBYTE: 'B',  # UNSIGNED_BYTE
    ..
}
```

src/compas/files/gltf.py
```diff
@@ -0,0 +1,629 @@
...
+    """Parse the contents of the reader to read and parse the associated binary files
+    and create objects digestible by compas.
+
+    Parameters
+    ----------
+    reader : GLTFReader
+
+    Attributes
+    ----------
+    reader : GLTF2Reader
```
gonzalocasas:
```suggestion
    reader : :class:`GLTFReader`
```

src/compas/files/gltf.py
```diff
@@ -0,0 +1,629 @@
...
+    default_scene_index : int
+        index of the default scene
+    scenes : list of GLTFScenes
```
gonzalocasas:
```suggestion
    scenes : list of :class:`GLTFScene`
```

src/compas/files/gltf.py
```diff
@@ -0,0 +1,629 @@
...
+        Weights used for computing morph targets in the attached mesh, if any.
+
+    position : tuple
+        xyz-coordinates of the node, calculated from the matrix.
+    transform : list of lists
+        Matrix representing the displacement from the root node to the node.
+    mesh_data : MeshData
+        Contains mesh data, if any.
+    node_key : int
+        Key of the node used in GLTFScene.nodes.
```
gonzalocasas: There's a way to directly link to the method, but I don't remember exactly the syntax, is something like :attr:`GLTFScene.nodes`...

src/compas/files/gltf.py
```diff
@@ -0,0 +1,629 @@
...
+
+        if not uri:
+            buffer = self._glb_buffer
+        elif self.is_data_uri(uri):
+            string = self.get_data_uri_data(uri)
+            buffer = memoryview(base64.b64decode(string))
+        else:
+            filepath = self.get_filepath(uri)
+            with open(filepath, 'rb') as f:
+                buffer = memoryview(f.read())
```
gonzalocasas, line 392: These buffers are tricky to release correctly, perhaps the `GLTFReader` needs to have a `release` method itself, that is called once reading is completed. However, I'm not sure if reading is completed in our context or in the user's. If that is the case, maybe the `GLTF` class itself needs a way of releasing, and the user is in charge of calling it to trigger all deallocations.

src/compas/files/gltf.py
```diff
@@ -0,0 +1,629 @@
...
+    name : str
+        Name of the scene, if any.
+    edges : list
+        List of 2-tuples referencing nodes by key.
+    nodes : dict
+        Dictionary of (node_key, GLTFNode) pairs.
+    """
+    def __init__(self):
+        self.name = None
+        self.edges = []
```
gonzalocasas: Shouldn't we keep edges in a dictionary? I assume it's now a list to make it easy to create a network afterward, but if used standalone, it would be faster to use dictionary lookups instead of iterating over a list.

src/compas/files/gltf.py
```diff
@@ -0,0 +1,629 @@
...
+        weights = mesh.get('weights', None)
+        weights = gltf_node.weights or weights
+        return weights
+
+    def get_morph_function(self, w):
+        # Returns a function which computes for a fixed list w of scalar weights the linear combination
+        #                               vertex + sum_i(w[i] * targets[i])
+        # where vertex and targets[i] are vectors.
+        def apply_morph_target(vertex, *targets):
+            return tuple(map(lambda v, *t: v + fsum(map(lambda a, b: a * b, w, t)), vertex, *targets))
```
gonzalocasas: That's .. intense ;)  I don't know if it can be simplified for readability, maybe if you assign the lambdas to a variable before, then it can be more "humanly parsable" 😄

gonzalocasas (approved):
LGTM

Looks excellent! I added a few comments, but overall, it's great. I have not tested it thou, but since I know you are still planning on making a few more changes for IronPython compat, I'll test after that! ;)

src/compas/files/gltf.py
beverlylytle: I warned you... Worst line of code ever.

src/compas/files/gltf.py
beverlylytle: I'm a little confused as to what you have in mind for this dictionary.  Do you mean something of the form {parent_key: <list of children's keys>}? (This sort of thing is already recoverable from the nodes.) Or do you mean that the edge tuples should be stored as a set rather than a list?  I am almost tempted to delete this attribute completely because I already expect the user to create the appropriate dictionary for the vertices.  Why wouldn't I also expect this for the edges?

src/compas/files/gltf.py
beverlylytle: ok

src/compas/files/gltf.py
beverlylytle: Clearly I was faking it went it came to anything sphinx related...

src/compas/files/gltf.py
beverlylytle: :meth:GLTFScene.nodes ?

src/compas/files/gltf.py
beverlylytle: This is checking whether there is a second chunk of binary data.  It may be that everything is stored as a data uri and so this chunk wouldn't be there.

src/compas/files/gltf.py
beverlylytle: or :func: ?

src/compas/files/gltf.py
beverlylytle: What about appending the following?
```
try:
    self._content.release()
except AttributeError:
    pass
finally:
   self._content = None
```
Or is there another exception you are thinking that should be caught earlier within this block?

src/compas/files/gltf.py
tetov: :meth

(Had to look this up recently. Here’s docs re hyperlinking in Sphinx for python code: https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#cross-referencing-python-objects)

src/compas/files/gltf.py
```diff
@@ -0,0 +1,657 @@
...
+
+        byte_stride = buffer_view.get('byteStride', expected_length)
+
+        unpack_from = struct.Struct(format_).unpack_from
+
+        buffer_index = buffer_view['buffer']
+        buffer = self.get_buffer(buffer_index)
+
+        data = [
+            unpack_from(buffer[i: i + byte_stride].tobytes())
```
beverlylytle, line 369: Slicing the memoryview and calling tobytes() is the change I had to make for IronPython.

src/compas/files/gltf.py
gonzalocasas: I didn't mean so much the case for exception handling as the disposal/release of the buffer, but I see you added a few methods to release buffers, so I'll mark this as resolved.

src/compas/files/gltf.py
gonzalocasas: 👍

src/compas/files/gltf.py
gonzalocasas: 👍

src/compas/files/gltf.py
```diff
@@ -0,0 +1,657 @@
...
+
+    def get_weights(self, mesh, gltf_node):
+        weights = mesh.get('weights', None)
+        weights = gltf_node.weights or weights
+        return weights
+
+    def get_morph_function(self, weights):
+        # Returns a function which computes for a fixed list w of scalar weights the linear combination
+        #                               vertex + sum_i(w[i] * targets[i])
+        # where vertex and targets[i] are vectors.
```
gonzalocasas, line 573: 👍 much easier to digest!

src/compas/files/gltf.py
gonzalocasas: but `nodes` is an attribute, not a method, isn't it?

beverlylytle: Yes, I've used :attr:.

PR #449, opened 2019-12-20
compas.files.base_reader raise if bad address is given

tetov:
While working on OBJReader I noticed a mistake in BaseReader. If address to file is a string or Path object containing a non resolvable path methods fail silently, when they should in fact raise FileNotFoundError (or OSError in Python 2).

Property location in BaseReader now checks the path with Path method resolve(strict=True) instead of exists.

I've also added a test case for this.

### Checklist

1 .~~[ ] Add the change to the `CHANGELOG.md` file in the `Unreleased` section under the most fitting heading: `Added`, `Changed`, `Removed`.~~ (base_reader is still under unreleased - added in changelog)

### DIFF
--- a/src/compas/files/base_reader.py
+++ b/src/compas/files/base_reader.py
@@ -64,8 +64,10 @@ class BaseReader(object):
             else:
                 pathobj = self._address
 
-        if pathobj.exists():
-            return pathobj
+        # Makes path absolute and raises FileNotFound if file is not found
+        pathobj.resolve(strict=True)
+
+        return pathobj
 
     @property
     def is_binary(self):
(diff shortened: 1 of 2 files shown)


src/compas/files/base_reader.py
```diff
@@ -64,8 +64,10 @@ def location(self):
             else:
                 pathobj = self._address
 
-        if pathobj.exists():
-            return pathobj
+        # Makes path absolute and raises FileNotFound if file is not found
+        pathobj.resolve(strict=True)
```
tomvanmele, line 72: perhaps this should then be added to the "Raises" section of the docstring...

PR #454, opened 2020-01-13
mesh overhaul

tomvanmele:

### Checklist


tomvanmele:
* mesh attribute getter and setter functions have been merged into one.
* only modified attributes are actually stored
* iterators return a mutable view of a combination of default and modified attributes
* standard mesh algorithms are added to a public n-gon mesh class
* the base mesh implementation and related base functionality have been moved to a core submodule

src/compas/datastructures/mesh/core/basemesh.py
```diff
@@ -42,30 +43,127 @@
...
+
+
+class VertexAttributeView(AttributeView, collections.MutableMapping):
+
+    def __init__(self, defaults, attr):
+        self.defaults = defaults
+        self.attr = attr
+
+    def __getitem__(self, key):
+        if key in self.attr:
```
gonzalocasas: How about changing this part to assume it is there, and move the fallback behavior to an `except`? something like:

```python
try:
    return self.attr[key]
except KeyError:
    return self.defaults[key]
```

That way, you don't incur any extra cost for the common case in which the attribute is defined.

src/compas/datastructures/mesh/core/basemesh.py
```diff
@@ -42,30 +43,127 @@
...
 
+
 __all__ = ['BaseMesh']
 
 
-TPL = """
-================================================================================
-Mesh summary
-================================================================================
+class AttributeView(object):
```
gonzalocasas, line 60: Although internal, these classes (`*AttributeView`) would need a few lines of doc to clarify intent/purpose.

src/compas/datastructures/mesh/core/basemesh.py
```diff
@@ -637,40 +771,49 @@ def from_vertices_and_faces(cls, vertices, faces):
...
         if isinstance(faces, mapping):
             for fkey, vertices in faces.items():
                 mesh.add_face(vertices, fkey)
         else:
             for face in iter(faces):
                 mesh.add_face(face)
-
         return mesh
 
+    def to_vertices_and_faces(self):
```
gonzalocasas, line 797: I think this method is core enough that it would really require a few unit tests on it (if not already present).

src/compas/datastructures/mesh/core/basemesh.py
```diff
@@ -1287,59 +1159,635 @@ def delete_face(self, fkey):
...
+        key : int
+            The vertex identifier.
+        name : str
+            The name of the attribute
+        value : obj, optional
+            The value of the attribute.
+        
+        Returns
+        -------
+        None
```
gonzalocasas: Is this really how to document different return values? I would have assumed only inline comment for this, and this would be assumed to be a 3-tuple return.

gonzalocasas (commented):
Review still in progress but I'll push out a few comments...

gonzalocasas:
LGTM!

The core `basemesh.py` is really too big to review, but in general, I think it's good to go.

The only thing I would like to be included if possible before merge would be a method to unset an attribute. I think this is now not possible because the `vertices_attribute` method relies on value being something in order to act as a setter (as discussed last week).

PR #455, opened 2020-01-13
adding base scene implementation

tomvanmele:

### Checklist


tomvanmele:
a very simple first version of a scene object.
to do:

- add scene graph
- add naming convention for nodes
- add find function for nodes
- ...

src/compas_rhino/scene/scene.py
```diff
@@ -0,0 +1,118 @@
...
+
+__all__ = ['Scene', 'SceneNode']
+
+
+class SceneNode(object):
+
+    def __init__(self, scene, item, **kwargs):
+        self.scene = scene
+        self.item = item
+        self.artist = Artist.build(item, **kwargs)
```
gonzalocasas, line 17: It might be a good thing to move the artist build to be lazily initialized on first usage via a property.

src/compas_rhino/scene/scene.py
```diff
@@ -0,0 +1,118 @@
...
+from __future__ import print_function
+
+import compas_rhino
+from compas_rhino.artists import Artist
+
+
+__all__ = ['Scene', 'SceneNode']
+
+
+class SceneNode(object):
```
gonzalocasas, line 12: `SceneNode` is completely generic, it should not be on `compas_rhino`, it should be part of core.

src/compas_rhino/scene/scene.py
```diff
@@ -0,0 +1,118 @@
...
+
+    def __init__(self, scene, item, **kwargs):
+        self.scene = scene
+        self.item = item
+        self.artist = Artist.build(item, **kwargs)
+        self.nodes = []
+
+
+class Scene(object):
```
gonzalocasas, line 21: Same here, `Scene` should be a core object, not a Rhino-only. There might be a rhino-specific implementation (or sub-class), but the concept of scene should be present in core.

src/compas_rhino/scene/scene.py
```diff
@@ -0,0 +1,118 @@
...
+        self.nodes.append(node)
+        return node
+
+    def update(self):
+        compas_rhino.rs.EnableRedraw(False)
+        for node in self.nodes:
+            node.artist.draw()
+        compas_rhino.rs.EnableRedraw(True)
+
+    def save(self, path, width=1920, height=1080, scale=1,
```
gonzalocasas: I am not sure `save` a scene is logically the same as screenshoting the current view. I would expect saving a scene means saving all meshes and scene configuration into some file format (like the glTF things that is in progress)

gonzalocasas (changes requested):
Moving scene/scenenode concepts to core is pretty fundamental to me, the rest is nitpicking.

tomvanmele:
it is just a temp location, but we can already move it to `compas`. into `compas.scene`?

tomvanmele:
or `compas.scenes`

src/compas_rhino/scene/scene.py
```diff
@@ -0,0 +1,118 @@
...
+__all__ = ['Scene', 'SceneNode']
+
+
+class SceneNode(object):
+
+    def __init__(self, scene, item, **kwargs):
+        self.scene = scene
+        self.item = item
+        self.artist = Artist.build(item, **kwargs)
+        self.nodes = []
```
Licini, line 18: Shouldn't we have a `parentNode` property?

PR #456, opened 2020-01-16
API-docs: Autosummary for modules, numpy docstring formatting

tetov:


tetov:
I'm not getting a listing of the functions in `compas_ghpython.helpers`, I don't understand why. Would love some input on `compas_ghpython.helpers.__init__`.

tomvanmele:
you have to add a corresponding entry in `docs/api`

tetov:
[I might be dense but wouldn't this suffice?](https://github.com/compas-dev/compas/blob/master/docs/api/compas_ghpython.helpers.rst)

tomvanmele:
ha sorry i confused helpers and utilities

src/compas_ghpython/helpers/__init__.py
```diff
@@ -5,6 +5,13 @@
 
 .. currentmodule:: compas_ghpython.helpers
 
+.. autosummary::
+    :toctree: generated/
+    mesh_draw
```
tomvanmele, line 11: not sure but you might have to add a blank line here

tetov: That worked! Weird that compas_ghpython.geometry worked without the line break, I added it to both just in case.

tomvanmele:
so merge?

tetov:
Yup!

PR #481, opened 2020-02-17
Fixed draw_lines() function to use new Point attribute x y z

yck011522:

- [x ] Bug fix in a **backwards-compatible** manner.

### Checklist

### DIFF
--- a/src/compas_ghpython/utilities/drawing.py
+++ b/src/compas_ghpython/utilities/drawing.py
@@ -81,9 +81,9 @@ def draw_lines(lines):
     """
     rg_lines = []
     for l in iter(lines):
-        sp = l['start']
-        ep = l['end']
-        rg_lines.append(Line(Point3d(*sp), Point3d(*ep)))
+        sp = l.start
+        ep = l.end
+        rg_lines.append(Line(Point3d(sp.x, sp.y, sp.z), Point3d(ep.x, ep.y, ep.z)))
     return rg_lines
 
 


tomvanmele:
unless i misunderstand or unless this is implemented very differently in `compas_ghpython` compared to `compas_rhino`, i don't think this makes sense. 

the input of the draw function is a list of dicts with a particular structure. the dict representing a line always has the keys "start" and "end". the values of these keys are the XYZ locations of the start and end points of the line, respectively. the XYZ locations should be lists.

the equivalent of what you are suggesting is that we stop passing the start and end points separately and instead introduce the key "line" in the input dict. the value of this key could then be a tuple of two lists of XYZ coordinates or a COMPAS Line object.

the only problem with that is that Rhino apparently does some kind of type checking and fails when encountering COMPAS objects such as Line and Point, even though these implement the necessary protocols to be used in the same way as lists.

if we do what i think you are suggesting, the draw functions only work with COMPAS objects and no longer with native Python objects.

yck011522:
@brgcode 
The PR is due to the fact that the existing code doesn't work on a `Line` object. 
This is the fix I am able to put together to make it work again.

The input for the function as the name implies should be a list of `Line` data type. where now Line is a proper `Line` object instead of a tuple of two tuples...  And the current `Line` object do not have a 'start' and 'end' attribute.

yck011522:
The following code should work in my opinion (In ghpython context):

```
from compas_ghpython.utilities import draw_lines

l = Line(Point(0,1,2),Point(5,2,4))
output = draw_lines([l])
```
Yet it will produce an error before my PR.

I am open to other implementation that will make this work. But this is what I'm able to put together.

tomvanmele:
understood, but what you suggest would only work with COMPAS objects and not in the general case using native Python objects. currently the following procedures are supported.

**the line is a tuple of two lists of XYZ coordinates**
```python
line = ([0.0, 0.0, 0.0], [1.0, 0.0, 0.0])
lines = [{'start': line[0], 'end': line[1]}]
draw_lines(lines)
```

**the line is a COMPAS line object**
```python
line = Line(Point(0, 0, 0), Point(1, 0, 0))
lines = [{'start': list(line[0]), 'end': list(line[1])}]
# or
# lines = [{'start': list(line.start), 'end': list(line.end)}]
draw_lines(lines)
```
the list conversion is due to the fact that Rhino complains when the input points are COMPAS objects even though these objects support iteration and thus behave exactly like a list. we could add this list conversion in the background to make the two code snippets exactly the same for both cases.

the reason for the dict is that many other "keys" are supported to customise how objects are displayed. for example in the case of lines: "color", "name", "layer", "arrow".
```python
lines = [{'start': line[0], 'end': line[1], 'color': ..., 'name': ..., 'layer': ..., 'arrow': ...}]
draw_lines(lines)
```

we could update this to the following, which could perhaps be more intuitive.
```python
lines = [{'line': line, 'color': ..., 'name': ..., 'layer': ..., 'arrow': ...}]
draw_lines(lines)
```

having said that, in most cases, all of this is handled in the background by artists. the draw functions are lower level and their use is somewhat deprecated if an appropriate artist is available. furthermore, it is very likely that in the near future all drawing will be handled implicitly by scene objects.

as a disclaimer, i am not 100% up to date about the state of `compas_ghpython` so all my comments are based on `compas_rhino`. i have added @romanarust as reviewer for some more targeted feedback.

anyway, not sure if any of this helps you with your problem, but this is how things are currently implemented and the change you suggest would break compatibility with the rest of the code base.

yck011522:
Hi @brgcode ,

I'm not sure if I understand everything you say correctly. But you are suggesting that I shouldn't use the `draw_*` funcions anymore but perhaps the `*Artist` in `compas_ghpython`.

It is just that there are not much Artists in the `compas_ghpython` package at the moment. Should I close this PR and open a new issue about the missing Artist instead?

tomvanmele:
ideally we use artists for everything but i guess that is not always possible at the moment. in the meantime, the way to draw a line using the draw functions is to use the dict approach. in your case this would be something like
```python
line = Line(Point(0, 0, 0), Point(1, 0, 0))
draw_lines([{'start': list(line[0]), 'end': list(line[1])}])
```
if this is not what you want, you could indeed make a PR for more artists...

tomvanmele:
@yck011522 i will close this, but we can revisit this at some other point.

PR #487, opened 2020-03-04
add CONDA_EXE to bootstraper

Licini:
add path of conda to compas_bootstraper to enable conda operations in Rhino

### Checklist

### DIFF
--- a/src/compas_rhino/install.py
+++ b/src/compas_rhino/install.py
@@ -113,7 +113,7 @@ def install(version=None, packages=None):
         # Take either the CONDA environment directory or the current Python executable's directory
         python_directory = os.environ.get('CONDA_PREFIX', None) or os.path.dirname(sys.executable)
         environment_name = os.environ.get('CONDA_DEFAULT_ENV', '')
-        conda_exe = os.environ.get('CONDA_EXE', None)
+        conda_exe = os.environ.get('CONDA_EXE', '')
 
         compas_bootstrapper = os.path.join(ipylib_path, 'compas_bootstrapper.py')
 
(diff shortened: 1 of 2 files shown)


src/compas_rhino/install.py
```diff
@@ -123,6 +125,8 @@ def install(version=None, packages=None):
             with open(compas_bootstrapper, 'w') as f:
                 f.write('ENVIRONMENT_NAME = r"{}"\n'.format(environment_name))
                 f.write('PYTHON_DIRECTORY = r"{}"\n'.format(python_directory))
+                if conda_exe:
+                    f.write('CONDA_EXE = r"{}"\n'.format(conda_exe))
```
tomvanmele: why is this a conditional statement? if the exe is None, should that not also be reflected in the bootstrapper?

Licini: mmm, just couldn't think of a case where bootstraper needs to declare 'conda not exist' rather than just skip providing the variable. 
otherwise we can change it to 
```python
if conda_exe:
    f.write('CONDA_EXE = r"{}"\n'.format(conda_exe))
else:
    f.write('CONDA_EXE = None\n'.)
```
But you see there still needs to be if condition, cuz it has to be assigned as real None value, not in quotes

Licini: wait, I think assign empty string will do it

src/compas_rhino/install.py
```diff
@@ -113,6 +113,8 @@ def install(version=None, packages=None):
         # Take either the CONDA environment directory or the current Python executable's directory
         python_directory = os.environ.get('CONDA_PREFIX', None) or os.path.dirname(sys.executable)
         environment_name = os.environ.get('CONDA_DEFAULT_ENV', '')
+        conda_exe = os.environ.get('CONDA_EXE', '')
```
Licini, line 116: @tomvanmele how about now

gonzalocasas:
👍

PR #492, opened 2020-03-11
Include MMA code from Krister Svanberg

ricardoavelino:

- [  ] Bug fix in a **backwards-compatible** manner.
- [  ] Breaking change: bug fix or new feature that involve incompatible API changes.

### Checklist


tomvanmele:
looks good.
the only thing missing that i can see is the conditional import through the `__init__` file of `compas.numerical`

you can check the other solvers to see how that works...

tomvanmele:
your mma init file also needs updating.... check the dr > dr_numpy example...

ricardoavelino:
I changed the __init__ in compas.numerical, there is no need to change the one in the MMA folder since it already contains the conditional importing . Will do a new pull request with these changes

PR #494, opened 2020-03-11
Include MMA code from Krister Svanberg

ricardoavelino:

### Checklist

1. [X] Add the change to the `CHANGELOG.md` file in the `Unreleased` section under the most fitting heading: `Added`, `Changed`, `Removed`.
1. [X] Run all tests on your computer (i.e. `invoke test`).


src/compas/numerical/mma/mma_numpy.py
```diff
@@ -2,17 +2,849 @@
...
+from scipy.linalg import solve
+import numpy as np
+import logging
+import os
+import time
+
+
+"""
+Orginal work Written by Krister Svanberg in Matlab. This is the python version of the code written
+by Arjen Deetman.
```
gonzalocasas, line 15: Do you refer to this code? https://github.com/arjendeetman/GCMMA-MMA-Python
If you used that as a reference, we cannot include it because it's under a GPL license, that is not compatible with our MIT license.

gonzalocasas (changes requested):

src/compas/numerical/mma/mma_numpy.py
ricardoavelino: Hi Gonzalo.
Thanks for the feedback. I will see what I can do and reopen this pull request in the future.
Thanks.

PR #496, opened 2020-03-16
GLTF export

beverlylytle:
Addresses issue #477 .

### Checklist


tests/compas/files/test_gltf.py
```diff
@@ -1,9 +1,10 @@
+import json
 import os
 
 import pytest
 
 import compas
-from compas.files import GLTF
+from compas.files import GLTF, GLTFContent
```
beverlylytle:
```suggestion
from compas.files import GLTF
from compas.filers import GLTFContent
```

src/compas/files/gltf/data_classes.py
```diff
@@ -0,0 +1,589 @@
...
+        self.name = name
+        self.extras = extras
+        self.extensions = extensions
+
+    def to_dict(self, node_index_by_key, accessor_index):
+        skin_dict = {'joints': [
+            node_index_by_key.get(item)
+            for item in self.joints
+            if node_index_by_key.get(item)
+        ]}
```
gonzalocasas: I always find these multiline list comprehensions hard-ish to read. Perhaps could be split? And perhaps use `filter` instead of the duplicated `node_index_by_key.get` (although, `.get` is constant access, so, depending on the size of these lists, maybe your existing solution is better.

```suggestion
        node_indices = [node_index_by_key.get(item) for item in self.joints]
        skin_dict = {'joints': list(filter(None, node_indices))}
```

src/compas/files/gltf/data_classes.py
```diff
@@ -0,0 +1,589 @@
+from __future__ import print_function
```
gonzalocasas, line 1: As a general comment to all these data classes, wouldn't it make sense to rename all the `to_dict/from_dict` to `to_data/from_data` since they seem to be the same as other `to/from data` thingies in compas?

src/compas/files/gltf/gltf_children.py
```diff
@@ -0,0 +1,64 @@
...
+    def __iter__(self):
+        return iter(self._value)
+
+    def __len__(self):
+        return len(self._value)
+
+    def __bool__(self):
+        return bool(self._value)
+
+    def check(self, v):
```
gonzalocasas: maybe `check_context_nodes` or `check_nodes` would be more descriptive?

src/compas/files/gltf/gltf_children.py
```diff
@@ -0,0 +1,64 @@
+from __future__ import print_function
+from __future__ import division
+from __future__ import absolute_import
+
+
+class GLTFChildren(object):
+    def __init__(self, context, value):
+        self._value = list(value)
```
gonzalocasas: Considering the `_value` is a list, I would rather call it `_values` or `_children` or some other plural. ;)

src/compas/files/gltf/gltf_mesh.py
```diff
@@ -0,0 +1,185 @@
...
+        if len(faces[0]) > 3:
+            raise Exception('Invalid mesh. Expected mesh composed of points, lines xor triangles.')
+        for face in faces:
+            if len(face) != len(faces[0]):
+                # This restriction could be removed by splitting into multiple primitives.
+                raise NotImplementedError('Invalid mesh. Expected mesh composed of points, lines xor triangles.')
+
+    @classmethod
+    def validate_vertices(cls, vertices):
+        if len(vertices) > 4294967295:
```
gonzalocasas, line 142: Where is this limitation coming from? Is this a uint32 overflow? Does it apply to Python 2 only?

src/compas/files/gltf/gltf_mesh.py
```diff
@@ -0,0 +1,185 @@
...
+        self._key = key
+
+    @property
+    def key(self):
+        return self._key
+
+    @property
+    def vertices(self):
+        if not self.weights:
+            return list(itertools.chain(*[primitive.attributes['POSITION'] for primitive in self.primitive_data_list]))
```
gonzalocasas: clever! 😎

src/compas/files/gltf/gltf.py
```diff
@@ -0,0 +1,125 @@
...
+    source_glb = 'https://raw.githubusercontent.com/KhronosGroup/glTF-Sample-Models/master/2.0/BoxInterleaved/glTF-Binary/BoxInterleaved.glb'
+    filepath_glb = os.path.join(compas.APPDATA, 'data', 'gltfs', 'khronos', 'BoxInterleaved.glb')
+
+    download_file_from_remote(source_glb, filepath_glb, overwrite=False)
+
+    gltf = GLTF(filepath_glb)
+    gltf.read()
+
+    default_scene_key = gltf.content.default_scene_key or 0
+    default_scene = gltf.content.scenes[default_scene_key]
```
gonzalocasas: Would it make sense to add a `gltf.content.default_or_first_scene` getter method or property?

src/compas/files/gltf/__init__.py
```diff
@@ -0,0 +1,19 @@
+from __future__ import absolute_import
```
gonzalocasas, line 1: It looks like glTF is still not included in the docs yet.

tests/compas/files/test_gltf.py
```diff
@@ -40,22 +42,105 @@ def sparse_gltf():
...
+def animated_gltf():
+    return os.path.join(BASE_FOLDER, 'fixtures', 'AnimatedMorphCube.glb')
+
+
+@pytest.fixture
+def textured_gltf():
+    return os.path.join(BASE_FOLDER, 'fixtures', 'BoxTextured.glb')
+
+
+def test_from_gltf_edges_loaded(
```
gonzalocasas: Q: maybe the tests could be split, right? Every time you start a new glTF read, it feels like that could be a new test...

src/compas/files/gltf/gltf_mesh.py
gonzalocasas: Question: would it make sense to move this line to `helpers` to something like `get_unweighted_mesh_vertices` so that it's near the alternative code path of `get_weighted_mesh_vertices`?

src/compas/files/gltf/helpers.py
```diff
@@ -0,0 +1,39 @@
...
+def get_matrix_from_col_major_list(matrix_as_list):
+    return [[matrix_as_list[i + j * 4] for j in range(4)] for i in range(4)]
+
+
+def matrix_to_col_major_order(matrix):
+    return [matrix[i][j] for j in range(4) for i in range(4)]
+
+
+def get_morph_function(weights):
+    # Returns a function which computes for a fixed list w of scalar weights the linear combination
```
gonzalocasas: I would turn this into normal docstring instead of inlined comment.

src/compas/files/gltf/helpers.py
```diff
@@ -0,0 +1,39 @@
...
+    def weighted_sum(vertex_coordinate, *targets_coordinate):
+        return vertex_coordinate + math.fsum(map(apply_weight, weights, targets_coordinate))
+
+    def apply_morph_target(vertex, *targets):
+        return tuple(map(weighted_sum, vertex, *targets)) + ((vertex[-1],) if len(vertex) == 4 else ())
+
+    return apply_morph_target
+
+
+def get_weighted_mesh_vertices(mesh, weights):
```
gonzalocasas, line 36: Despite being implementation details, I would add docstrings to this as well, because it's not entirely clear what a weighted mesh vertex is on a quick glance.

src/compas/files/gltf/gltf_exporter.py
```diff
@@ -0,0 +1,396 @@
...
+                    type_ = TYPE_VEC3
+                    target_dict[attr] = self._get_accessor(target[attr], component_type, type_, True)
+                targets.append(target_dict)
+
+            primitive_dict = primitive_data.to_dict(indices_accessor, attributes, targets, self._material_index_by_key)
+
+            primitives.append(primitive_dict)
+        return primitives
+
+    def _get_accessor(self, data, component_type, type_, include_bounds=False):
```
gonzalocasas: These `_get_*` functions confuse me a bit. Not the content, but the name, because they do more than just getting stuff, they have side effects on the gltf dictionary. Would a prefix like `_allocate_` be more fitting than `_get_` perhaps?

src/compas/files/gltf/gltf_exporter.py
```diff
@@ -0,0 +1,396 @@
...
+        buffer_view_index = self._get_buffer_view(bytes_)
+        accessor_dict = {
+            'bufferView': buffer_view_index,
+            'count': count,
+            'componentType': component_type,
+            'type': type_,
+        }
+        if include_bounds:
+            try:
+                _ = [e for e in data[0]]
```
gonzalocasas, line 362: I think this line deserves a comment explaining why this is like this.

gonzalocasas (approved):
**Excellent stuff!!** 🚀 :shipit: 
It's quite some amount of code, but it's really cleanly structured, great job!

I added a few comments, but I see no big issues at all, mostly just questions and/or very small requests.

image

src/compas/files/gltf/data_classes.py
beverlylytle: That's the mathematician in me, finding such dictionary/list comprehensions completely natural.  Thanks for making me look at this code again because I see that there is a bug.  Since `0` is falsey, I have accidentally filtered out the zeroth node from all skins.  Since I would need a custom filter function, I think I would be most pythonic to do an amalgam of our two versions.

src/compas/files/gltf/data_classes.py
beverlylytle: yep

src/compas/files/gltf/gltf_mesh.py
beverlylytle: https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#primitiveindices
so I suppose there actually should be a few more checks along these lines...

src/compas/files/gltf/gltf_mesh.py
```diff
@@ -0,0 +1,185 @@
...
+        return faces
+
+    def shift_indices(self, indices, shift):
+        return [index + shift for index in indices]
+
+    def group_indices(self, indices, group_size):
+        it = [iter(indices)] * group_size
+        return list(zip(*it))
+
+    @classmethod
```
tomvanmele, line 107: just out of curiosity, why is this not `@staticmethod`?

beverlylytle: Because I have used it in the `classmethod` `from_vertices_and_faces`.

tomvanmele: so it is a helper method for one of the other `classmethod`s?

beverlylytle: at the moment, yes, but i am rethinking that.  suggestions?

src/compas/files/gltf/gltf_exporter.py
beverlylytle: I went with `_construct_`.

src/compas/files/gltf/gltf_mesh.py
tomvanmele: is it important that they are part of the class definition? they don't seem to be using any of the attributes or methods related to the class. so perhaps they can just be helper functions?

tomvanmele: in any case, i think you can also just mark them as static which would already get rid of this unused and implicit `cls` argument...

beverlylytle: Yes, I will make it a helper method.

src/compas/files/gltf/__init__.py
beverlylytle: will fix...

PR #498, opened 2020-03-19
Changed an odd implementation + fixed some typos and docstring issues

rjodon:
HI, 

I spent a day installing compas/reading the doc and tutorials. 
I have made 3 commits: 

- One is about a few typos in the user documentation.
- The second is about type declaration in docstring.
- Finally, I rewrote something because I found that it was kinda odd. 

Cheers, 
Robin

PS: these are actually my first commits on the project :)

### DIFF
--- a/src/compas_plotters/meshplotter.py
+++ b/src/compas_plotters/meshplotter.py
@@ -143,19 +143,19 @@ class MeshPlotter(Plotter):
         ----------
         keys : list
             The keys of the vertices to plot.
-        radius : {list, dict}
+        radius : {float, dict}
             A list of radii for the vertices.
-        text : {list, dict}
+        text : {{'index', 'key'}, str, dict}
             Strings to be displayed on the vertices.
-        facecolor : {list, dict}
+        facecolor : {color, dict}
             Color for the vertex circle fill.
-        edgecolor : {list, dict}
+        edgecolor : {color, dict}
             Color for the vertex circle edge.
-        edgewidth : {list, dict}
+        edgewidth : {float, dict}
             Width for the vertex circle edge.
-        textcolor : {list, dict}
+        textcolor : {color, dict}
             Color for the text to be displayed on the vertices.
-        fontsize : {list, dict}
+        fontsize : {int, dict}
             Font size for the text to be displayed on the vertices.
 
         Returns
@@ -305,11 +305,11 @@ class MeshPlotter(Plotter):
         ----------
         keys : list
             The keys of the edges to plot.
-        width : {list, dict}
+        width : {float, dict}
             Width of the mesh edges.
-        color : {list, dict}
+        color : {color, dict}
             Color for the edge lines.
-        text : {list, dict}
+        text : {{'index', 'key'}, str, dict}
             Strings to be displayed on the edges.
         textcolor : rgb tuple or dict of rgb tuples
             Color for the text to be displayed on the edges.
@@ -392,17 +392,17 @@ class MeshPlotter(Plotter):
         ----------
         keys : list
             The keys of the edges to plot.
-        text : {list, dict}
+        text : {{'index', 'key'}, str, dict}
             Strings to be displayed on the edges.
-        facecolor : {list, dict}
+        facecolor : {color, dict}
             Color for the face fill.
-        edgecolor : {list, dict}
+        edgecolor : {color, dict}
             Color for the face edge.
-        edgewidth : {list, dict}
+        edgewidth : {float, dict}
             Width for the face edge.
-        textcolor : {list, dict}
+        textcolor : {color, dict}
             Color for the text to be displayed on the edges.
-        fontsize : {list, dict}
+        fontsize : {int, dict}
             Font size for the text to be displayed on the edges.
 
         Returns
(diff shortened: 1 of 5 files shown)


src/compas_plotters/meshplotter.py
```diff
@@ -20,13 +20,26 @@
...
+     Parameters
+     ----------
+     keys: list
+         The list of keys
+     value: {dict, None}
+         A value or the already formed dictionary
+     default:
+         A default value to set if no value
+
+
```
tomvanmele: perhaps remove one of the blanks

src/compas_plotters/meshplotter.py
```diff
@@ -153,19 +166,19 @@ def draw_vertices(self,
...
             Color for the vertex circle edge.
-        edgewidth : list
+        edgewidth : {list, dict}
             Width for the vertex circle edge.
-        textcolor : list
+        textcolor : {list, dict}
             Color for the text to be displayed on the vertices.
-        fontsize : list
+        fontsize : {list, dict}
             Font size for the text to be displayed on the vertices.
```
tomvanmele: the docs are indeed wrong. however, the values can actually also be single values, and in the case of `text` you can even use the keywords "key" or "index". so perhaps it should be
```python
radius : {float, list, dict}
text : {{'index', 'key'}, str, list, dict}
facecolor : {color, list, dict}
edgecolor : {color, list, dict}
edgewidth : {float, list, dict}
textcolor : {color, list, dict}
fontsize : {int, list, color}
```

src/compas_plotters/meshplotter.py
```diff
@@ -20,13 +20,26 @@
...
+         A default value to set if no value
+
+
+     Returns
+     -------
+     dict
+         A dictionary
+     """
+    value = value or dict.fromkeys(keys, default)
+    return {key: value.get(key, default) for key in keys}
```
tomvanmele: considering the fact that the `value` can also be a single value (see below), i am not sure this will produce the same outcome...

src/compas_plotters/meshplotter.py
```diff
@@ -315,11 +328,11 @@ def draw_edges(self,
...
+        width : {list, dict}
             Width of the mesh edges.
-        color : list
+        color : {list, dict}
             Color for the edge lines.
-        text : dict
+        text : {list, dict}
             Strings to be displayed on the edges.
         textcolor : rgb tuple or dict of rgb tuples
             Color for the text to be displayed on the edges.
```
tomvanmele: same as a above

src/compas_plotters/meshplotter.py
```diff
@@ -402,17 +415,17 @@ def draw_faces(self,
...
             Color for the face edge.
-        edgewidth : list
+        edgewidth : {list, dict}
             Width for the face edge.
-        textcolor : list
+        textcolor : {list, dict}
             Color for the text to be displayed on the edges.
-        fontsize : list
+        fontsize : {list, dict}
             Font size for the text to be displayed on the edges.
```
tomvanmele: same

src/compas_plotters/networkplotter.py
```diff
@@ -156,19 +156,19 @@ def draw_nodes(self,
...
+        edgecolor : {list, dict}
             Color for the node circle edge.
-        edgewidth : list
+        edgewidth : {list, dict}
             Width for the node circle edge.
-        textcolor : list
+        textcolor : {list, dict}
             Color for the text to be displayed on the nodes.
-        fontsize : list
+        fontsize : {list, dict}
```
tomvanmele: same

src/compas_plotters/networkplotter.py
```diff
@@ -242,15 +242,15 @@ def draw_edges(self,
...
             Color for the edge lines.
-        text : list
+        text : {list, dict}
             Strings to be displayed on the edges.
-        textcolor : list
+        textcolor : {list, dict}
             Color for the text to be displayed on the edges.
-        fontsize : list
+        fontsize : {list, dict}
             Font size for the text to be displayed on the edges.
```
tomvanmele: same

src/compas_plotters/meshplotter.py
rjodon: Ok, I was not sure of that. 
It was difficult to me to see all the possible cases.

src/compas_plotters/meshplotter.py
rjodon: OK

src/compas_plotters/meshplotter.py
tomvanmele: yes i know. the plotters are due for a thorough update :)

src/compas_plotters/meshplotter.py
rjodon: Ok, the first test made me thing that value was either a dict or None. 
So, basically, if you use a single value, then you set this single value to the whole dict?

src/compas_plotters/meshplotter.py
rjodon: OK

src/compas_plotters/meshplotter.py
tomvanmele: yes that sounds about right

src/compas_plotters/meshplotter.py
tomvanmele: actually my comment was not entirely correct. it should be
```python
radius : {float, dict}
text : {{'index', 'key'}, str, dict}
facecolor : {color, dict}
edgecolor : {color, dict}
edgewidth : {float, dict}
textcolor : {color, dict}
fontsize : {int, dict}
```

PR #499, opened 2020-03-23
Fix unsafe loop + fix deprecation warnings for import collections ABCs

rjodon:

### Checklist

### DIFF
--- a/src/compas/datastructures/mesh/core/operations/substitute.py
+++ b/src/compas/datastructures/mesh/core/operations/substitute.py
@@ -2,7 +2,6 @@ from __future__ import print_function
 from __future__ import absolute_import
 from __future__ import division
 
-
 __all__ = [
     'mesh_substitute_vertex_in_faces'
 ]
@@ -48,7 +47,7 @@ def mesh_substitute_vertex_in_faces(mesh, old_vkey, new_vkey, fkeys=None):
 
     # apply to all faces if there is none chosen
     if fkeys is None:
-        fkeys = mesh.faces()
+        fkeys = list(mesh.faces())
 
     # substitute vertices
     for fkey in fkeys:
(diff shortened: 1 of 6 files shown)


src/compas/datastructures/mesh/core/operations/substitute.py
```diff
@@ -48,7 +48,8 @@ def mesh_substitute_vertex_in_faces(mesh, old_vkey, new_vkey, fkeys=None):
 
     # apply to all faces if there is none chosen
     if fkeys is None:
-        fkeys = mesh.faces()
+        mesh_cp = copy.deepcopy(mesh)  # ifno deepcopy, next loop is unsafe as we modify face dict while looping on it
+        fkeys = mesh_cp.faces()
```
tomvanmele: why not just
```python
fkeys = list(mesh.faces())
```

rjodon: Just a bad habit from myself. Of course your suggestion is much more efficient.

src/compas/datastructures/mesh/core/operations/substitute.py
```diff
@@ -48,7 +47,7 @@ def mesh_substitute_vertex_in_faces(mesh, old_vkey, new_vkey, fkeys=None):
 
     # apply to all faces if there is none chosen
     if fkeys is None:
-        fkeys = mesh.faces()
+        fkeys = list(mesh_cp.faces())
```
tomvanmele: should be `mesh` not `mesh_cp`

rjodon: It should be done now. I will add a git hook to avoid fixing a commit in a hury and not running the tests.

PR #501, opened 2020-03-27
Update RhinoMesh

romanarust:
Update RhinoMesh such that it can be created from Grasshopper as well
`RhinoMesh.from_geometry(mesh)` and `compas_mesh = RhinoMesh.from_geometry(mesh).to_compas()`

### DIFF
--- a/src/compas_rhino/geometry/mesh.py
+++ b/src/compas_rhino/geometry/mesh.py
@@ -45,13 +45,19 @@ class RhinoMesh(RhinoGeometry):
         guid = compas_rhino.select_mesh()
         return cls.from_guid(guid)
 
+    @classmethod
+    def from_geometry(cls, geometry):
+        mesh = cls()
+        mesh.geometry = geometry
+        return mesh
+
     @property
     def vertices(self):
-        return [map(float, vertex) for vertex in compas_rhino.rs.MeshVertices(self.guid)]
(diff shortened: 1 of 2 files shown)


src/compas_rhino/geometry/mesh.py
```diff
@@ -45,13 +45,19 @@ def from_selection(cls):
...
+    @classmethod
+    def from_geometry(cls, geometry):
+        mesh = cls()
+        mesh.geometry = geometry
+        return mesh
+
     @property
     def vertices(self):
-        return [map(float, vertex) for vertex in compas_rhino.rs.MeshVertices(self.guid)]
+        return [map(float, vertex) for vertex in compas_rhino.rs.MeshVertices(self.geometry)]
```
tomvanmele, line 56: does this work in Rhino?

romanarust: yes, it is handled by rs.coercemesh: https://github.com/mcneel/rhinoscriptsyntax/blob/2fd9823efacf2048b740faf6a747401c17728967/Scripts/rhinoscript/utility.py#L1029

tomvanmele: i guess it does because of the coercing mechanism in the background?

romanarust: ah, yes see my updated comment

PR #507, opened 2020-04-14
List class attributes, document joint types.

tetov:

Display class attributes from [`autoattribute`](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#directive-autoattribute) in Sphinx class documentation.

This came up in discussion related to `compas.robots.Joint.type` here: https://github.com/compas-dev/compas_fab/pull/119#discussion_r389408529

I still can't link to these though (with `:data:`/`:any:`/`:attr:`).

I'm hosting the docs built from this branch. Have a look at the API reference of `compas.robots.Joint` [here](https://p.tetov.se/pr_class_attr/api/generated/compas.robots.Joint.html#compas.robots.Joint).

### Checklist

### DIFF
--- a/docs/_templates/autosummary/class.rst
+++ b/docs/_templates/autosummary/class.rst
@@ -19,3 +19,14 @@
     {%- endfor %}
     {% endif %}
     {% endblock %}
+
+    {% block attributes %}
+    {% if attributes %}
+    .. rubric:: Attributes
+
+    .. autosummary::
+    {% for item in attributes %}
+        ~{{ name }}.{{ item }}
+    {%- endfor %}
+    {% endif %}
+    {% endblock %}
--- a/docs/_templates/autosummary/class.rst
+++ b/docs/_templates/autosummary/class.rst
@@ -19,14 +19,3 @@
     {%- endfor %}
     {% endif %}
     {% endblock %}
-
-    {% block attributes %}
-    {% if attributes %}
-    .. rubric:: Attributes
-
-    .. autosummary::
-    {% for item in attributes %}
-        ~{{ name }}.{{ item }}
-    {%- endfor %}
-    {% endif %}
-    {% endblock %}
--- a/src/compas/robots/model/joint.py
+++ b/src/compas/robots/model/joint.py
@@ -192,8 +192,7 @@ class Axis(object):
 
 
 class Joint(object):
-    """Representation of the kinematics and dynamics of a joint and its safety
-    limits.
+    """Representation of the kinematics and dynamics of a joint and its safety limits.
 
     Attributes
     ----------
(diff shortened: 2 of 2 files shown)


tomvanmele:
i had removed this because it makes a bit of a mess of properties. since there are usually not too many attributes to document, i had imagined this to be more convenient by doing it more explicitly in a section in the class docstring. would that also be an option?

tetov:
I agree. They also show up in child classes (`compas_fab.robots.Constraint` types shows up in `compas_fab.robots.JointConstraint` including `POSITION` & `ORIENTATION` which makes it a bit confusing.)

What do you think @gonzalocasas?

tomvanmele:
the inheritance problem can be solved by tweaking the rst template (the old version was like this. it is still in the repo under `class.rst.copy`) but all the other problems stay the same...

tetov:
> 
> 
> the inheritance problem can be solved by tweaking the rst template (the old version was like this. it is still in the repo under `class.rst.copy`) but all the other problems stay the same...

Good to know.

Maybe a section of the class docstring titled "Class attributes"?

tomvanmele:
just to make sure we are talking about the same things... by class attributes you mean actual "class-level" attributes?

gonzalocasas:
I like the idea of auto document as much as possible, but adding this now without re-checking existing docs means a lot will be duplicated (see `Frame` for instance). Also, IF we do this, attributes should go before the methods section.

tetov:
> just to make sure we are talking about the same things... by class attributes you mean actual "class-level" attributes?

Yes, `Joint.REVOLUTE` not 
```
j = Joint()
j.position
```

> I like the idea of auto document as much as possible, but adding this now without re-checking existing docs means a lot will be duplicated (see `Frame` for instance). Also, IF we do this, attributes should go before the methods section.

They would if the were hand documented in the docstring. Should they be in a separate section from other attributes?

tomvanmele:
agreed with autodoc'ing as much as possible, but the support for properties is not great, which is why abandoned it after trying many different variations. if possible i would stick to the manually included version for now.

afaik, they would just go in the `Attributes` section, but we could register an additional section for `Class Attributes` specifically if you prefer...

tomvanmele:
https://stackoverflow.com/questions/28437869/how-should-i-document-class-and-object-attributes-using-numpys-style

tomvanmele:
i think this would be the cleanest solution, but have you tried if this actually renders as a section in the docstring?

tetov:
> 
> 
> i think this would be the cleanest solution, but have you tried if this actually renders as a section in the docstring?

Hosting it here :) <https://p.tetov.se/pr_class_attr/api/generated/compas.robots.Joint.html#compas.robots.Joint>.

I'll also add a reference to this issue in Sphinx. Seems a bit stale though. https://github.com/sphinx-doc/sphinx/issues/3141

tomvanmele:
okay cool

Issue #510, opened 2020-04-17
Numba 0.49 dropping Python 2 support.


tetov:
[Numba 0.49.0](https://numba.pydata.org/numba-doc/latest/release-notes.html#version-0-49-0-apr-16-2020) dropped Python 2 support. This showed up in a [travis job when pip tried to build llvmlite 0.32 from sdist](https://travis-ci.com/github/tetov/compas_fab/jobs/320507740#L315) (https://github.com/compas-dev/compas_fab/pull/138) . Previous versions of llvmlite provided wheels for py27.

I think pinning Numba to `numba<0.49` should be enough.

tomvanmele:
@gonzalocasas how does stuff like this relate to the transpiling? making our code backwards compatible is one thing, but we can't keep missing out on updates because of it...

tetov:
Or `numba <0.49; python_version >= '2.7' and python_version < '3.0'`.

Functions requiring numba are already marked with `_numba` right? And they didn't work from Rhino before. So if Rhino is the only reason to keep supporting Python 2, dropping numba support for Python 2 doesn't change much.

We could make Python 3 a build/install requirement, but still support IPY from Rhino/Grasshopper.

tomvanmele:
yes, you are right. the Python 3 build will indeed still use the latest versions.

tetov:
Fixed in #518?

gonzalocasas:
I would say so

PR #511, opened 2020-04-17
Pin numba and llvmlite for py27

tetov:

Fixing #510.

### Checklist

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -10,7 +10,7 @@ matplotlib >= 2.2; python_version >= '3.5' and sys_platform != 'win32'
 networkx
 llvmlite >= 0.31, < 0.32; python_version < '3.6'
 numba < 0.48; python_version < '3.6'
-numba
+numba; python_version >= '3.6'
 numpy >= 1.15.4
 pillow
 planarity
--- a/requirements.txt
+++ b/requirements.txt
@@ -8,6 +8,8 @@ matplotlib >= 2.2, < 3.0; python_version >= '2.7' and python_version < '3.0'
 matplotlib >= 2.2, < 3.1; python_version >= '3.5' and sys_platform == 'win32'
 matplotlib >= 2.2; python_version >= '3.5' and sys_platform != 'win32'
 networkx
+llvmlite >= 0.31, < 0.32; python_version < '3.6'
+numba < 0.48; python_version < '3.6'
 numba
 numpy >= 1.15.4
 pillow


requirements.txt
```diff
@@ -8,6 +8,8 @@ matplotlib >= 2.2, < 3.0; python_version >= '2.7' and python_version < '3.0'
 matplotlib >= 2.2, < 3.1; python_version >= '3.5' and sys_platform == 'win32'
 matplotlib >= 2.2; python_version >= '3.5' and sys_platform != 'win32'
 networkx
+llvmlite >= 0.31, < 0.32; python_version < '3.6'
+numba < 0.48; python_version < '3.6'
 numba
```
tomvanmele: shouldn't there be a python version constraint on the second line as well?
afaik, none of these get processed in order...

tetov: They don't? I got lucky then! I'll fix that.

tetov: Hm but that makes the `llvmlite` situation tricky. `numba==0.47.0` has `llvmlite>=0.31` in it's `requirements.txt` which means pip will install `0.32` if it does numba first.

CHANGELOG.md
```diff
@@ -19,6 +19,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 - Updated Blender installation docs for latest release.
+- Changed `requirements.txt` for Python 2 environments. Last version supporting
```
gonzalocasas: This does not go to the changelog, is a non-functional change.

data/temp.obj
```diff
@@ -1,6 +1,6 @@
 # OBJ
 # COMPAS
-# version: 0.15.5-aea3df39
+# version: 0.15.5-92cbf6f8
```
gonzalocasas: ?

requirements.txt
```diff
@@ -8,7 +8,9 @@ matplotlib >= 2.2, < 3.0; python_version >= '2.7' and python_version < '3.0'
 matplotlib >= 2.2, < 3.1; python_version >= '3.5' and sys_platform == 'win32'
 matplotlib >= 2.2; python_version >= '3.5' and sys_platform != 'win32'
 networkx
-numba
+llvmlite >= 0.31, < 0.32; python_version < '3.6'
```
gonzalocasas, line 11: If this is for py27 support, then the constraint should be `<=2.7` instead of 3.6 ( i assume you rely on the fact that we are not building any 3.x lower than 3.6 but that makes the requirement semantically incorrect.

gonzalocasas: Do we really need to put this llvm requirement here? Smells fishy

gonzalocasas (changes requested):

data/temp.obj
tomvanmele: i added some header info (for example, COMPAS version) to OBJ files written by COMPAS. so if he ran the tests that would make sense...

requirements.txt
tomvanmele: except for `llvmlite` they are listed alphabetically. would be weird if that worked out by chance until now :)

requirements.txt
tomvanmele: seems weird that we all of a sudden need it at all indeed...

tetov: I used 3.6 because the referenced version of numba is the last to support anything below 3.6.

Without the llvmlight requirement: pip will install `numba==0.47.0` if on python 2. The `requirements.txt` for `numba==0.47.0` specifies `llvmlite>=0.31` so pip will install `llvmlite==0.32`. `llvmlite==0.32` has no wheel for py27 on pypi and build fails (tested on travis/win and my linux installation).

requirements.txt
tetov: True! I just couldn't believe there's no way to specify install order! (without something silly like `cat requirements.txt | xargs pip install -`)

requirements.txt
tetov: The alternative would be to ask the numba devs to release `numba==0.47.1` with a stricter version specifier for llvmlite.

tomvanmele: none of this makes sense. not son long ago everything worked fine for py27 so there must be a version of `numba` that works without specifying anything else...

tomvanmele: @gonzalocasas remind me again why we need a py27 build. can we not just do the transpile thing and check for py27 syntax?

tetov: `llvmlite==0.32` was released [7 hours ago](https://pypi.org/project/llvmlite/0.32.0/#history) and is shipped without a wheel for py27 because they dropped support for py27. That is what has changed.

Maybe it can still be built from sdist on py27 with some tinkering, but I haven't tried.

data/temp.obj
gonzalocasas: 👍

requirements.txt
gonzalocasas: @brgcode you are right, we don't really need a py27 built at all. Ideally we have the transpiler in place, but considering now syntax should all be py27 compatible anyway, and library dependencies do not affect IronPython stuff, the immediate thing to do would be to add IronPython to the travis build, so that we test it automatically.

I don't know how good the coverage is thou. @tetov you added/were adding/were thinking about adding coverage reports, right?

tetov: @gonzalocasas I've only added coverage reports to compas_fab. I can set it up for compas too. 

Do you mean executing pytest with IronPython without installating the package?

data/temp.obj
tetov: I run my git workflow on auto pilot sometimes..

```bash
ga .
gcmsg ”Commit msg”
gp
```

I can either try to be more careful or I can create a pre-commit hook for geometry files haha.

requirements.txt
tomvanmele: i have the feeling we have not yet really pin-pointed the problem.

in my opinion the goal is not to provide a working py27 build (py27 is history), but to make sure stuff will work in Rhino and GH. we are also not running any tests on `compas_rhino` and `compas_ghpython`.

therefore would it not make sense to include a "fake" py27 build just for the sake of Rhino/GH compatibility, potentially with a minimal requirements file?

@gonzalocasas ping

tetov:
If we are not supporting installation on py27 then this PR is unnecessary, I agree.

‐‐‐‐‐‐‐ Original Message ‐‐‐‐‐‐‐

gonzalocasas:
I suggest we close this in favor of https://github.com/compas-dev/compas/pull/518

PR #518, opened 2020-04-22
Refactor CI builds and add unit testing for IronPython

gonzalocasas:
Since Python 2.7 support is being dropped by all our dependencies, we cannot keep building on it but since we need to keep supporting IronPython 2.7 for the foreseeable future, we must ensure that at least syntax is 2.7 compatible so that it runs in Rhino and friends.

In this PR:
 - Added a reworked travis configuration to have:
   - One build on each of the 3 main OSes (Linux, Mac, Windows) on the latest Python stable release available. (currently, 3.8 on win/linux, and 3.7 on mac)
    - One build on python 3.6 (linux) because it is still probably the most widely used version.
    - One build on IronPython 2.7.8 (windows). This is the same as the version embedded on Rhino 6 for Windows (not on Mac, that uses 2.7.9 and some differences affect the behavior of buffers, in particular, visible on the glTF support)
  - Added unit test runner for IronPython. To this purpose, a tiny `pytest` clone was added, because `pytest` does not support IronPython. It's a minimal tool but it is able to run all tests we have creating an output that resembles that of `pytest`.
  - Fixed some bugs uncovered by running the tests on IronPython:
    - Added precision to the `Mesh.from_stl` to fix mesh test that was collapsing some vertices on ipy
    - Added fallback to the incomplete bytearray support of IronPython 2.7.8 for the glTF exporter
    - Tweaked the `compas.geometry.allclose` function to also compare recursively, just like the `numpy.allclose` so that lots of tests can use that instead of numpy's version, and hence run on IronPython.
  - Skipped two tests that depend on `planarity`

### Checklist

1. [ ] ~~Add the change to the `CHANGELOG.md` file in the `Unreleased` section under the most fitting heading: `Added`, `Changed`, `Removed`.~~

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,7 +5,8 @@ ezdxf
 imageio <= 2.6; python_version < '3.5'
 imageio >= 2.7; python_version >= '3.5'
 matplotlib >= 2.2, < 3.0; python_version >= '2.7' and python_version < '3.0'
-matplotlib >= 2.2, < 3.1; python_version >= '3.5' and sys_platform == 'win32'
+matplotlib >= 2.2, < 3.1; python_version >= '3.5' and python_version <= '3.7' and sys_platform == 'win32'
+matplotlib >= 3.1; python_version >= '3.8' and sys_platform == 'win32'
 matplotlib >= 2.2; python_version >= '3.5' and sys_platform != 'win32'
 networkx
 numba
--- a/src/compas/geometry/_core/basic.py
+++ b/src/compas/geometry/_core/basic.py
@@ -110,7 +110,8 @@ def allclose(l1, l2, tol=1e-05):
     .. [1] https://docs.scipy.org/doc/numpy/reference/generated/numpy.allclose.html
 
     """
-    if any(fabs(a - b) > tol for a, b in zip(l1, l2)):
+
+    if any(not allclose(a, b, tol) if hasattr(a, '__iter__') else fabs(a - b) > tol for a, b in zip(l1, l2)):
         return False
     return True
 
--- a/tests/compas/datastructures/test_mesh.py
+++ b/tests/compas/datastructures/test_mesh.py
@@ -561,12 +561,12 @@ def test_face_curvature():
 # boundary
 # --------------------------------------------------------------------------
 
-def test_vertices_on_boundary():
-    mesh = Mesh.from_obj(compas.get('quadmesh.obj'))
-    assert sorted(mesh.vertices_on_boundary()) == [0, 1, 2, 3, 4, 5, 6, 7, 14, 15, 17, 33, 35, 37, 38, 39, 40, 41, 42, 43, 44, 45, 53, 62, 71, 73, 74, 75, 76, 84, 85, 86, 87, 88, 89, 98]
+# def test_vertices_on_boundary():
+#     mesh = Mesh.from_obj(compas.get('quadmesh.obj'))
+#     assert mesh.vertices_on_boundary() == [0, 1, 2, 3, 4, 5, 6, 7, 14, 15, 17, 33, 35, 37, 38, 39, 40, 41, 42, 43, 44, 45, 53, 62, 71, 73, 74, 75, 76, 84, 85, 86, 87, 88, 89, 98]
 
-    mesh = Mesh.from_obj(compas.get('boxes.obj'))
-    assert mesh.vertices_on_boundary() == []
+#     mesh = Mesh.from_obj(compas.get('boxes.obj'))
+#     assert mesh.vertices_on_boundary() == []
 
 
 def test_vertices_on_boundaries():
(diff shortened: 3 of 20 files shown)


requirements.txt
```diff
@@ -5,7 +5,8 @@ ezdxf
 imageio <= 2.6; python_version < '3.5'
 imageio >= 2.7; python_version >= '3.5'
 matplotlib >= 2.2, < 3.0; python_version >= '2.7' and python_version < '3.0'
-matplotlib >= 2.2, < 3.1; python_version >= '3.5' and sys_platform == 'win32'
+matplotlib >= 2.2, < 3.1; python_version >= '3.5' and python_version <= '3.7' and sys_platform == 'win32'
+matplotlib >= 3.1; python_version >= '3.8' and sys_platform == 'win32'
```
brgcode, line 9: nice :)

requirements.txt
```diff
@@ -5,7 +5,8 @@ ezdxf
...
 matplotlib >= 2.2; python_version >= '3.5' and sys_platform != 'win32'
```
tomvanmele, line 10: :)

src/compas/geometry/_core/basic.py
```diff
@@ -110,7 +110,8 @@ def allclose(l1, l2, tol=1e-05):
     .. [1] https://docs.scipy.org/doc/numpy/reference/generated/numpy.allclose.html
 
     """
-    if any(fabs(a - b) > tol for a, b in zip(l1, l2)):
+
+    if any(not allclose(a, b, tol) if hasattr(a, '__iter__') else fabs(a - b) > tol for a, b in zip(l1, l2)):
```
tomvanmele, line 114: ?

tomvanmele: which extreme boundary case are we accommodating here?

tests/compas/datastructures/test_mesh.py
```diff
@@ -49,12 +49,12 @@ def test_from_ply():
 
 
 def test_from_stl():
-    mesh = Mesh.from_stl(compas.get('cube_ascii.stl'))
+    mesh = Mesh.from_stl(compas.get('cube_ascii.stl'), '6f')
```
tomvanmele, line 52: still a question for me how do reliably deal with precision. this is btw, why i started adding header info to the files that we write... i think precision should be part of that...

tests/compas/geometry/test_basics.py
```diff
@@ -34,7 +33,7 @@
 ]
 )
 def test_angle_vectors(u, v, angle):
-    assert angle_vectors(u, v) == pytest.approx(angle)
+    assert close(angle_vectors(u, v), angle)
```
tomvanmele, line 36: so you prefer that we rely on our own interpretation of "close enough"?

tests/compas/geometry/test_bbox.py
```diff
@@ -50,5 +54,11 @@ def test_oriented_bounding_box_xy_numpy(coords, expected):
...
-    print(oriented_bounding_box_numpy(coords).tolist())
-    assert expected == oriented_bounding_box_numpy(coords).tolist()
+    if compas.IPY:
+        return
+
+    from compas.geometry import oriented_bounding_box_numpy
+
+    results = oriented_bounding_box_numpy(coords).tolist()
+    for result, expected_values in zip(results, expected):
+        assert allclose(result, expected_values, tol=1e-3)
```
tomvanmele, line 64: we should have a general strategy for precision...

tomvanmele (commented):
precision seems to be a thing

src/compas/geometry/_core/basic.py
gonzalocasas: is not a boundary case at all, this recursively compares if the element is an iterable, ie `list of list of floats` can be compared. This is how `np.allclose` behaves. The reasons for replicating the behavior is not just consistency, but actually the need to remove all `numpy` from a lot of unit tests that were using `np.allclose` to check results (but were not using `numpy` at all other than that).

tests/compas/datastructures/test_mesh.py
gonzalocasas: Headers sounds like a sound idea ;)  Additionally, the serialization formats used by compas should have a schema somewhere, and support versioning of themselves; but all that is a discussion for another PR.

tests/compas/geometry/test_basics.py
gonzalocasas: Not really, again and just like the change of `np.allclose` to `compas.geometry.allclose`, the idea is was only to make the tests run. In this case, using `pytest.approx` was not really adding a lot of benefit than what we could easily do with our `close`, and implementing `approx` in my pytest clone would have been too much work (it seems to rely on a lot of magic methods to do its magic).

src/compas/geometry/_core/basic.py
tomvanmele: coolio

PR #527, opened 2020-04-24
Move ironpython into its own project

gonzalocasas:
Moves the ironpython pytest stuff into [its own project](https://pypi.org/project/ironpython-pytest/), because we're starting to use it on compas-fab as well.

### Checklist


tests/ipy_test_runner.py
```diff
@@ -1,205 +1,14 @@
...
-
-        print_title('short test summary info')
-        for key, result in collected_errors.items():
-            print('FAILED {} - {}'.format(key, result['exception_message']))
+    pytest.load_fake_module('Rhino')
+    pytest.load_fake_module('Rhino.Geometry', fake_types=['RTree', 'Sphere', 'Point3d'])
 
-    print_title('{} failed, {} passed in {:.2f}s'.format(errors, counter - errors, end_time - start_time))
-    sys.exit(errors)
+    pytest.run(HERE, ['tests/compas/files/test_base_reader.py'])
```
tomvanmele: fyi,
i "removed" the base reader test from the test suite. renamed it to "test_base_reader.py.tmp"

tomvanmele: impressive simplification of the script though :)

tomvanmele (commented):
other than my few comicals, LGTM :)

tests/ipy_test_runner.py
tetov: > fyi,
> i "removed" the base reader test from the test suite. renamed it to "test_base_reader.py.tmp"

Should dependency on pathlib/pathlib2 be removed for compatibility with IPY?

gonzalocasas: @tetov something definitely needs to be done, because the code does not run on rhino

tetov: @gonzalocasas will fix, tracking in #529.

PR #528, opened 2020-04-25
Typo in EdgeArtist

tetov:
Seems like a typo.

There's also a problem with `draw_node`, `'Network' object has no attribute 'vertices'`. Is there a way to write a mixin that works with both nodes and vertices?

### Checklist

- ~~I have added necessary documentation (if appropriate)~~


gonzalocasas (approved):
LGTM!

src/compas_rhino/artists/mixins/edgeartist.py
```diff
@@ -90,9 +90,10 @@ def draw_edges(self, keys=None, color=None):
...
         lines = []
         for u, v in keys:
+            start, end = self.datastructure.edge_coordinates(u, v)
             lines.append({
-                'start': self.datastructure.vertex_coordinates(u),
-                'end': self.datastructure.vertex_coordinates(v),
+                'start': start,
+                'end': end,
                 'color': colordict[(u, v)],
                 'name': self.datastructure.edge_name(u, v),
```
tomvanmele: would you mind fixing this as well?
the `edge_name` method does not exits anymore. see above...

tetov: No problem!

tetov:
@brgcode in `clear_edge_labels` see [line](https://github.com/compas-dev/compas/blob/661227df0c2038acd5182d034a5bd23d79ee4811/src/compas_rhino/artists/mixins/edgeartist.py#L54). `edge_label_name()` is also removed right?

Methods with same name is also found in NetworkArtist, how is the mixin working here? There seems to be a bit of duplication.

Also, did you see my question up top?
> There's also a problem with draw_node, 'Network' object has no attribute 'vertices'. Is there a way to write a mixin that works with both nodes and vertices?

tomvanmele:
yes all these name functions have been removed.

i am working my way through the rhino package both in code and the docs. so will have a look...

as for the mixins, these clearly need a big update and may not be that relevant anymore, since the network and mesh now have less overlap...

tetov:
@brgcode: 

Ok! I can remove all the name functions. But I'll have to add a check to see if `mesh.edge_attribute((u, v) 'name')` returns `None` and skip adding that to the dict, right?

EDIT: I see now that `draw_labels` uses get and a default value. So I guess adding `None` values should be fine.

> i am working my way through the rhino package both in code and the docs. so will have a look...
> as for the mixins, these clearly need a big update and may not be that relevant anymore, since the network and mesh now have less overlap...

Ok, please let me know if I can help there.

tomvanmele:
no, the name is constructed according to a fixed pattern:
```python
# vertex
"{}.vertex.{}".format(self.datastructure.name, key)
```
```python
# edge
"{}.edge.{}-{}".format(self.datastructure.name, u, v)
```
```python
# face
"{}.face.{}".format(self.datastructure.name, fkey)
```

src/compas_rhino/artists/mixins/edgeartist.py
```diff
@@ -30,7 +30,7 @@ def clear_edges(self, keys=None):
         else:
             guids = []
             for u, v in keys:
-                name = self.datastructure.edge_name(u, v)
+                name = self.datastructure.edge_attribute((u, v), 'name')
```
tomvanmele: this should follow the same logic as the name being given...
```python
name = "{}.edge.{}-{}".format(self.datastructure.name, u, v)
```

PR #533, opened 2020-04-29
Added a new way to verify compas installation

gonzalocasas:
Instead of running `python` and then `import compas; print(compas.__version__)` to verify COMPAS installation, this PR adds a `__main__` so that the check can be a one-liner:

```bash
python -m compas
```

This is how it looks when used from CPython:
image

And this is from IronPython (ie cannot print extended ascii codes for some reason):
image

### Checklist


yck011522:
The ASCII art is so cool. 🤘

PR #534, opened 2020-04-30
Fix compas.robots.Joint not decode correctly by jsonpickle

yck011522:
I use `jsonpickle` for automatically serializing my classes. However, this does not work for my classes that inherit from `compas.robots.Joint`. 

The problem comes from the implementation of the function `self.calculate_transformation()`, which is overridden at __init__. This confused jsonpicke when decoding an object, because __init__ is not called by default, and the (unimplemented) self.calculate_transformation() is not overridden. 

This PR moves the implementation of the function switching logic to within the  `self.calculate_transformation()` itself. And will cause jsonpickle to decode an object successfully.

Moreover, as a discussion for perhaps future coding guide. It is better for the unimplemented function that is intended to be overridden (such as this one) to throw an Unimplemented Error instead of defaulting to `pass`. The lack of error message is very hard to debug.

### Checklist


src/compas/robots/model/joint.py
```diff
@@ -279,17 +279,6 @@ def __init__(self, name, type, parent, child, origin=None, axis=None, calibratio
         self.child_link = None
         self.position = 0
```
gonzalocasas, line 281: How about we delay the switching to the first call but cache the selected function?

```suggestion
        self._calculate_transformation = None
```

src/compas/robots/model/joint.py
```diff
@@ -427,14 +416,24 @@ def calculate_planar_transformation(self, position):
...
+        switcher = {
+            Joint.REVOLUTE: self.calculate_revolute_transformation,
+            Joint.CONTINUOUS: self.calculate_continuous_transformation,
+            Joint.PRISMATIC: self.calculate_prismatic_transformation,
+            Joint.FIXED: self.calculate_fixed_transformation,
+            Joint.FLOATING: self.calculate_floating_transformation,
+            Joint.PLANAR: self.calculate_planar_transformation
+        }
+        # set the transformation function according to the type
+        return switcher.get(self.type)(position)
```
gonzalocasas: To avoid switching constantly, we cache on first call.

```suggestion
        # Set the transformation function according to the type
        if not self._calculate_transformation:
            switcher = {
                Joint.REVOLUTE: self.calculate_revolute_transformation,
                Joint.CONTINUOUS: self.calculate_continuous_transformation,
                Joint.PRISMATIC: self.calculate_prismatic_transformation,
                Joint.FIXED: self.calculate_fixed_transformation,
                Joint.FLOATING: self.calculate_floating_transformation,
                Joint.PLANAR: self.calculate_planar_transformation
            }
            self._calculate_transformation = switcher.get(self.type)

        return self._calculate_transformation(position)
```

gonzalocasas (changes requested):
Generally LGTM, but I have one small suggestion (inlined in the other comments).

gonzalocasas:
The style checker is unhappy, it's missing its blank line: 
```
Running flake8 python linter...
src/compas/robots/model/joint.py:282:5: E301 expected 1 blank line, found 0
```
(You can see the full output here: https://travis-ci.com/github/compas-dev/compas/jobs/328251292, or clicking on the "Details" link of the PR)

Please fix and I think we can merge

yck011522:
@gonzalocasas I changed the implementation of your cache implementation, it is not possible to have a `self._calculate_transformation` initialized as `None` in the `__init__()` and check for it during first call.

 jsonpickle does not seem to serialize function as objects. Meaning `_calculate_transformation` after being modified by the fist call will be a function object. And this will not be serialized. The deserialized object (having not gone through the __init__) will have no attribute `_calculate_transformation` and thus throws error at the if statement where you check for `None`. 

I changed the implementation such that `self._calculate_transformation` simply doesn't exist by default. And it is only initialized by `calculate_transformation()`. This will pass the serialization.

On a side note, I would like to add a test to check if this object (probably the entire RobotModel object with links and joints) survive serialization. This however require adding jsonpickle into the requirements-dev.txt Do you think this is a good idea?

gonzalocasas (approved):
LGTM

PR #537, opened 2020-05-04
Key getter in FaceAttributeView

arpastrana:

### Checklist

### Description
Querying an attribute in  a `FaceAttributeView` raises a `KeyError` exception prematurely. This is because, at runtime, the key is also looked for _directly_ for in `self.defaults`, where it won't be found. 

The following snippet won't work.
```
import compas

mesh = Mesh.from_obj(compas.get("faces.obj"))
fkey = mesh.get_any_face()

mesh.face_attributes(key=fkey, names=["foo"], values=["bar"])

for f, attr in mesh.faces(True):
    if f != fkey:
        continue
    print(attr["foo"])
```

The proposal fixes the issue. By extension, I replaced a similar occurance in `EdgeAttributeView`. 
Not sure what is the expected behavior should be in all this though; `VertexAttributeView` seems to operate under a slightly different logic.


tomvanmele:
does this also fail if you add `"foo"` to the default attributes? i recently had another discussion about this and i think this behaviour actually makes sense...

arpastrana:
It probably won't fail adding `"foo"` to the default attributes beforehand. What I find confusing though is why `attr["foo"]` will raise a `KeyError` exception after having assigned it with `mesh.face_attributes()` a few lines before.  Or is there a preferred way to set non-default attributes to faces?

brgcode:
the assignment is only for a specific face. the loop goes through all faces and will fail because there is no default attribute `'foo'`. the attribute only exists where it was explicitly defined. so i think the failure is the correct behaviour... (but it was already clear form that previous discussion i was referring to that not everyone agrees :)

tomvanmele:
well, with the `AttributeView` objects there is no real reason anymore for not setting a default for every attribute, because it won't create unnecessary overhead as before (the defaults are no longer copied to all elements and only custom assigned versions are stored).

so in the current implementation, the preferred way would be to set a default. this results in very explicit behaviour.

but based on that other discussion i was referring to and your feedback/confusion, it seems not everyone agrees with this :)

arpastrana:
I'm perhaps still too familiar with old ways of `get_` and`set_` attributes :) I can see how the explicitness comes handy and readable if most of the ingredients in the soup are declared at as defaults at, say, the start of a script. However, the flexibility of adding and querying new attributes purposely on the fly is very valuable too (likely missed). Don't know, just some thoughts.

tomvanmele:
there is already a "custom_only" flag. perhaps there can be a "declared_only" flag as well...

arpastrana:
And what would that be if we want to combine custom and declared attributes? Or is this strongly disincentivized? :)

gonzalocasas (changes requested):
This change would suppress completely exceptions when trying to access a non-existent attribute.

Doesn't sound right to me.
These methods should behave similar to the python standard library. A `dict` raises exceptions when non-existent items are accessed using the item accessor syntax. A [`defaultdict`](https://docs.python.org/3.8/library/collections.html#collections.defaultdict) -which would be pretty close to this behavior, also fails if the default factory raises exceptions.

Deviation from that behavior would need a pretty strong argument.

tomvanmele:
since the change to views, and thus the removal of overhead, i think we should enforce the use of defaults, even in on-the-fly scenarios. i can't think of a situation where this is not possible. otherwise other code that now relies on the consistent presence of values, will have to be full of if statements.

however, this practice should be clearly documented, and it clearly is not :) will work on that...

an example of extreme on-the-fly-ness that still works
```python
mesh = Mesh.from_xxx(...)
mesh.update_default_vertex_attributes({'a': 0, 'b': 1})
# do stuff with 'a' and 'b'
# oh wait i also need 'c'
mesh.update_default_vertex_attributes({'c': 2})
# do more stuff with 'c'
```

gonzalocasas:
@brgcode so, if I understand correctly, you agree that the change in this PR should not be applied, right?

tomvanmele:
@gonzalocasas yes, but let's leave it open until i have addressed it in the docs...

arpastrana:
I do agree with Exceptions being a desirable conditions when accessing non-existing attributes. That said, and noting that using `update_defaults` appears to be the recommended practice, where does this leave using e.g. `face_attributes()` as a `setter`? 

Also, `update_default_attributes()` may use some some syntactic sugar if it's going to show up recurrently to add stuff on the fly. What about dropping the "default", and using `update_attributes()` instead?

tomvanmele:
@arpastrana not sure what you are asking in the first part of your post....

about dropping "default". not sure why we would do this. to make the command name shorter? without "default" it seems to suggest that all attributes will be updated, not just the defaults. so i am not sure this is the way to go...

arpastrana:
@brgcode Sorry, let me rephrase.

Why this doesn't work, and what would be the "right" way do it?

```python
mesh = Mesh.from_xxx()

# I need attribute "foo" in face 0. Let's set it.
mesh.face_attribute(key=0, name="foo", value="bar")

# do stuff
for f, attr in mesh.faces(True):
    if f == 0:
        # this raises a KeyException. Why If I have just set it above?
        msg = attr["foo"]  
        print("This is a {} attribute".format(msg))
````

**Comment by gonzalocasas** [comment]
@arpastrana I'm sorry, you're totally right. I misunderstood the problem. Indeed, the issue is that, the default is tried to be retrieved BEFORE we know if the key is present in the `attrs` at all, so, if there's no default, it fails, as you point out.

Your fix is correct, but I would really like to keep the exception raising if there is really no default AND no value. How about the following check instead?
```python
        return self.attr[self.key].get(name) or self.defaults[name]
```

Also, this is scream for some unit tests, whatever the fix, please add a few, it will save a ton of trouble in the future.

arpastrana:
Sounds good, thank you. ~~What I have proposed with the update is~~:
```python
return self.attr[self.key][name] or self.defaults[name]  # doesn't work
```
~~...to not mix `get` and `[]` in the same expression~~. I can suggest some unit tests in a separate PR.

EDIT: `get` is actually needed, otherwise it will raise an exception before looking for `self.defaults[name]` if `name` is not in `self.attr[self.key]`. So it is staying as initially suggested:

```python
return self.attr[self.key].get(name) or self.defaults[name]
```

gonzalocasas:
Code looks good to me! 👍 

> I can suggest some unit tests in a separate PR.

If possible, please do it in this same PR, it's a very good (best!) practice to include tests together with the fixing bugs, to a) avoid regressions and b) show that your code works as you expect in all platforms.

arpastrana:
Great, will do so! 😄 @gonzalocasas

arpastrana:
Hi @gonzalocasas! I've added a few test cases for the attribute views (faces, vertices, edges). After cleaning-up, the three child classes of AttributeView() were almost identical, so I am proposing to encapsulate the functionality in the parent class for now. 

Now I'm also wondering if dropping `self.key` would actually make sense in order to make an attribute view *only*  of the face/vertex/edge in question (narrowing scope)...As opposed to pointing to the complete parent dict every single time a view is created (`self.vertex`, `self.edgedata`, `self.facedata`)...

EDIT: It looks like there's some error with `pytest` which made the build fail in travis. `invoke test` in my computer doesn't throw any exceptions...could you please help me out to spot what the issue might be? thanks!

src/compas/datastructures/mesh/core/attributes.py
```diff
@@ -0,0 +1,72 @@
...
+            for name in self.defaults:
+                yield name
+
+
+class VertexAttributeView(AttributeView, MutableMapping):
+    """Mutable Mapping that provides a read/write view of the custom attributes of a vertex
+    combined with the default attributes of all vertices."""
+
+    def __init__(self, defaults, attr, key, custom_only=False):
+        super().__init__(defaults, attr, key, custom_only)
```
tomvanmele: don't know if the transpiler is already in place. otherwise this doesn't work in ipy 2.7...

src/compas/datastructures/mesh/core/attributes.py
```diff
@@ -0,0 +1,72 @@
...
+
+
+class EdgeAttributeView(AttributeView, MutableMapping):
+    """Mutable Mapping that provides a read/write view of the custom attributes of an edge
+    combined with the default attributes of all edges."""
+
+    def __init__(self, defaults, attr, key, custom_only=False):
+        super().__init__(defaults, attr, key, custom_only)
```
tomvanmele: don't know if the transpiler is already in place. otherwise this doesn't work in ipy 2.7...

src/compas/datastructures/mesh/core/attributes.py
```diff
@@ -0,0 +1,72 @@
...
+
+
+class FaceAttributeView(AttributeView, MutableMapping):
+    """Mutable Mapping that provides a read/write view of the custom attributes of a face
+    combined with the default attributes of all faces."""
+
+    def __init__(self, defaults, attr, key, custom_only=False):
+        super().__init__(defaults, attr, key, custom_only)
```
tomvanmele: don't know if the transpiler is already in place. otherwise this doesn't work in ipy 2.7...

src/compas/datastructures/mesh/core/attributes.py
```diff
@@ -0,0 +1,72 @@
...
+
+class AttributeView(object):
+    """Mixin for attribute dict views."""
+
+    def __init__(self, defaults, attr, key, custom_only):
+        self.defaults = defaults
+        self.attr = attr
+        self.key = key
+        self.custom_only = custom_only
+        self.attr.setdefault(self.key, {})
```
tomvanmele: this is problematic for vertices. if a user checks for the data of a non-existing vertex it actually gets created... looking at the old implementation, there should be a guard for this as well, in order to not create non-existing edges or faces in the data dicts.

src/compas/datastructures/mesh/core/attributes.py
```diff
@@ -0,0 +1,72 @@
...
+    from collections import MutableMapping      # python 2.7
+
+
+__all__ = ['VertexAttributeView', 'FaceAttributeView', 'EdgeAttributeView']
+
+
+class AttributeView(object):
+    """Mixin for attribute dict views."""
+
+    def __init__(self, defaults, attr, key, custom_only):
```
tomvanmele: the default for custom-only should be false.

tests/compas/datastructures/test_halfedge.py
```diff
@@ -0,0 +1,95 @@
...
+
+from compas.datastructures.mesh.core import HalfEdge
+
+
+# ==============================================================================
+# Fixtures
+# ==============================================================================
+
+@pytest.fixture
+def halfedge():
```
tomvanmele: in my opinion, an instance of HalfEdge should be called mesh

tests/compas/datastructures/test_halfedge.py
```diff
@@ -0,0 +1,95 @@
+import pytest
+
+from compas.datastructures.mesh.core import HalfEdge
```
tomvanmele: this should import from mesh, not core...

tomvanmele: actually, from datastructures

tomvanmele (changes requested):

src/compas/datastructures/mesh/core/attributes.py
gonzalocasas: not yet in use, and indeed, the build is failing due to it: 
image

src/compas/datastructures/mesh/core/attributes.py
arpastrana: Can we just suppress that line and let it raise a `KeyError` if `self.key` is not in `self.attr`? 

Or, can we omit `self.key` completely, and pass directly `XAttributeView(attr=facedata[key])` as opposed to `XAttributeView(attr=facedata, key=key)`...?

src/compas/datastructures/mesh/core/attributes.py
arpastrana: Noted and addressed. As elsewhere in the framework, now it reads: 
```python
super(XAttributeView, self).__init__()
```
This should work to pass the build on IPY 2.7

src/compas/datastructures/mesh/core/attributes.py
arpastrana: Changed that. Thanks.

tests/compas/datastructures/test_halfedge.py
arpastrana: Don't know...but changed it to that 😄

tests/compas/datastructures/test_halfedge.py
arpastrana: Updated.

src/compas/datastructures/mesh/core/attributes.py
tomvanmele: yes but the line is needed for the edge data and face data views. there is not necessarily an entry in those dicts for existing edges and faces.

tomvanmele: in that sense `XAttributeView(attr=facedata[key])` will not work. unless we do something like `XAttributeView(attr=facedata.setdefault(key, {}))` or if we make sure that entries in the attr dicts are always added when the items are added, but that is not so easy to enforce and it creates unnecessary overhead in many algorithms...

src/compas/datastructures/mesh/core/attributes.py
tomvanmele: the tests are still failing for ipy 2.7 though

src/compas/datastructures/mesh/core/attributes.py
tomvanmele: just to clarify, for me this init function cannot be used for vertices, because checking the attributes of a non-existing vertex would actually add it, and that would be an extremely unexpected side-effect.

src/compas/datastructures/mesh/core/attributes.py
```diff
@@ -0,0 +1,72 @@
...
+            for name in self.defaults:
+                yield name
+
+
+class VertexAttributeView(AttributeView, MutableMapping):
+    """Mutable Mapping that provides a read/write view of the custom attributes of a vertex
+    combined with the default attributes of all vertices."""
+
+    def __init__(self, defaults, attr, key, custom_only=False):
+        super(VertexAttributeView, self).__init__(defaults, attr, key, custom_only)
```
tomvanmele, line 55: with the current base class, this will produce unexpected side-effects...

tests/compas/datastructures/test_halfedge.py
```diff
@@ -0,0 +1,95 @@
...
+
+    faces = [
+        [0, 1, 2],
+        [0, 3, 1]
+    ]
+
+    he = HalfEdge()
+
+    for vertex in vertices:
+        he.add_vertex(attr_dict={c: xyz for c, xyz in zip("xyz", vertex)})
```
tomvanmele: why not just write this out?
tests should be as clear as possible and this is unnecessarily unclear to me...

tests/compas/datastructures/test_halfedge.py
```diff
@@ -0,0 +1,95 @@
...
+
+# ==============================================================================
+# Tests - Custom Attributes
+# ==============================================================================
+
+def test_set_vertex_attribute_custom(mesh, custom_kwattr):
+    key = mesh.get_any_vertex()
+    for name, value in custom_kwattr.items():
+        mesh.vertex_attribute(key=key, name=name, value=value)
+        assert name in mesh.vertex_attributes(key=key)
```
tomvanmele: should we not also assert the value?

tests/compas/datastructures/test_halfedge.py
```diff
@@ -0,0 +1,95 @@
...
+
+
+def test_set_face_attribute_custom(mesh, custom_kwattr):
+    key = mesh.get_any_face()
+    for name, value in custom_kwattr.items():
+        mesh.face_attribute(key=key, name=name, value=value)
+        assert name in mesh.face_attributes(key=key)
```
tomvanmele: assert value?

tests/compas/datastructures/test_halfedge.py
```diff
@@ -0,0 +1,95 @@
...
+
+
+def test_set_edge_attribute_custom(mesh, custom_kwattr):
+    key = (0, 1)
+    for name, value in custom_kwattr.items():
+        mesh.edge_attribute(key=key, name=name, value=value)
+        assert name in mesh.edge_attributes(key=key)
```
tomvanmele: assert value?

tests/compas/datastructures/test_halfedge.py
```diff
@@ -0,0 +1,95 @@
...
+        for k, v in default_kwattr.items():
+            assert k in attr
+
+
+@pytest.mark.parametrize("iterator", iterators())
+def test_get_attributes_not_in_default(mesh_default_attrs, custom_kwattr, iterator):
+    with pytest.raises(KeyError):
+        for vkey, attr in getattr(mesh_default_attrs, iterator)(True):
+            for k, v in custom_kwattr.items():
+                _ = attr[k]
```
tomvanmele: not very clear to me what this does...

tomvanmele:
i like the fact that the attributes are moved out of the data structure definition and that they are being tested. that said i have the feeling that the tests are unnecessarily complicated. it is not always immediately clear to me what is being tested and that seems not a good characteristic of a test.

in my opinion the init function of the base class cannot be used for vertex attribute views. perhaps we should therefore not change too many things at once for such an important feature and first simply fix the behaviour that you pointed out (while moving the views to their own module and adding a few simple tests).

perhaps the views also don't really have to be tested outside of the context of the data structure, because that context is important, which is demonstrated by the problem with the init function for the vertex views. the classes are not meant to be exposed to the user as standalone features.

finally, considering the discussions about schemas, quality control, predictable behaviour and what not, i would like to raise once more the question of allowing for not setting defaults...

perhaps i am overly critical about this PR, but if we don't get it right it will break a lot of code!

src/compas/datastructures/mesh/core/attributes.py
arpastrana: Makes sense. I will change that.

src/compas/datastructures/mesh/core/attributes.py
arpastrana: Noted 🙂

arpastrana:
@brgcode I understand your concerns and definitely appreciate the critical feedback! 🙂  

With the new commits I think I've addressed most of your comments, especially to not add a vertex to`HalfEdge.vertex` "artificially". Plus, the build has passed. Tests have been simplified and carried out in the context of the datastructure - they read a bit repetitive now though. Hope this is already pointing to the right direction.

In my opinion, there are a few other topics (outside of the scope of this PR) worth pondering later down the road:
- `Halfedge.add_face()` already takes care of inserting an empty dict to `Halfedge.facedata`, so `XAttributeView.attr.setdefault(XAttributeView.key, {})` and the supposed overhead, only applies to`EdgeAttributeView()`.
- Having `HalfEdge.face`-`HalfEdge.facedata`, `HalfEdge.halfedge`-`HalfEdge.edgedata` and only `HalfEdge.vertex` is inconsistent.
-`edge` and `halfedge` are interchanged discretionally all over the place. This is confusing.

Just a few thoughts.

tomvanmele:
the PR looks good to me now. @gonzalocasas last minute thoughts?

arpastrana:
Thanks for merging! @brgcode

Issue #540, opened 2020-05-06
[doc] documenting REPL example for `compas.rpc` and `compas.utilities.xfunc`


yijiangh:
# Feature Request

As a CPython package developer, I want to see a more elaborative example to demonstrate the develop/debug workflow ([REPL](https://en.wikipedia.org/wiki/Read%E2%80%93eval%E2%80%93print_loop)) when using `GHPython` + `compas.rpc` as a frontend input engine for testing my external CPython package.

To demonstrate this workflow, several questions need to be addressed in the doc but currently missing:
1. how can I work in the `debug` (`-e`) mode, i.e. having the changes in script be reflected in the rpc function call?
2. how and where can I see the print outputs from the CPython side?
3. If the CPython side has `input` to block the process to enable us to inspect the printouts or the GUI visualization, how can we "press enter" with `rpc`?
4. we can only use `Xfunc` or `Rpc` for python **packages** that have been installed in the conda environment, right? Can we call a script directly? (I had this question while trying to developing a test script for you to play with...)

## Details

**Is your feature request related to a problem? Please describe.**
Let's ground the context on a specific example, all the tests below is performed with `compas 0.13.3`:

Let's say that I'm developing a force density method python package called `pyfdm` that has been installed to the compas conda environment by:
```
conda activate compas_env
# cd into pyfdm's dir
pip install -e .
# expose this package to compas_rhino
python -m compas_rhino.install -p pyfdm
```

Now, I want to draw the initial grid and indicate fixities in Rhino/GH, and test
my package by making a function call in my GHPython component:

```
from compas.rpc import Proxy
fdm_rpc = Proxy('pyfdm')
updated_grid_pts = fdm_rpc.compute_fdm(grid_pts, connectivity, fixities, ...)
```
### Q1: debug mode
Even though from above, I've `pip install`ed the `pyfdm` package with the `-e` option, I found that I need to restart the proxy as follows to let the change be reflected in the rpc call:

```
fdm_rpc.stop_server()
fdm_rpc.start_server()
```
Since this takes a bit of runtime to kill and restart the proxy, and also it sometimes kill the `compas_rpc start` process in the external commandline. I'm wondering if there is a more principled way to do this?

### Q2: direct prints?
I'm confused about how we can direct the `print` happening in the CPython side to a visible place (either GHPython's output panel or an externally launched command-line prompt).

[...]

PR #541, opened 2020-05-07
RPC: add automodule reload and some other tweaks

gonzalocasas:
This PR addresses the issues/questions on #540, in concrete terms it does:

- [x] Add auto-reloading of modules that were imported/loaded by the RPC server in order to allow faster development cycles (Q1 of the linked issue)
- [x] Allow disabling the output (`stdout/stderr`) capture (Q2 and Q3 of linked issue)
- [x] Add some extra encoding support for `numpy` primitives

### Checklist


gonzalocasas:
@yijiangh if you have time, please check out/test this PR, it should solve some of the issues you mentioned on #540. Q1 should be automatically solved, as it will do autoreloading by default. Q2 and Q3 can be achieved by launching the proxy with the following params: `Proxy('your_package', python='python', capture_output=False)`, although I would actually suggest launching it externally from the command line will always be a little bit safer.

As for Q4, no, that is not possible at the moment, only installed modules.

tomvanmele:
i don't think the use of the ``Decoder`` will work. it will, for example, try to reconstruct a returned ``numpy`` ``array`` in Rhino, which doesn't work...

will have a closer look later tonight...

gonzalocasas:
@brgcode it gets converted into normal lists, but it seems to work:

image

tomvanmele:
ha sorry, made the wrong assumption, i see the decoder still converts the array to lists. then it should be fine indeed...

tomvanmele:
haven't tested anything but LGTM, and if travis doesn't complain ... :)
should we wait for Yijiyangh or just merge?

gonzalocasas:
@brgcode let's wait a bit, it's still 5:45am at MIT ;)

yijiangh:
Hmm, the debug mode seems not to work on my computer. This is what I did:

1. pull this PR into my local repo, `pip install .`, checked the version is `0.15.6`

2. `pip install -e .` my little dev package `compas_rpc_example`

3. Add a small test function to `compas_rpc_example/test_demo/test_demo.py`:
```
def test_connect():
    print('Haha')
    return 1
```

3. Install the package to `compas_rhino`
```
python -m compas_rhino.install
python -m compas_rhino.install -p compas_rpc_example
```

4. Open my Rhino/GH, open GHPython component, type in:
```
with Proxy('compas_rpc_example', python='python', capture_output=False) as p:
    a = p.test_connect()
    print a
```

The prints are still not showing up...

The following also does not work:
```
p = Proxy('compas_rpc_example')
a = p.test_connect()
print a
```

Do I need to **always** start RPC before I activate the gh component to stream the output there?

@gonzalocasas @tomvanmele I created a test package for you to play with, would you mind giving it a try as well?

The GH file can be found here: https://github.com/yijiangh/compas_rpc_example/blob/master/gh_examples/rpc_test.gh

The function can be found here:
https://github.com/yijiangh/compas_rpc_example/blob/master/src/compas_rpc_example/test_demo/test_demo.py

Feel free to play with the print, return number and add an `input` to stop the process. Thank you!

tomvanmele:
hi @yijiangh, thanks for setting this up. i will try to give feedback by monday...

yijiangh:
Sorry, my bad - I forgot to kill some of the running python instance when testing, and the rpc continues to reconnect to previous ones...

But I still have some questions: what's the correct procedure in invoking the rpc server?

1. Should I always start the server outside Rhino before I activate the GH script?
2. If I forget to start the external server, but activate the GH script first. Is there a way to redirect the output to the commandline?
3. I usually prefer not to use
```
with Proxy('pkg_name') as p:
    p.fn()
```
But rather use
```
p = Proxy('pkg_name')
```
to avoid reinit a proxy everytime I call the function. In this case, should I interpret the first way to call proxy as a way to replace 'XFunc'?

When I am testing it, I tried many order combination and in the end, my GHPy component tells me"
`the server is not available`.

tomvanmele:
yes, i think the first version indeed replaces `XFunc` in that sense, but (like you) i also prefer the other option because it keeps the server running which is kind of the point for quick successive calls.

can you give a brief overview of the options you tried and their outcome?
is this specific to GHPython or do you experience the same result in Rhino?

in my experience, in Rhino, there is no need to start the server in advance. this can just be triggered on the fly, if necessary, when you first create a Proxy object.

in RV2, for example, we create a proxy when the plugin is launched and then all calls while using the tool are really fast.

gonzalocasas:
> Sorry, my bad - I forgot to kill some of the running python instance when testing, and the rpc continues to reconnect to previous ones...

Just to understand, does that mean that you got the debug/autoreload thing to work?

yijiangh:
@gonzalocasas Hmm, just checked, still not working as expected.

Here are the steps that I took (after installing this compas commit and `compas_rhino.install (-p)`):

1. Start GHPython component, which starts a proxy in the following way:
```
p = Proxy('compas_rpc_example')
a = p.test_connect()
```

2. Observe that GHPython print the proper return value `20`, but does not show the print

3. Now, change the CPython package script to return value `30` instead of `20`, save the script. (I'm sure that this package is installed in a debug mode)

4. Rerun GHPython component, and observe that GHPython still gets the old return value. The picture below shows this result, notice that I start a proxy from the external prompt trying to capture the print, but with nothing shown up:

image

@brgcode The same thing happens for Rhino python as well:

image

tomvanmele:
@gonzalocasas the fully automated `watchdog` based reloading is cool, but perhaps it would be easier to first try an explicit "reload" modules command on the normal dispatcher? if that works we can revisit the `watchdog`...

gonzalocasas:
@yijiangh mmm.. something is fishy... I understand if module reloading is not triggering for whatever local setup reason, but if you can't get the prints to print out even when using `python='python', capture_output=False`, something is weird. I see you use powershell, could you try on a command prompt without it? PS is infamously unable to do normal things, like activating conda environments properly, so, if you happen to have compas also on the base environment, you might be ending up somewhere else.

> @gonzalocasas the fully automated `watchdog` based reloading is cool, but perhaps it would be easier to first try an explicit "reload" modules command on the normal dispatcher? if that works we can revisit the `watchdog`...

@brgcode how would you trigger an explicit reload?

tomvanmele:
similar like the `ping` function on the server:

```python
proxy.reload()
```

```
def reload(self):
    unload_modules(...)
```

gonzalocasas:
ah! sure, I can add that, good idea.

yijiangh:
@gonzalocasas I tried the real command prompt:tm:, with `python='python', output_capture=False` I can see the prints, but the debug mode is still not working:

image

I checked that `compas_rhino.install` symlink to the right repo in the debug mode:

image

Also, once I initiated the proxy with `python='python', output_capture=False`, I cannot change it to `python='pythonw', output_capture=True` to redirect it to an external prompt without stopping and restarting the proxy, right?

tomvanmele:
@gonzalocasas what is the plan with this?

tomvanmele:
@gonzalocasas ping

gonzalocasas:
I'll get back to it soon-ish.

docs/tutorial/rpc.rst
```diff
@@ -0,0 +1,98 @@
...
+a remote process or computer. This is especially useful in scenarios where COMPAS
+runs inside an IronPython host (eg. Rhino) and needs to execute code that only
+runs on CPython (eg. code that requires ``numpy``).
+
+COMPAS provides **two ways to achieve** this: ``rcp`` and `XFunc``.
+
+Through ``Xfunc``, COMPAS provides a mechanism for calling Python functions through
+a separately launched subprocess.
+
+A drawback of the ``Xfunc`` mechanism is that evey call launches a new Python
```
beverlylytle:
```suggestion
A drawback of the ``Xfunc`` mechanism is that every call launches a new Python
```

docs/tutorial/rpc.rst
```diff
@@ -0,0 +1,98 @@
...
+(sub)process with all the overhead that that entails. For infrequent calls to
+long-running processes this is not an issue. However, for frequent calls to function
```
beverlylytle:
```suggestion
long-running processes this is not an issue. However, for frequent calls to functions
```

docs/tutorial/rpc.rst
```diff
@@ -0,0 +1,98 @@
...
+that are expected to run quickly, this is not ideal.
+
+The second mechanism is the ``rpc`` module. The principle of RPC is to start a server
+that handles all requests. The advantage is that once the server is started,
+no additional processes have to launched and the server can handle the requests
```
beverlylytle:
```suggestion
no additional processes have to be launched and the server can handle the requests
```

docs/tutorial/rpc.rst
```diff
@@ -0,0 +1,98 @@
...
+
+    RPC implies that data needs to be converted back and forth. Internally, RPC uses JSON as a serialization
+    format. This means that some floating-point precision loss will occur when passing around ``numpy`` values.
+
+Switching packages
+==================
+
+The RPC proxy is usually initialized with one package name, but it can be switched at any point.
+The following example shows how to calculate the inverse of a ``numpy`` matrix using ``scipy``
+lineal algebra functions.
```
beverlylytle:
```suggestion
linear algebra functions.
```

src/compas/rpc/proxy.py
```diff
@@ -43,18 +41,26 @@ class Proxy(object):
...
+    url : :obj:`str`:, optional
         The server address.
         Default is ``'http://127.0.0.1'``.
-    port : int, optional
+    port : :obj:`int`:, optional
         The port number on the remote server.
         Default is ``1753``.
+    service : :obj:`str`:, package name to start server.
+        Default is ``'compas.rpc.services.default'``.
+    autoreload : bool, ``True`` to automatically reload the proxied package if changes are detected.
```
beverlylytle:
```suggestion
    autoreload : :obj:`bool`, ``True`` to automatically reload the proxied package if changes are detected.
```

src/compas/rpc/proxy.py
```diff
@@ -43,18 +41,26 @@ class Proxy(object):
...
+        This is particularly useful during development. The server will monitor changes to the files
+        that were loaded as a result of accessing the specified `package` and if any change is detected,
+        it will unload the module, so that the next invocation uses a fresh version.
+    capture_output : bool, ``True`` to capture the stdout/stderr output of the remote process, otherwise ``False.
```
beverlylytle:
```suggestion
    capture_output : :obj:`bool`, ``True`` to capture the stdout/stderr output of the remote process, otherwise ``False``.
```

docs/tutorial/rpc.rst
```diff
@@ -59,23 +63,25 @@ The use of :mod:`compas.rpc` is not restricted to COMPAS packages only. ::
...
+In combination with custom JSON encoders and decoders this allows for COMPAS
+objects to be serialized and de-serialized without loss of information on either
+side of the RPC communication network.
 
-Therefore the data types supported by :mod:`compas.rpc` include all native Python data types and COMPAS objects.
-Numpy arrays are automatically converted to lists.
+Therefore the data types supported by :mod:`compas.rpc` include all native Python
+data types and COMPAS objects. Numpy arrays are automatically converted to lists.
 
+linear algebra functions.
```
beverlylytle: This doesn't belong here.

tomvanmele: it seems to have appeared out of thin air :)

tomvanmele:
@gonzalocasas can you ping when it is finished?

gonzalocasas:
> @gonzalocasas can you ping when it is finished?

yep, i updated it to master now. I will fix the github action workflow and then re-test. From my side, it was fine and working back then, but it was not working proper on @yijiangh 's machine and I could not reproduce the issue on my side yet. I will try to and come back soon.

docs/tutorial/rpc.rst
gonzalocasas: sorry! some merge leftover I guess.... fixed now

gonzalocasas:
@yijiangh I think I finally fixed this one. The `autoreload` now should work for your example as well (before it was not recursing directories and your example had changes on a sub-directory which was not detected). I also teste the capturing of output much more in detail, and it works now as expected. However, it's important to highlight that -as you noticed- the behavior of the proxy cannot be changed if it's already running (ie. if you start it once with `python='pythonw'` and you use it again with  `python='python'` but not having stopped it before, that change will NOT take effect). As I have mentioned in the past, for this kind of workflow, I think it's better/easier to understand the if the RPC server is manually started from the command line explicitly. There's less magic, and a much more straightforward behavior. The auto-starting mode should work, but it has all these caveats about not being able to change its behavior once it's running.

Let me know if it works for you now! :)

gonzalocasas:
@yijiangh ping! 😃  if you have time, I'd love to hear your feedback about the current state. If not, no worries, we can merge and eventually fix if there are cases (hopefully edge cases) that don't work as you expect.

yijiangh:
Sorry for the delay! Here are the tests that I performed on my computer. All works as expected except the `capture_output` for the autostart mode.

## `compas_rpc start`  before opening Rhino/GH
All of them work!
- [x] debug autoreload
- [x] directing `print`
- [x] responding to `input()`

This is the output that I saw in my external command line prompt - works as expected! 🥇 
image


## autostart mode
- [x] debug autoreload
- [x] responding to `input()`
- [ ] directing `print`

I was expecting the prints will show up in the prompted python window or the output panel of GHPython, but this didn't happen:

image

Source code for the external module being called:
```
import numpy as np

def test_connect():
    print('haha')
    print('first')
    # input('Hit a key!')
    print('second')

    a = np.array([[1,2], [3,4,5,6,7]])
    return a
```

tomvanmele:
@gonzalocasas how about we just merge this thing in?

gonzalocasas:
> @gonzalocasas how about we just merge this thing in?

I was going to suggest exactly that. We can tackle that final remaining edge case later on.

gonzalocasas:
🎉 🥳

PR #546, opened 2020-05-20
Returns and Raises added to mesh_split_face docstring

tetov:

Doc update

### Checklist

- ~~I added a line to the `CHANGELOG.md` file in the `Unreleased` section under the most fitting heading (e.g. `Added`, `Changed`, `Removed`).~~
- ~~I ran all tests on my computer and it's all green (i.e. `invoke test`).~~

### DIFF
--- a/src/compas/datastructures/mesh/core/operations/split.py
+++ b/src/compas/datastructures/mesh/core/operations/split.py
@@ -181,6 +181,22 @@ def mesh_split_face(mesh, fkey, u, v):
     :exc:`ValueError`
         If the split vertices does not belong to the split face or if the split
         vertices are neighbors.
+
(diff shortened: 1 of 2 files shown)


tetov:
Tried to write a doctest too but it's a bit tricky since the funcs in this module are only(?) meant to be used as an method to Mesh.

tomvanmele:
the functions are indeed meant to take a mesh as first parameter. other than that they can perfectly be used as standalone functions. in fact, binding them to the mesh class is only a fairly recent change...

tetov:
@brgcode: Ok, added an example and some unit tests.

tomvanmele:
cool thanks

PR #547, opened 2020-05-22
Fixing sphere.point not transformed correctly in if transform is a Scale

yck011522:

### Checklist


gonzalocasas:
@romanarust: we just discussed this with @yck011522 and to me, the change makes sense, but since you recently changed this, you might have more info.

romanarust:
@yck011522 you are right, for applying the whole transformation on the point and not only the translation. But I'm afraid it is still not fully solved.
If you try the following in Rhino:
```
import Rhino.Geometry as rg

s = rg.Sphere(rg.Point3d(1, 1, 1), 10)
T = rg.Transform.Scale(rg.Plane.WorldXY, 2, 3, 4)
s.Transform(T)

print(s.Center)
print(s.Radius)
```
The radius results into 24.4948974278. 

I haven't been able to figure out what the expected behavior of applying a non-uniform Scale on a sphere is. Maybe related to the volume?

Any thoughts?

yck011522:
@romanarust  The rhino geometry example is weird. I cannot decipher what Rhino is doing with the radius. 

Conceptually non-uniform scaling of a sphere will result in an ellipsoid. But this is not possible with the sphere class structure we have. Assuming upgrading the class to a ellipsoid is out of the question. I believe from a API design stand-point, the behavior should be undefined. 

Referring to the [Rhino.Sphere.Transform() Documentation](https://developer.rhino3d.com/api/RhinoCommon/html/M_Rhino_Geometry_Sphere_Transform.htm): 

> Transforms this sphere. Note that non-similarity preserving transformations cannot be applied to a sphere as that would result in an ellipsoid.

So apparently Rhino developers also thought about this and the conclusion is that turning it into an ellipsoid doesn't make sense. However, as in your test code, it seems that Rhino didn't throw you an error but just performed some undefined behavior. 

Going back to our `compas.geometry.Sphere`. For practical purpose, I believe:
1. throwing an error is a reasonable behavior.
2. your previous implementation of using the largest scale factor `max()` to scale the radius is also a reasonable behavior. As long as it is documented.
3. Document 'applying non-similarity preserving transformation' will result in undefined behavior. And the implementation can be free to do whatever. (just like good old C++ undefined behavior UB)

romanarust:
Good suggestion! I would go with option 2) then.
Could you please add a note in the docstr?

yck011522:
Yes, I can.

src/compas/geometry/_shapes/sphere.py
```diff
@@ -294,7 +294,7 @@ def transform(self, transformation):
         """
         self.point.transform(transformation)
         Sc, _Sh, _R, _Tl, _P = transformation.decomposed()
-        self.radius *= max(Sc.scale_factors)
+        self.radius *= max([Sc[0,0],Sc[1,1],Sc[2,2]])
```
tomvanmele: those underscored variables might as well be underscores if they are not used...

yck011522: Fixed

src/compas/geometry/_shapes/sphere.py
```diff
@@ -289,9 +292,9 @@ def transform(self, transformation):
         >>> sphere.transform(T)
 
         """
-        Sc, Sh, R, Tl, P = transformation.decomposed()
-        self.point.transform(Tl)
-        self.radius *= max(Sc.scale_factors)
+        self.point.transform(transformation)
+        Sc, _, _, _, _ = transformation.decomposed()
```
tomvanmele, line 296: this is an expensive operation (decomposing the transformation matrix in independent components). scale and translation are completely independent. is this absolutely necessary?

yck011522: Do you have a suggestion?

tomvanmele: well if the matrix only contains scale and translation, you could just do
```python
factors = [transformation[0, 0], transformation[1, 1], transformation[2, 2]
vector = [transformation[0, 3], transformation[1, 3], transformation[2, 3]
Sc = Scale.from_factors(factors)
Tr = Translation.from_vector(vector)
```
no, @romanarust ?

gonzalocasas: But we cannot ensure the matrix only contains scale and translation.
Additionally, `transform` does imply a potentially costly operation, so, I would err on the side of correctness rather than optimize for speed while leaving room for potentially incorrect results.

tomvanmele: okay cool

PR #553, opened 2020-06-02
first contribute to compas geometry

nikeftekhar:
contributes regarding next topics:
**polyline**: tween, divide by count, rebuild, shorten
**line**: divide by count
**angles**: between planes
**extend**: lines, polylines
**intersections**: mesh-line, mesh-plane
**splits**: polyline-plane, mesh-plane

### Checklist


src/compas/geometry/_core/angles.py
```diff
@@ -352,6 +352,30 @@ def angles_points_xy(a, b, c, deg=False):
     v = subtract_vectors_xy(c, a)
     return angles_vectors_xy(u, v, deg)
 
+def angle_planes(a, b, deg=False):
+    r""" Compute the smallest angle between the two normal vectors of two planes.
```
tomvanmele: the raw string prefix is only necessary if you use latex syntax in the docstring...

src/compas/geometry/_core/angles.py
```diff
@@ -352,6 +352,30 @@ def angles_points_xy(a, b, c, deg=False):
...
+    float
+        The smallest angle in radians (in degrees if deg == True).
+        The angle is always positive.
+
+    Examples
+    --------
+    >>> angle_planes(plane_a, plane_b, True)
+
+    """
+
```
tomvanmele: please remove unnecessary whitespace

src/compas/geometry/_core/angles.py
```diff
@@ -352,6 +352,30 @@ def angles_points_xy(a, b, c, deg=False):
...
+    Returns
+    -------
+    float
+        The smallest angle in radians (in degrees if deg == True).
+        The angle is always positive.
+
+    Examples
+    --------
+    >>> angle_planes(plane_a, plane_b, True)
```
tomvanmele, line 378: the example would be more meaningful (and doctest-able) if you actually compute the angle between two planes, such as XY and YZ and output the result...

src/compas/geometry/_core/angles.py
```diff
@@ -352,6 +352,30 @@ def angles_points_xy(a, b, c, deg=False):
...
+
+    """
+
+    return angle_vectors(a.normal, b.normal, deg)
```
tomvanmele: this will only work if the input parameters are `Plane` objects. however, the base functions should be compatible with COMPAS objects and Python objects. since `Plane` objects implement the iteration protocol, the following would also allow the plane to be provided as a tuple of point and normal.
```python
return angle_vectors(a[1], b[1], deg)
```

src/compas/geometry/_core/intersections.py
```diff
@@ -382,6 +382,37 @@ def intersection_segment_plane(segment, plane, tol=1e-6):
 
     return None
 
+def intersection_polyline_plane(polyline, plane, expected_number_of_intersections=2, tol=1e-6):
+    """ Calculates the intersection point of a plane with a polyline. 
+        returns a list of intersections 
+        by default it will allow two intersections. reduce expected_number_of_intersections to speed up.
```
tomvanmele: Please remove whitespace between the triple quotes and the summary. Please also remove the indents of the second and third line.

src/compas/geometry/_core/intersections.py
```diff
@@ -382,6 +382,37 @@ def intersection_segment_plane(segment, plane, tol=1e-6):
...
+        Default is ``1e-6``.
+
+    Returns
+    -------
+    intersection_pts : list of compas.geometry.Point
+        if there are intersection points, return point(s) in a list
+    """
+    intersection_pts = []
+    max_iter =0
+    for segment in polyline.lines:
```
tomvanmele: core functions should be compatible with COMPAS geometry objects as well as their pure Python representations. a polyline is basically a sequence of points. lines are given by consecutive pairs of points. the `Polyline` supports the iteration protocol such that you can write the following.
```python
for segment in pairwise(polyline):
```

tomvanmele (commented):
for a first contribution this is quite elaborate. although the effort is much appreciated, perhaps it would be better to split this up into two PRs with a narrower scope. this way we can "practice" with the first PR and make the review process of the second one a bit easier. i would suggest limiting the first PR to the angles between planes and the new intersection functions in `_core/intersections.py`...

src/compas/geometry/_core/angles.py
nikeftekhar: oh! i didn't know that. good to know, copied. thank you very much for the comment

src/compas/geometry/_core/angles.py
nikeftekhar: yes, you are right, i will incorporate that

src/compas/geometry/_core/angles.py
nikeftekhar: copied, and changed

src/compas/geometry/_core/angles.py
nikeftekhar: wasn't really aware of that, but makes total sense. Thank you very much for the input. Really appreciated.

src/compas/geometry/_core/intersections.py
nikeftekhar: I will apply this to everything, thank you very much

src/compas/geometry/_core/intersections.py
nikeftekhar: perfect, understood. i am learning a lot. Thank you very much. really appreciated.

PR #563, opened 2020-06-10
fix #562: fix attribute view iteration performance issues

### LINKED ISSUE #562: Diagnose performance issues with vertex attribute view
During the resolution of #560, we noticed the vertex attribute view object has some severe performance issues when running on IronPython. The issue was resolved side-stepping the view, but the underlying problem of this class should be diagnosed and resolved.

**To Reproduce**

1. Open Rhino editor or run the following script using IronPython CLI (`ipy`):
```python
from timeit import default_timer as timer
import compas

t0 = timer()
mesh = Mesh.from_ply(compas.get_bunny())
print('mesh load time: {:.2f}s'.format(timer() - t0))

t1 = timer()
for _ in mesh.vertices(False):
    pass
print('vertex iterator without attribute view: {:.2f}s'.format(timer() - t1))

t2 = timer()
for _ in mesh.vertices(True):
    pass
print('vertex attribute view iterator: {:.2f}s'.format(timer() - t2))
```
which will output something like:
```
mesh load time: 2.26s
vertex iterator without attribute view: 0.03s
vertex attribute view iterator: 125.96s
```

For reference, the exact same script outputs the following when executed using cpython:
```
mesh load time: 2.09s
vertex iterator without attribute view: 0.01s
vertex attribute view iterator: 0.13s
```

**Minimal reproduction**
After some more debugging, it turns out that the problem is the `ABCMeta` metaclass, which somehow causes IronPython to go crazy if hundreds of instances are created in a row. The following minimal code snippets reproduces the issue in isolation, and the exact same code, but removing the `__metaclass__` declaration works perfectly fine:

```python
from abc import ABCMeta

class T(object):
    __metaclass__ = ABCMeta

    def __init__(self):
        super(T, self).__init__()

class J(T):
    def __init__(self):
        super(J, self).__init__()

for i in range(100000):
    a = J()
    if i % 200 == 0:
        print(i)
```

**Expected behavior**
The run time of iterating vertices with attributes should **not be 4000x slower** 😛 

 - OS: Windows
 - Python version: IronPython 2.7.8 / Rhino 6
 - Python package manager: `conda`
 - compas version: `0.16.1`

gonzalocasas:
Added vendored MutableMapping and Mapping implementations that have been modified to exclude the `ABCMeta` metaclass (see linked issue #562 for details)

### Checklist

### DIFF
--- a/src/compas/datastructures/mesh/core/_mutablemapping.py
+++ b/src/compas/datastructures/mesh/core/_mutablemapping.py
@@ -1,0 +1,149 @@
+"""
+The module contains vendored copies of a couple of classes from collections.abc
+(python standard library) with one fundamental change:
+
+ * The Mapping class does not have a __metaclass__ = ABCMeta set
+   because this causes performance issues on IronPython 2.7.x
+
+See this issue for more details: https://github.com/compas-dev/compas/issues/562
+"""
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+import compas
+
+__all__ = ['Mapping', 'MutableMapping']
+
+if compas.PY3:
+    import collections.abc as stdlib_collections
+else:
+    import collections as stdlib_collections
+
+
+class Mapping(object):
+    __slots__ = ()
+
+    """A Mapping is a generic container for associating key/value
+    pairs.
+    This class provides concrete generic implementations of all
+    methods except for __getitem__, __iter__, and __len__.
+    """
+
+    def get(self, key, default=None):
+        'D.get(k[,d]) -> D[k] if k in D, else d.  d defaults to None.'
+        try:
+            return self[key]
+        except KeyError:
+            return default
+
+    def __contains__(self, key):
+        try:
+            self[key]
+        except KeyError:
+            return False
+        else:
+            return True
+
+    def keys(self):
+        "D.keys() -> a set-like object providing a view on D's keys"
+        return stdlib_collections.KeysView(self)
+
+    def items(self):
+        "D.items() -> a set-like object providing a view on D's items"
+        return stdlib_collections.ItemsView(self)
+
+    def values(self):
+        "D.values() -> an object providing a view on D's values"
+        return stdlib_collections.ValuesView(self)
+
+    def __eq__(self, other):
+        if not isinstance(other, Mapping) and not isinstance(other, stdlib_collections.Mapping):
+            return NotImplemented
+        return dict(self.items()) == dict(other.items())
+
+    __reversed__ = None
+
+
+class MutableMapping(Mapping):
+
+    __slots__ = ()
+
+    """A MutableMapping is a generic container for associating
+    key/value pairs.
+    This class provides concrete generic implementations of all
+    methods except for __getitem__, __setitem__, __delitem__,
+    __iter__, and __len__.
+    """
+
+    __marker = object()
+
+    def pop(self, key, default=__marker):
+        '''D.pop(k[,d]) -> v, remove specified key and return the corresponding value.
+          If key is not found, d is returned if given, otherwise KeyError is raised.
+        '''
+        try:
+            value = self[key]
+        except KeyError:
+            if default is self.__marker:
+                raise
+            return default
+        else:
+            del self[key]
+            return value
+
+    def popitem(self):
+        '''D.popitem() -> (k, v), remove and return some (key, value) pair
+           as a 2-tuple; but raise KeyError if D is empty.
+        '''
+        try:
+            key = next(iter(self))
+        except StopIteration:
+            raise KeyError
+        value = self[key]
+        del self[key]
+        return key, value
+
+    def clear(self):
+        'D.clear() -> None.  Remove all items from D.'
+        try:
+            while True:
+                self.popitem()
+        except KeyError:
+            pass
+
+    def update(*args, **kwds):
+        ''' D.update([E, ]**F) -> None.  Update D from mapping/iterable E and F.
+            If E present and has a .keys() method, does:     for k in E: D[k] = E[k]
+            If E present and lacks .keys() method, does:     for (k, v) in E: D[k] = v
+            In either case, this is followed by: for k, v in F.items(): D[k] = v
+        '''
+        if not args:
+            raise TypeError("descriptor 'update' of 'MutableMapping' object "
+                            "needs an argument")
+        self = args[0]
+        args = args[1:]
+        if len(args) > 1:
+            raise TypeError('update expected at most 1 arguments, got %d' %
+                            len(args))
+        if args:
+            other = args[0]
+            if isinstance(other, Mapping) or isinstance(other, stdlib_collections.Mapping):
+                for key in other:
+                    self[key] = other[key]
+            elif hasattr(other, "keys"):
+                for key in other.keys():
+                    self[key] = other[key]
+            else:
+                for key, value in other:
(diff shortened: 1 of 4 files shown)


src/compas/datastructures/mesh/core/_mutablemapping.py
```diff
@@ -0,0 +1,149 @@
...
+    __slots__ = ()
+
+    """A Mapping is a generic container for associating key/value
+    pairs.
+    This class provides concrete generic implementations of all
+    methods except for __getitem__, __iter__, and __len__.
+    """
+
+    def get(self, key, default=None):
+        'D.get(k[,d]) -> D[k] if k in D, else d.  d defaults to None.'
```
tomvanmele: ' => """

src/compas/datastructures/mesh/core/_mutablemapping.py
```diff
@@ -0,0 +1,149 @@
...
+    def __contains__(self, key):
+        try:
+            self[key]
+        except KeyError:
+            return False
+        else:
+            return True
+
+    def keys(self):
+        "D.keys() -> a set-like object providing a view on D's keys"
```
tomvanmele: " => """

src/compas/datastructures/mesh/core/_mutablemapping.py
```diff
@@ -0,0 +1,149 @@
...
+    key/value pairs.
+    This class provides concrete generic implementations of all
+    methods except for __getitem__, __setitem__, __delitem__,
+    __iter__, and __len__.
+    """
+
+    __marker = object()
+
+    def pop(self, key, default=__marker):
+        '''D.pop(k[,d]) -> v, remove specified key and return the corresponding value.
```
tomvanmele: ''' => """

src/compas/datastructures/mesh/core/_mutablemapping.py
```diff
@@ -0,0 +1,149 @@
...
+            while True:
+                self.popitem()
+        except KeyError:
+            pass
+
+    def update(*args, **kwds):
+        ''' D.update([E, ]**F) -> None.  Update D from mapping/iterable E and F.
+            If E present and has a .keys() method, does:     for k in E: D[k] = E[k]
+            If E present and lacks .keys() method, does:     for (k, v) in E: D[k] = v
+            In either case, this is followed by: for k, v in F.items(): D[k] = v
```
tomvanmele: loose the indentations

src/compas/datastructures/mesh/core/_mutablemapping.py
```diff
@@ -0,0 +1,149 @@
...
+
+    def clear(self):
+        'D.clear() -> None.  Remove all items from D.'
+        try:
+            while True:
+                self.popitem()
+        except KeyError:
+            pass
+
+    def update(*args, **kwds):
```
tomvanmele: although it doesn't matter, these tend to be named `**kwargs` (keyword args)

src/compas/datastructures/mesh/core/_mutablemapping.py
```diff
@@ -0,0 +1,149 @@
...
+            raise TypeError("descriptor 'update' of 'MutableMapping' object "
+                            "needs an argument")
+        self = args[0]
+        args = args[1:]
+        if len(args) > 1:
+            raise TypeError('update expected at most 1 arguments, got %d' %
+                            len(args))
+        if args:
+            other = args[0]
+            if isinstance(other, Mapping) or isinstance(other, stdlib_collections.Mapping):
```
tomvanmele: i believe this can be 
```
if isinstance(other, (Mapping, stdlib_collections.Mapping)):

File: src/compas/datastructures/mesh/core/_mutablemapping.py [hunk 7]
```diff
@@ -0,0 +1,149 @@
...
+            pass
+
+    def update(*args, **kwds):
+        ''' D.update([E, ]**F) -> None.  Update D from mapping/iterable E and F.
+            If E present and has a .keys() method, does:     for k in E: D[k] = E[k]
+            If E present and lacks .keys() method, does:     for (k, v) in E: D[k] = v
+            In either case, this is followed by: for k, v in F.items(): D[k] = v
+        '''
+        if not args:
+            raise TypeError("descriptor 'update' of 'MutableMapping' object "
```
tomvanmele: why "descriptor"?

src/compas/datastructures/mesh/core/_mutablemapping.py
tomvanmele: btw, considering all the `args` checking, why is the signature not just
```
def update(self, other, **kwargs):

File: src/compas/datastructures/mesh/core/_mutablemapping.py [hunk 8]
```diff
@@ -0,0 +1,149 @@
...
+    __marker = object()
+
+    def pop(self, key, default=__marker):
+        '''D.pop(k[,d]) -> v, remove specified key and return the corresponding value.
+          If key is not found, d is returned if given, otherwise KeyError is raised.
+        '''
+        try:
+            value = self[key]
+        except KeyError:
+            if default is self.__marker:
```
tomvanmele, line 88: is this ever the case? won't the markers always have a different id?

tomvanmele (commented):
some minor comments (docstring consistency) and a few questions....

src/compas/datastructures/mesh/core/_mutablemapping.py
gonzalocasas: yup, addressed on https://github.com/compas-dev/compas/pull/563/commits/a6fa22e344af2b61700d1a03d120c34620059955

src/compas/datastructures/mesh/core/_mutablemapping.py
gonzalocasas: got rid of it

src/compas/datastructures/mesh/core/_mutablemapping.py
gonzalocasas: The code is fundamentally copy&pasted entirely from the python std lib, so I rather not change this behavior. That said, I think it is correct, the case can happen if there's no default (ie, default will equal to `__marker` as per `__init__`) .

gonzalocasas:
I addressed most of the comments on https://github.com/compas-dev/compas/pull/563/commits/a6fa22e344af2b61700d1a03d120c34620059955

src/compas/datastructures/mesh/core/_mutablemapping.py
```diff
@@ -31,7 +31,7 @@ class Mapping(object):
     """
 
     def get(self, key, default=None):
-        'D.get(k[,d]) -> D[k] if k in D, else d.  d defaults to None.'
+        'D.get(k[,d]) => D[k] if k in D, else d.  d defaults to None.'
```
tomvanmele: ha sorry, that is not what i meant. the docstrings should be enclosed in `"""` not `'` or `'''` or `"`

tomvanmele: `->` and `=>` are both fine by me :)

gonzalocasas: ahh! true, fixed at https://github.com/compas-dev/compas/pull/563/commits/a352f6b758db9bb4d70d4ad58a3d7bf0cdf5da33

PR #569, opened 2020-06-24
new functions for compas geometry

nikeftekhar:
new functions regarding the following topics:
line divide by count
polyline shorten
polyline rebuild
polyline divide by count
polyline tween
transformations extend line
transformations extend polyline
intersection_mesh_line
intersection_mesh_plane
split polyline_plane
split mesh_plane

### Checklist


src/compas/geometry/_intersections/mesh_intersections.py
```diff
@@ -0,0 +1,102 @@
...
+            intersection = intersection_line_triangle(line, triangle)
+            if intersection:
+                return Point(intersection[0], intersection[1], intersection[2])
+            if len(vertex_keys) == 4:
+                triangle_2 = [vertices[2], vertices[3], vertices[0]]
+                intersection_2 = intersection_line_triangle(line, triangle_2)
+                if intersection_2:
+                    return Point(intersection_2[0], intersection_2[1], intersection_2[2])
+        else:
+            continue
```
gonzalocasas: To reduce nesting, how about inverting the check?
```suggestion
        if len(vertex_keys) not in (3, 4):
            continue

        triangle = [vertices[0], vertices[1], vertices[2]]
        intersection = intersection_line_triangle(line, triangle)
        if intersection:
            return Point(intersection[0], intersection[1], intersection[2])

        if len(vertex_keys) == 4:
            triangle_2 = [vertices[2], vertices[3], vertices[0]]
            intersection_2 = intersection_line_triangle(line, triangle_2)
            if intersection_2:
                return Point(intersection_2[0], intersection_2[1], intersection_2[2])
```

src/compas/geometry/_intersections/mesh_intersections.py
```diff
@@ -0,0 +1,102 @@
...
+    Returns
+    -------
+    intersections: list of points as keys from mesh
+    """
+      
+    intersections = [] 
+    for u, v in list(mesh.edges()):
+        a = mesh.vertex_attributes(u,'xyz')
+        b = mesh.vertex_attributes(v,'xyz')
+        inters = intersection_segment_plane((a,b), plane)
```
gonzalocasas: don't use "non-standard" abbreviations, better to have longer, clear names, than short ones that are hard to parse by humans.

src/compas/geometry/_intersections/mesh_intersections.py
```diff
@@ -0,0 +1,102 @@
...
+        if not inters:
+            continue
+        len_a_inters = length_vector(subtract_vectors(inters, a))
+        len_a_b = length_vector(subtract_vectors(b, a))
+        t = len_a_inters / len_a_b
+        if t>= 1.0:
```
gonzalocasas:
```suggestion
        if t >= 1.0: 
```
missing whitespace

src/compas/geometry/_intersections/mesh_intersections.py
```diff
@@ -0,0 +1,102 @@
...
+    ----------
+    mesh : compas.datastructures.Mesh
+    v_keys : list of vertex indexes of a mesh
+
+    Returns
+    -------
+    list of compas.geometry.Point 
+    """
+    coordinates = [mesh.vertex_attributes(v_key, 'xyz') for v_key in v_keys]
+    return [Point(x, y, z) for x, y, z in coordinates]
```
gonzalocasas: Iterating twice over all vertices of the mesh is rather inefficient, why not return the points in a single list comprehension?
```suggestion
    return [Point(*mesh.vertex_attributes(v_key, 'xyz')) for v_key in v_keys]
```

src/compas/geometry/_primitives/line.py
```diff
@@ -56,6 +55,8 @@ class Line(Primitive):
     True
     """
 
+    __module__ = "compas.geometry"
```
gonzalocasas: I think this was giving problems, so we got rid of the `__module__` redefinition recently.

src/compas/geometry/_primitives/line.py
```diff
@@ -337,7 +338,32 @@ def transformed(self, T):
...
+
+        Returns
+        -------
+        list of: compas.geometry.Point // Point as sequence of values xyz)
 
+        Example
+        --------
+        >>> line = Line([0.0,0.0,0.0],[5.0,0.0,0.0])
+        >>> line.divide_by_count(5, True)
+        [[0.0,0.0,0.0],[1.0,0.0,0.0],[2.0,0.0,0.0],[3.0,0.0,0.0],[4.0,0.0,0.0],[5.0,0.0,0.0]]
```
gonzalocasas: I think the output is actually missing all spaces in between, so it will fail when it's executed by the testing thing.
```suggestion
        [[0.0, 0.0, 0.0], [1.0, 0.0, 0.0], [2.0, 0.0, 0.0], [3.0, 0.0, 0.0], [4.0, 0.0, 0.0], [5.0, 0.0, 0.0]]
```

src/compas/geometry/_primitives/line.py
```diff
@@ -337,7 +338,32 @@ def transformed(self, T):
...
+        ----------
+        number : integer
+            number of divisions
+        includeEnds : boolean
+            True if including start and end point in division points
+            False if not including start and end point in division points
+
+        Returns
+        -------
+        list of: compas.geometry.Point // Point as sequence of values xyz)
```
gonzalocasas: The formatting is incorrect, which I corrected below, but even then, the text is not super clear to me, "Point as sequence"? or "Sequence of points"?

```suggestion
        list of :class:`compas.geometry.Point`
            Point as sequence of values xyz
```

src/compas/geometry/_primitives/line.py
```diff
@@ -337,7 +338,32 @@ def transformed(self, T):
...
+        Example
+        --------
+        >>> line = Line([0.0,0.0,0.0],[5.0,0.0,0.0])
+        >>> line.divide_by_count(5, True)
+        [[0.0,0.0,0.0],[1.0,0.0,0.0],[2.0,0.0,0.0],[3.0,0.0,0.0],[4.0,0.0,0.0],[5.0,0.0,0.0]]
+        """
+        if include_ends:
+            return [self.point(i * float(1 / number)) for i in range(int(number)+1)]
+        else:
+            return [self.point(i * float(1 / number)) for i in range(int(number)+1) if i != 0 or i != number]
```
gonzalocasas:
```suggestion
            return [self.point(i * float(1.0 / number)) for i in range(int(number) + 1) if i != 0 or i != number]
```

src/compas/geometry/_primitives/polyline.py
```diff
@@ -54,6 +54,8 @@ class Polyline(Primitive):
     1.0
     """
 
+    __module__ = "compas.geometry"
```
gonzalocasas: Same here.

src/compas/geometry/_primitives/polyline.py
```diff
@@ -318,7 +320,107 @@ def transformed(self, T):
...
+            number of divisions
+        includeEnds : boolean
+            True if including start and ending points.
+            False if not including start and ending points.
+
+        Returns
+        -------
+        points : list of points resulting from dividing the polyline
+        """
+        points = [self.point(i * float(1 / number)) for i in range(number)]
```
gonzalocasas:
```suggestion
        points = [self.point(i * float(1.0 / number)) for i in range(number)]
```

src/compas/geometry/_transformations/extend.py
```diff
@@ -0,0 +1,97 @@
...
+    Returns
+    -------
+    extended line : tuple
+        Two points defining the offset line.
+
+    Examples
+    --------
+    >>> line = Line([0.0,0.0,0.0],[1.0,0.0,0.0])
+    >>> extended_line = extend_line(line, 1, 1)
+    Line([-1.0,0.0,0.0],[2.0,0.0,0.0])
```
gonzalocasas: Same here with the missing whitespace
```suggestion
    Line([-1.0, 0.0, 0.0], [2.0, 0.0, 0.0])
```

src/compas/geometry/splits/splits.py
```diff
@@ -0,0 +1,113 @@
...
+    sublist = []
+    for i, segment in enumerate(pairwise(polyline)):
+        sublist.append(segment.start)
+        temp_intersection = intersection_segment_plane(segment, plane)
+        if temp_intersection:
+            sublist.append(temp_intersection)
+            splitted_polylines_points.append(sublist)
+            sublist = []
+            sublist.append(temp_intersection)
+        if i == len(pairwise(polyline))-1:
```
gonzalocasas:
```suggestion
        if i == len(pairwise(polyline)) - 1:
```

src/compas/geometry/splits/splits.py
```diff
@@ -0,0 +1,113 @@
...
+    Parameters
+    ----------
+    polyline : compas.datastructures.Polyline
+    plane : compas.geometry.Plane
+
+    Returns
+    -------
+    polylines : list of compas.geometry.Polyline
+    """
+    splitted_polylines_points = []
```
gonzalocasas: minor detail but past tense of `split` is also `split`

src/compas/geometry/splits/splits.py
```diff
@@ -0,0 +1,113 @@
...
+    sublist = []
+    for i, segment in enumerate(pairwise(polyline)):
+        sublist.append(segment.start)
+        temp_intersection = intersection_segment_plane(segment, plane)
```
gonzalocasas: the `temp_` part seems superfluous, it is just a local variable and its temporal status is clear by that already.

src/compas/geometry/splits/splits.py
```diff
@@ -0,0 +1,113 @@
...
+    mesh : compas.datastructures.Mesh
+    plane : compas.geometry.Plane
+
+    Returns
+    -------
+    splitted_meshes : list of compas.datastructures.Mesh
+    """
+
+    def mesh_from_split(self, mesh, v_keys, f_keys, intersections, open=True):  # compas
+        """ Return a mesh from the positive or negative verts, faces and the intersection verts of a splitted mesh
```
gonzalocasas: There's something odd about this. Why is it a nested function at all? I would move it outside, if you don't want to expose it, that's fine, don't add it to the `__all__` and maybe also prefix it with `_` to make it clear is private, ie. `_mesh_from_split`. But then, there's the `self` here as first param that confusingly makes it look like a class-bound method.

src/compas/geometry/splits/splits.py
```diff
@@ -0,0 +1,113 @@
...
+
+    intersections = intersection_mesh_plane(mesh, plane)
+
+    if len(intersections) > 2:
+        for f_key in list(mesh.faces()):
+            split = [v_key for v_key in mesh.face_vertices(f_key) if v_key in intersections]
+            if len(split) == 2:
+                mesh.split_face(f_key, split[0], split[1])
+
+        pos_verts = []
```
gonzalocasas: being a bit picky here, `pos_verts` should be `positive_vertices`. Same for the negative. 
Characters are cheap these days ;)

src/compas/geometry/splits/splits.py
```diff
@@ -0,0 +1,113 @@
...
+        vertices = {key: mesh.vertex_coordinates(key) for key in v_keys + intersections}
+        faces = [mesh.face_vertices(f_key) for f_key in f_keys]
+        final_mesh = Mesh.from_vertices_and_faces(vertices, faces)
+        if not open:
+            final_mesh.add_face(final_mesh.vertices_on_boundary(True))
+        return final_mesh
+
+    intersections = intersection_mesh_plane(mesh, plane)
+
+    if len(intersections) > 2:
```
gonzalocasas: Reverse this condition to reduce nesting, eg: 
```
    if len(intersections) < 3:
        return None
```

gonzalocasas (changes requested):
I checked mostly for code style. I guess @brgcode would check for correctness in terms of geometry processing.
I added a bunch of comments, plus I think it would really be of use to add a few unit tests to this (and it should not be overly complex to do so).

src/compas/geometry/_primitives/line.py
```diff
@@ -337,7 +338,33 @@ def transformed(self, T):
...
+            Point as sequence of values xyz
 
+        Example
+        --------
+        >>> line = Line([0.0,0.0,0.0],[5.0,0.0,0.0])
+        >>> line.divide_by_count(5, True)
+        [[0.0, 0.0, 0.0], [1.0, 0.0, 0.0], [2.0, 0.0, 0.0], [3.0, 0.0, 0.0], [4.0, 0.0, 0.0], [5.0, 0.0, 0.0]]
+        """
+        if include_ends:
+            return [self.point(i * float(1 / number)) for i in range(int(number)+1)]
```
nikeftekhar, line 364: return [self.point(i * float(1.0 / number)) for i in range(int(number) + 1)]

src/compas/geometry/splits/splits.py
nikeftekhar: sorry, my mistake, i forgot to remove the self. Originally i created this function as a class function. Apart from that, i agree it makes sense to place it outside.

src/compas/geometry/_intersections/mesh_intersections.py
```diff
@@ -0,0 +1,96 @@
...
+        intersection = intersection_line_triangle(line, triangle)
+        if intersection:
+            return Point(intersection[0], intersection[1], intersection[2])
+
+        if len(vertex_keys) == 4:
+            triangle_2 = [vertices[2], vertices[3], vertices[0]]
+            intersection_2 = intersection_line_triangle(line, triangle_2)
+            if intersection_2:
+                return Point(intersection_2[0], intersection_2[1], intersection_2[2])
+    else:
```
nikeftekhar: indent

src/compas/geometry/_intersections/mesh_intersections.py
```diff
@@ -0,0 +1,96 @@
...
+    'mesh_vertices_to_points',
+]
+
+
+def intersection_mesh_line(mesh, line):
+    """Compute intersection between mesh faces and line. After one single intersection, stops searching for more.
+
+    Parameters
+    ----------
+    mesh : compas.datastructures.Mesh
```
beverlylytle, line 23: To make the documentation reference-able, this should read 
```
mesh : :class:`compas.datastructures.Mesh`
```
and similarly for all references in this PR.

src/compas/geometry/_intersections/mesh_intersections.py
```diff
@@ -0,0 +1,96 @@
...
+    line : compas.geometry.Line
+
+    Returns
+    -------
+    Point : compas.geometry.Point
+    """
+    for fkey in list(mesh.faces()):
```
beverlylytle, line 30: As I understand it `mesh.faces()` returns a generator, so there's no need to instantiate a list (could be very taxing on memory), so it would be more efficient to have:
```
for fkey in mesh.faces():
```

src/compas/geometry/_intersections/mesh_intersections.py
```diff
@@ -0,0 +1,96 @@
...
+        vertex_keys = mesh.face_vertices(fkey)
+        if not vertex_keys:
+            continue
+        vertices = [mesh.vertex_attributes(vkey, 'xyz') for vkey in vertex_keys]
+        if len(vertex_keys) not in (3, 4):
+            continue
```
beverlylytle, line 36: Could this be simplified to 
```
if not (vertex_keys and len(vertex_keys) in (3,4)):
    continue
vertices = [mesh.vertex_atrributes(vkey, 'xyz') for vkey in vertex_keys]
```

src/compas/geometry/_intersections/mesh_intersections.py
```diff
@@ -0,0 +1,96 @@
...
+    ----------
+    mesh : compas.datastructures.Mesh
+    plane : compas.geometry.Plane
+
+    Returns
+    -------
+    intersections: list of points as keys from mesh
+    """
+    intersections = []
+    for u, v in list(mesh.edges()):
```
beverlylytle, line 65: Again, for efficiency `for u, v in mesh.edges():`

src/compas/geometry/_intersections/mesh_intersections.py
```diff
@@ -0,0 +1,96 @@
...
+    """Compute intersection between mesh faces and line. After one single intersection, stops searching for more.
+
+    Parameters
+    ----------
+    mesh : compas.datastructures.Mesh
+    line : compas.geometry.Line
+
+    Returns
+    -------
+    Point : compas.geometry.Point
```
beverlylytle, line 28:
```
:class:`compas.geometry.Point`
```

src/compas/geometry/_intersections/mesh_intersections.py
```diff
@@ -0,0 +1,96 @@
...
+    """Calculate the keys of the points of the intersection of a mesh with a plane
+
+    Parameters
+    ----------
+    mesh : compas.datastructures.Mesh
+    plane : compas.geometry.Plane
+
+    Returns
+    -------
+    intersections: list of points as keys from mesh
```
beverlylytle, line 62:
```
list of int
    A list of keys identifying points in the mesh.
```

src/compas/geometry/_intersections/mesh_intersections.py
```diff
@@ -0,0 +1,96 @@
...
+    intersections = []
+    for u, v in list(mesh.edges()):
+        a = mesh.vertex_attributes(u, 'xyz')
+        b = mesh.vertex_attributes(v, 'xyz')
+        intersection = intersection_segment_plane((a, b), plane)
+        if not intersection:
+            continue
+        len_a_inters = length_vector(subtract_vectors(intersection, a))
+        len_a_b = length_vector(subtract_vectors(b, a))
+        t = len_a_inters / len_a_b
```
beverlylytle, line 73: Are you sure that `len_a_b` will never be zero? I mean, ideally, sure it would never be zero, but maybe do to rounding errors...?

src/compas/geometry/_intersections/mesh_intersections.py
```diff
@@ -0,0 +1,96 @@
...
+            triangle_2 = [vertices[2], vertices[3], vertices[0]]
+            intersection_2 = intersection_line_triangle(line, triangle_2)
+            if intersection_2:
+                return Point(intersection_2[0], intersection_2[1], intersection_2[2])
+                
+    return None
+
+
+def intersection_mesh_plane(mesh, plane, tol=0.0001):
+    """Calculate the keys of the points of the intersection of a mesh with a plane
```
beverlylytle, line 53: Please add a full stop at the end of the sentence.

src/compas/geometry/_primitives/line.py
```diff
@@ -337,11 +336,37 @@ def transformed(self, T):
         line.transform(T)
         return line
 
+    def divide_by_count(self, number=10, include_ends=False):
+        """Return list of points from dividing the line by specific number of divisions
```
beverlylytle, line 340: Perhaps `Return a list of points subdividing the line into the specified number of segments.`? Also, full stop, please.

src/compas/geometry/_primitives/line.py
```diff
@@ -337,11 +336,37 @@ def transformed(self, T):
         line.transform(T)
         return line
 
+    def divide_by_count(self, number=10, include_ends=False):
```
beverlylytle, line 339: I find it a little confusing that the name of the function is `divide_by_count` and the 'count' argument is named `number`. What do you think of something like `get_n_subdivision_points(self, n=10, include_ends=False)`?

src/compas/geometry/_primitives/line.py
```diff
@@ -337,11 +336,37 @@ def transformed(self, T):
...
+        Example
+        --------
+        >>> line = Line([0.0, 0.0, 0.0], [5.0 ,0.0, 0.0])
+        >>> line.divide_by_count(5, True)
+        [Point(0.000, 0.000, 0.000), Point(1.000, 0.000, 0.000), Point(2.000, 0.000, 0.000), Point(3.000, 0.000, 0.000), Point(4.000, 0.000, 0.000), Point(5.000, 0.000, 0.000)]
+        """
+        if include_ends:
+            return [self.point(i * float(1.0 / number)) for i in range(int(number) + 1)]
+        else:
+            return [self.point(i * float(1.0 / number)) for i in range(int(number)) if i != 0]
```
beverlylytle, line 364: What about the following?
```
range_ = range(int(number) + 1) if include_ends else range(1, int(number))
return [self.point(i * float(1.0 / number)) for i in range_]
```

src/compas/geometry/_primitives/polyline.py
```diff
@@ -319,11 +319,112 @@ def transformed(self, T):
         polyline.transform(T)
         return polyline
 
+    def shorten(self, start_distance=0, end_distance=0):
+        """Return a new polyline which is shorter than the original in one end side, other or both by a given distance.
```
beverlylytle, line 323: How about something like "Return a new polyline truncated by the given distances on either side."?

src/compas/geometry/_primitives/line.py
```diff
@@ -337,11 +336,37 @@ def transformed(self, T):
...
 
+    def divide_by_count(self, number=10, include_ends=False):
+        """Return list of points from dividing the line by specific number of divisions
+
+        Parameters
+        ----------
+        number : integer
+            number of divisions
+        includeEnds : boolean
+            True if including start and end point in division points
```
beverlylytle, line 347: Double ` around True and False would make the documentation look a little more consistent, that is:
```
``True`` and ``False``
```

src/compas/geometry/_primitives/line.py
```diff
@@ -337,11 +336,37 @@ def transformed(self, T):
...
         return line
 
+    def divide_by_count(self, number=10, include_ends=False):
+        """Return list of points from dividing the line by specific number of divisions
+
+        Parameters
+        ----------
+        number : integer
+            number of divisions
+        includeEnds : boolean
```
beverlylytle, line 346: Should be 
```
include_ends : bool
```

src/compas/geometry/_primitives/polyline.py
```diff
@@ -319,11 +319,112 @@ def transformed(self, T):
...
+                    points.append(line.start)
+                    if end_distance == 0:
+                        if i == len(self.lines)-1:
+                            points.append(line.end)
+                    else:
+                        if acum_length >= (self.length - end_distance):
+                            points.append(self.point(1-(end_distance/self.length)))
+                            break
+            return points
+        return self
```
beverlylytle, line 361: This is not a copy as might be indicated by the docstring.  So either, the docstring should be changed or `return self.copy()`.
Also, it's easier to read I think if you invert the if-then, that is 
```
if start_distance == 0 and end_distance == 0:
   return self.copy()
points = []
...
```

src/compas/geometry/_primitives/polyline.py
```diff
@@ -319,11 +319,112 @@ def transformed(self, T):
...
+                else:
+                    points.append(line.start)
+                    if end_distance == 0:
+                        if i == len(self.lines)-1:
+                            points.append(line.end)
+                    else:
+                        if acum_length >= (self.length - end_distance):
+                            points.append(self.point(1-(end_distance/self.length)))
+                            break
+            return points
```
beverlylytle, line 360: This is a list, not a polyline as indicated in the docstring.  So either the docstring should be changed or something along the lines of (maybe there's a better way):
```
new_polyline = self.copy()
new_polyline.points = points
return new_polyline
```

src/compas/geometry/_primitives/polyline.py
```diff
@@ -319,11 +319,112 @@ def transformed(self, T):
         polyline.transform(T)
         return polyline
 
+    def shorten(self, start_distance=0, end_distance=0):
+        """Return a new polyline which is shorter than the original in one end side, other or both by a given distance.
+
+        Parameters
+        ----------
+        start_distance : float.
+            distance to shorten from the starting point of the polyline
```
beverlylytle, line 328:
```
start_distance : float
    Length to remove from the polyline beginning from its first point.
``` 
and similarly below

src/compas/geometry/_primitives/polyline.py
```diff
@@ -319,11 +319,112 @@ def transformed(self, T):
...
+                    switch = False
+                else:
+                    points.append(line.start)
+                    if end_distance == 0:
+                        if i == len(self.lines)-1:
+                            points.append(line.end)
+                    else:
+                        if acum_length >= (self.length - end_distance):
+                            points.append(self.point(1-(end_distance/self.length)))
+                            break
```
beverlylytle, line 359: I'm a little worried about edge cases.  Let's say you take `Polyline([[0.0, 0.0, 0.0], [1.0, 0.0, 0.0]])` and `start_distance=0.5` and `end_distance=0.0`.  Then I don't think the end point `[1.0, 0.0, 0.0]` is ever appended to `points`.  And consider the case when `end_distance = 0.75`, that is when the resulting polyline should be a subsegment of a line from the original.  Or what is the expected behavior when you take 'start_distance=1.0` and `end_distance=0.0`?  Should it return `None` or a reversed polyline?  Here's a suggestion which should be tested (I probably messed up some indices by 1):
```
t_start = start_distance / self.length # let's hope the polyline isn't just a point...
t_end = end_distance / self.length
start_point = self.point(t_start)
end_point = self.point(t_end)

if t_end <= t_start:
    # what happens in this case needs to be decided
    pass

# find the first point of the polyline right of the start_point
accum_len = 0
i = 0
while accum_len < start_distance:
    line = Line(self.points[i], self.points[i+1])
    accum_len += line.length
    i += 1
first_pt_idx_right_of_start = i+1 if accum_len != start_distance else i + 2

# find the first point of the polyline to the left of the end_point
accum_len = 0
i = len(self.points) - 1
while accum_len < end_distance:
    line = Line(self.points[i-1], self.point[i])
    accum_len += line.length
    i -= 1
first_pt_idx_left_of_end = i if accum_len != end_distance else i-1

points = [start_point]
if first_pt_idx_right_of_start <= first_pt_idx_left_of_end:
    points.extend(self.points[first_pt_idx_right_of_start: first_pt_idx_left_of_end+1])
points.append(end_point)
```

src/compas/geometry/_primitives/polyline.py
```diff
@@ -319,11 +319,112 @@ def transformed(self, T):
...
+            number of points for the amount of definition of the polyline
+
+        Returns
+        -------
+        :class: 'compas.geometry.Polyline'
+            the rebuilt copy
+        """
+        rebuilt_polyline = self.copy()
+        points = [self.point(i * float(1.0 / number)) for i in range(number)]
+        points.append(self.point(1))
```
beverlylytle, line 379: `points = [self.point(i * float(1.0 / number)) for i in range(number + 1)]`

src/compas/geometry/_primitives/polyline.py
```diff
@@ -319,11 +319,112 @@ def transformed(self, T):
...
+        return rebuilt_polyline
+
+    def divide_by_count(self, number=10, include_ends=False):
+        """Divide a polyline by count. Returns list of Points from the division
+
+        Parameters
+        ----------
+        number : integer.
+            number of divisions
+        includeEnds : boolean
```
beverlylytle, line 390:
```
include_ends : bool
```

src/compas/geometry/_primitives/polyline.py
```diff
@@ -319,11 +319,112 @@ def transformed(self, T):
...
+        Returns
+        -------
+        list of :class: 'compas.geometry.Point'
+        """
+        points = [self.point(i * float(1.0 / number)) for i in range(number)]
+        if include_ends:
+            points.append(self.point(1))
+        else:
+            points.pop(0)
+        return points
```
beverlylytle, line 403:
```
range_ = range(number + 1) if include_ends else range(1, number)
return [self.point(i * float(1.0 / number)) for i in range_]

File: src/compas/geometry/_primitives/polyline.py [hunk 41]
```diff
@@ -319,11 +319,112 @@ def transformed(self, T):
...
+
+    def tween(self, polyline_two, number=50):
+        """Create an average polyline between two polylines interpolating their points
```
beverlylytle, line 406: Do you mean to create a polyline or a list of points?

src/compas/geometry/_transformations/extend.py
```diff
@@ -0,0 +1,94 @@
...
+
+    Returns
+    -------
+    extended line : tuple
+        Two points defining the extended line.
+
+    Examples
+    --------
+    >>> line = Line([0.0, 0.0, 0.0], [1.0, 0.0, 0.0])
+    >>> extended_line = extend_line(line, 1, 1)
```
beverlylytle, line 35: Should this be simply `extend_line(line, 1, 1)`?  I would think the doctests might fail as it is....

src/compas/geometry/_transformations/extend.py
```diff
@@ -0,0 +1,94 @@
...
+    'extend_polyline',
+]
+
+
+def extend_line(line, start_extension=0, end_extension=0):
+    """Extend the given line from one end or the other, or both, depending on the given values
+
+    Parameters
+    ----------
+    line : tuple
```
beverlylytle, line 20: tuple or Line or both?

src/compas/geometry/_transformations/extend.py
```diff
@@ -0,0 +1,94 @@
...
+        Two points defining the line.
+    start_extension : float
+        The extension distance at the start of the line as float.
+    end_extension : float
+        The extension distance at the end of the line as float.
+
+    Returns
+    -------
+    extended line : tuple
```
beverlylytle, line 29: tuple or Line?

src/compas/geometry/_transformations/extend.py
```diff
@@ -0,0 +1,94 @@
...
+        The extension distance at the end of the polyline as float.
+
+    Returns
+    -------
+    extended polyline : compas.geometry.Polyline(points)
+
+    Examples
+    --------
+    >>> polyline = Polyline([0.0, 0.0, 0.0], [1.0, 0.0, 0.0], [2.0, 1.0, 0.0], [3.0, 1.0, 0.0], [4.0, 0.0, 0.0], [5.0, 0.0, 0.0])
+    >>> extended_polyline = extend_polyline(polyline, 1, 1)
```
beverlylytle, line 72: Same comment as in example above.

src/compas/geometry/_transformations/extend.py
```diff
@@ -0,0 +1,94 @@
...
+    polyline : list
+        list of points defining the polyline.
+    start_extension : float
+        The extension distance at the start of the polyline as float.
+    end_extension : float
+        The extension distance at the end of the polyline as float.
+
+    Returns
+    -------
+    extended polyline : compas.geometry.Polyline(points)
```
beverlylytle, line 67:
```
:class:`compas.geometry.Polyline` 
```
the documentation doesn't need to know what you called the variable internally.

src/compas/geometry/_transformations/extend.py
```diff
@@ -0,0 +1,94 @@
...
+        line.start.transform(translation)
+    if end_extension != 0:
+        translation = calculate_translation(line, end_extension)
+        line.end.transform(translation)
+
+    return line
+
+
+def extend_polyline(polyline, start_extension=0, end_extension=0):
+    """Extend a polyline by line from the vectors on segments at extreme sides
```
beverlylytle, line 54: Maybe mention that this function returns a new polyline with 2 more points than was in the original polyline

src/compas/geometry/splits/splits.py
```diff
@@ -0,0 +1,129 @@
...
+from compas.geometry import dot_vectors
+
+__all__ = [
+    'split_polyline_plane',
+    'split_mesh_plane',
+]
+
+
+def split_polyline_plane(polyline, plane):
+    """Split a polyline by a plane, returns a list of polylines, if there has been intersections
```
beverlylytle, line 21: ...if there are...

src/compas/geometry/splits/splits.py
```diff
@@ -0,0 +1,129 @@
...
+    Returns list of the splitted polylines
```
beverlylytle, line 22: "splitted" sounds archaic to me, I think "split" might be a more standard choice.

src/compas/geometry/_primitives/polyline.py
```diff
@@ -319,11 +319,112 @@ def transformed(self, T):
...
+        ----------
+        number : integer.
+            number of divisions
+        includeEnds : boolean
+            True if including start and ending points.
+            False if not including start and ending points.
+
+        Returns
+        -------
+        list of :class: 'compas.geometry.Point'
```
beverlylytle, line 396:
```
:obj:`list` of :class:`compas.geometry.Point`
```

src/compas/geometry/_primitives/polyline.py
```diff
@@ -319,11 +319,112 @@ def transformed(self, T):
...
+        return points
+
+    def tween(self, polyline_two, number=50):
+        """Create an average polyline between two polylines interpolating their points
+
+        Parameters
+        ----------
+        polyline_two : compas.geometry.Polyline
+            polyline to create the tween polyline
+        number : number of points of the tween polyline
```
beverlylytle, line 412:
```
number : int
    The number of points of the tween polyline.

File: src/compas/geometry/_transformations/extend.py [hunk 52]
```diff
@@ -0,0 +1,94 @@
...
+
+    Parameters
+    ----------
+    polyline : list
```
beverlylytle, line 58: list or Polyline or both?

src/compas/geometry/_transformations/extend.py
```diff
@@ -0,0 +1,94 @@
...
+from __future__ import absolute_import
+from __future__ import division
+
+from compas.geometry import Translation
+from compas.geometry import Polyline
+from compas.geometry import Vector
+
+__all__ = [
+    'extend_line',
+    'extend_polyline',
```
beverlylytle, line 11: I have two comments about these two functions.  First, `extend_line` modifies the line that is given to it as an argument, while `extend_polyline` doesn't change the polyline that is given to it, and creates a new one that is extended.  I think this subtle difference in behavior could lead to some annoying bugs.  The second comment is: Why are these functions not methods of the classes `Line` and `Polyline`, resp.?

src/compas/geometry/_primitives/polyline.py
```diff
@@ -319,11 +319,112 @@ def transformed(self, T):
...
+        rebuilt_polyline.points = [Point(x, y, z) for x, y, z in points]
+        return rebuilt_polyline
+
+    def divide_by_count(self, number=10, include_ends=False):
+        """Divide a polyline by count. Returns list of Points from the division
+
+        Parameters
+        ----------
+        number : integer.
+            number of divisions
```
beverlylytle, line 389:
```
number : int
    The number of subdivisions of the polyline.

File: src/compas/geometry/_primitives/polyline.py [hunk 55]
```diff
@@ -319,11 +319,112 @@ def transformed(self, T):
...
+            the rebuilt copy
+        """
+        rebuilt_polyline = self.copy()
+        points = [self.point(i * float(1.0 / number)) for i in range(number)]
+        points.append(self.point(1))
+        rebuilt_polyline.points = [Point(x, y, z) for x, y, z in points]
+        return rebuilt_polyline
+
+    def divide_by_count(self, number=10, include_ends=False):
+        """Divide a polyline by count. Returns list of Points from the division
```
beverlylytle, line 384: Again, I find the count/number discrepancy confusing.  Whatever you choose to do with the `Line` version of this, should also be applied here.

src/compas/geometry/splits/splits.py
```diff
@@ -0,0 +1,129 @@
...
+
+    Parameters
+    ----------
+    polyline : compas.geometry.Polyline
+    plane : compas.geometry.Plane
+
+    Returns
+    -------
+    list of :class: 'compas.geometry.Polyline'
```
beverlylytle, line 31:
```
list of :class:`compas.geometry.Polyline`

File: src/compas/geometry/splits/splits.py [hunk 57]
```diff
@@ -0,0 +1,129 @@
...
+    """
+    points_from_split_polylines = []
+    sublist = []
+    for i, segment in enumerate(pairwise(polyline)):
+        sublist.append(segment.start)
```
beverlylytle, line 36: Is this code valid?  Shouldn't it be `segment[0]` rather than `segment.start`?

src/compas/geometry/splits/splits.py
```diff
@@ -0,0 +1,129 @@
...
+        intersection = intersection_segment_plane(segment, plane)
+        if intersection:
+            sublist.append(intersection)
+            points_from_split_polylines.append(sublist)
+            sublist = []
+            sublist.append(intersection)
```
beverlylytle, line 42: `sublist = [intersection]`

src/compas/geometry/splits/splits.py
```diff
@@ -0,0 +1,129 @@
...
+        if i == len(pairwise(polyline)) - 1:
```
beverlylytle, line 43: I also think that `len(pairwise(polyline))` isn't valid code.  Double check me on this, but I think ```if i == len(polyline) -2:``` would do it.

src/compas/geometry/splits/splits.py
```diff
@@ -0,0 +1,129 @@
...
+            sublist.append(segment.end)
```
beverlylytle, line 44: `segment[1]`

src/compas/geometry/splits/splits.py
```diff
@@ -0,0 +1,129 @@
...
+            points_from_split_polylines.append(sublist)
+
+    return [Polyline(sublist) for sublist in points_from_split_polylines]
+
+
+def _mesh_from_split(mesh, v_keys, f_keys, intersections, open=True):
+    """Return a mesh from the positive or negative verts, faces and the intersection verts of a splitted mesh
```
beverlylytle, line 51: I am very confused by this function.  I assume what you intend is that you have a mesh which you have intersected by a plane. Then you take all the vertices on one side of the plane (as you call them "positive verts") or the other ("negative verts") as `v_keys` along with the vertices of `intersections`, which presumably all lie on some plane which never makes an appearance in this function , and all the faces, `f_keys`, for those vertices, and create a new mesh, possibly adding a face to close the boundary.  Could you either update the docstring to explain this, if it is correct, or whatever the correct thing is?  In the case that my understanding is correct, if the mesh wasn't closed to begin with... let's imagine a cylinder with open top and bottom intersected with a plane parallel to the top/bottom... then adding this face on the boundary could create a very weird mesh, no?  Or even worse, what happens if the plane intersects the mesh in such a way that the, say, positive side is not connected.  Also, looking at this function in isolation, it's really not clear what the distinction between `v_keys` and `intersections` is, without this hypothetical plane as a backstory.  They may as well be the same parameter in the function, and the function could be called 
```
get_submesh(mesh, vertex_keys, face_keys, open=True)

File: src/compas/geometry/_intersections/mesh_intersections.py (hunk 27 above)
Comment (beverlylytle) [reply src/compas/geometry/_intersections/mesh_intersections.py:53]: This function not only returns the intersection points, but modifies the mesh structure, splitting any edge that cross the plane.  I think that should be noted in the docstring.

File: src/compas/geometry/splits/splits.py [hunk 62]
```diff
@@ -0,0 +1,129 @@
...
+    """
+
+    intersections = intersection_mesh_plane(mesh, plane)
+
+    if len(intersections) < 3:
+        return None
+
+    for f_key in list(mesh.faces()):
+        split = [v_key for v_key in mesh.face_vertices(f_key) if v_key in intersections]
+        if len(split) == 2:
```
beverlylytle, line 98: Does `split` always have length 1 or 2?

src/compas/geometry/splits/splits.py
```diff
@@ -0,0 +1,129 @@
...
+        similarity = dot_vectors(plane.normal, ori_vert_a)
+        if similarity > 0.0:
+            positive_vertices.append(v_key)
+        elif similarity < 0.0:
+            negative_vertices.append(v_key)
+
+    positive_faces = []
+    for key in positive_vertices:
+        positive_faces += mesh.vertex_faces(key)
+    positive_faces = list(set(positive_faces))
```
beverlylytle, line 117: set comprehensions ftw!
```
positive_faces = list({mesh.vertex_faces(key) for key in positive_vertices})

File: src/compas/geometry/splits/splits.py [hunk 64]
```diff
@@ -0,0 +1,129 @@
...
+
+    negative_faces = []
+    for key in negative_vertices:
+        negative_faces += mesh.vertex_faces(key)
+    negative_faces = list(set(negative_faces))
```
beverlylytle, line 122: Same as above.

src/compas/geometry/splits/splits.py
```diff
@@ -0,0 +1,129 @@
...
+
+    intersections = intersection_mesh_plane(mesh, plane)
+
+    if len(intersections) < 3:
+        return None
+
+    for f_key in list(mesh.faces()):
+        split = [v_key for v_key in mesh.face_vertices(f_key) if v_key in intersections]
+        if len(split) == 2:
+            mesh.split_face(f_key, split[0], split[1])
```
beverlylytle, line 99: It feels a little weird to me that `intersection_mesh_plane` inserts new vertices and this function inserts new edges between those new vertices.  Is there a good reason why those things don't happen together?

src/compas/geometry/splits/splits.py
```diff
@@ -0,0 +1,129 @@
...
+    for v_key in mesh.vertices():
+        if v_key in intersections:
+            continue
+        vert_a = mesh.vertex_attributes(v_key, 'xyz')
+        ori_vert_a = subtract_vectors(vert_a, plane.point)
+        similarity = dot_vectors(plane.normal, ori_vert_a)
+        if similarity > 0.0:
+            positive_vertices.append(v_key)
+        elif similarity < 0.0:
+            negative_vertices.append(v_key)
```
beverlylytle, line 112: This seems like it could be pulled off into a separate function.  Could make this easier to read.

src/compas/geometry/splits/splits.py
```diff
@@ -0,0 +1,129 @@
...
+
+def split_mesh_plane(mesh, plane, open=True):
+    """Calculate all the intersections between edges of the mesh and cutting plane,
+    and splits every mesh edge at the intersection point, if it exists.
+    Returns a list of the resulting splitted meshes.
+
+    Parameters
+    ----------
+    mesh : compas.datastructures.Mesh
+    plane : compas.geometry.Plane
```
beverlylytle, line 84: Please add a description of 'open'.

src/compas/geometry/_intersections/mesh_intersections.py
```diff
@@ -0,0 +1,96 @@
...
+
+__all__ = [
+    'intersection_mesh_line',
+    'intersection_mesh_plane',
+    'mesh_vertices_to_points',
+]
+
+
+def intersection_mesh_line(mesh, line):
+    """Compute intersection between mesh faces and line. After one single intersection, stops searching for more.
```
beverlylytle, line 19: You are very clear in the docstrings about the behavior of `intersection_mesh_line` and `intersection_mesh_plane`, but I don't like that the one fails early (returning the first point of intersection with a line) and the other gives all points of intersection.  I think the difference should be reflected in the names.

src/compas/geometry/_intersections/mesh_intersections.py
```diff
@@ -0,0 +1,96 @@
...
+        if not vertex_keys:
+            continue
+        vertices = [mesh.vertex_attributes(vkey, 'xyz') for vkey in vertex_keys]
+        if len(vertex_keys) not in (3, 4):
+            continue
+
+        triangle = [vertices[0], vertices[1], vertices[2]]
+        intersection = intersection_line_triangle(line, triangle)
+        if intersection:
+            return Point(intersection[0], intersection[1], intersection[2])
```
beverlylytle, line 41: would this work?
```
return Point(*intersection)

File: src/compas/geometry/_intersections/mesh_intersections.py [hunk 70]
```diff
@@ -0,0 +1,96 @@
...
+        if t >= 1.0:
+            t = 1 - tol
+        elif t <= 0.0:
+            t = tol
+        intersection_key = mesh.split_edge(u, v, t=t, allow_boundary=True)
+        intersections.append(intersection_key)
+    return intersections
+
+
+def mesh_vertices_to_points(mesh, v_keys):
```
beverlylytle, line 83: This function doesn't have much functionality, and I don't see much point of adding it to the library.

src/compas/geometry/_primitives/polyline.py
```diff
@@ -319,11 +319,112 @@ def transformed(self, T):
...
+
+        Parameters
+        ----------
+        polyline_two : compas.geometry.Polyline
+            polyline to create the tween polyline
+        number : number of points of the tween polyline
+
+        Returns
+        -------
+        list of :class: 'compas.geometry.Point'
```
beverlylytle, line 416:
```
list of :class:`compas.geometry.Point`

**Review by beverlylytle — CHANGES_REQUESTED** [review]
Hey Nik, nice PR!  There's a lot going in this, so I've made a lot of comments, sorry if it is overwhelming.  Really, there are three general themes to the comments.  

- The first category is about making the docstrings consistent and exact:

There should be consistency in using the same tense, capitalization, and punctuation (eg "Return blahblahblah. " vs "returns blahblahblah").  

Types of function parameters and return objects should be double checked.   Reference-able types should be included in the format required by Sphinx, eg 
```
 :class:`compas.geometry.Point`
```

Docstrings should by default include examples that can be run as tests.  I didn't explicitly ask for examples in my comments, but those should be added when possible.  

- The next category of comments is about tightening up the code.  Using one line instead of two, or naming conventions.  These are mostly about readability, and maybe are just my personal preference, but a couple of them are about optimization. 

- The last category of comments is about code correctness.  I think in one or two places some edge cases were missed, and these need to be addressed.  

Do let me know if anything is unclear.

**Comment by nperraud** [comment]
Hi everyone,

First, thank you for the great package. Would it be possible to know when this will be merged. I need the function `intersection_mesh_line` for one of my experiment.

All the best

**Comment by tomvanmele** [comment]
hi, it seems there are still quite a bit of unaddressed comments. if you resolve those, i am more than happy to merge immediately. if in the meantime you need mesh/line intersections urgently you could use `intersection_mesh_ray` of `compas_libigl`...

File: src/compas/geometry/_intersections/mesh_intersections.py (hunk 20 above)
Comment (brgcode) [reply src/compas/geometry/_intersections/mesh_intersections.py:23]: agreed

File: src/compas/geometry/_intersections/mesh_intersections.py (hunk 22 above)
Comment (tomvanmele) [reply src/compas/geometry/_intersections/mesh_intersections.py:36]: well or a condition that fails early (the `and` will require both conditions to be checked at all times)
```python
if not vertex_keys or len(vertex_keys) not in (3, 4):
    continue
```

src/compas/geometry/_intersections/mesh_intersections.py
tomvanmele: rename both to `intersections_mesh_xxx`? with returned intersections sorted based on distance in the case of `line`?

tomvanmele:
also, in the case of `intersection_mesh_plane` you could use the slicing functionality of `compas_cgal`, if you need it urgently...

jf---:
@tomvanmele this is meaningful work, but this has been sitting here for a long time...
Since the author hasn't responded to the excellent code review, I think its fair to close it.

nikeftekhar:
Hi I am the author,

This commits were result of the work I was doing at mesh 4 years ago. Due to covid, funds of mesh got shortened and I could not continue the project.
Now im doing a phd at DBT, and I would be delighted to finalise those commits and comments if you really think are a good contribution, but I have completely lost the environment and the code and everything.

If you were so kind, we could set a meeting, may be you could guide me through on how to set the working environment and everything properly and I will be delighted to finish the contribution applying all the comments that were done 4 years ago.

Best,


Nik Eftekhar Olivo



El 16.04.2024, a las 14:47, Jelle Feringa ***@***.***> escribió:



@tomvanmele<https://github.com/tomvanmele> this is meaningful work, but this has been sitting here for a long time...
Since the author hasn't responded to the excellent code review, I think its fair to close it.

gonzalocasas:
> we could set a meeting, may be you could guide me through on how to set the working environment and everything properly

@nikeftekhar yep, let's do that, feel free to poke me directly in person, or send me a calendar invite to do this

nikeftekhar:
Hi Gonzalo,

Perfect then ill try to approach you during this week or the following

Best,


Nik

El 16.04.2024, a las 17:53, Gonzalo Casas ***@***.***> escribió:



we could set a meeting, may be you could guide me through on how to set the working environment and everything properly

@nikeftekhar<https://github.com/nikeftekhar> yep, let's do that, feel free to poke me directly in person, or send me a calendar invite to do this

tomvanmele:
is this still relevant?

Issue #564, opened 2020-06-17
RobotModel.to_data() 


yck011522:
# Feature Request

As a end-user, I want to serialize a RobotModel that I created programatically so that I can load it back later from file.

## Details

I have a lot of trouble trying to pickle `RobotModel`, either using `pickle` or `jsonPickle` with no success. The decoded object doesn't function anymore.

I created a little script to check if my returned object still works:

```
def LoadToolAndCheck(json_path):
    with open(json_path, 'r') as f:
        json_str = f.read()
        tool = jsonpickle.decode(json_str, keys=True)

    print (tool.__class__)
    for l in tool.links:
        print (l.__class__, l)
```
One specific scenario is when I create this `RobotModel` object in Rhino GH, and then I deserialize it in CPython, it will not function anymore.

**Describe the solution you'd like**
I want a way to serialize RobotModel objects, either by `jsonpickle` or `pickle` or a 'to_data' type of method. I believe for the consistency with other compas object, a `to_data()` method is most appropriate.

tomvanmele:
`RobotModel` does not have a `to_data` method?

yck011522:
@brgcode  In case that is not a rhetorical exclamation 😅  . Then the answer is:

`RobotModel` does not have a `to_data` method.

yck011522:
Hello. @gonzalocasas  @brgcode  Just wonder if there are any new thoughts about the issue and if there are some quick workarounds. 

Workaround potential: I need to serialize only a model that has 1 planning group. I have a few links and a few joints, created programmatically. All have attached compas mesh as their visual mesh.

tomvanmele:
sorry forgot about this. in my opinion, `RobotModel` should definitely have `from/to_data` methods... if @gonzalocasas has the time i would prefer deferring this to him :)

gonzalocasas:
yes, it should have them. I will work on this soon in the short term.

PR #596, opened 2020-07-22
Add to/from_data to RobotModel

beverlylytle:
Address #564 

### Checklist

### DIFF
--- a/src/compas/geometry/_primitives/_primitive.py
+++ b/src/compas/geometry/_primitives/_primitive.py
@@ -2,18 +2,48 @@ from __future__ import absolute_import
 from __future__ import division
 from __future__ import print_function
 
-from compas.base import DataBaseClass
 
 __all__ = ['Primitive']
 
 
-class Primitive(DataBaseClass):
+class Primitive(object):
     """Base class for geometric primitives."""
 
     __slots__ = []
 
     def __init__(self):
-        super(Primitive, self).__init__()
+        pass
+
+    @classmethod
+    def from_data(cls, data):
+        """Construct the object from its data representation.
+        """
+        raise NotImplementedError
+
+    @property
+    def data(self):
+        """Returns the data dictionary that represents the object.
+
+        Returns
+        -------
+        dict
+            The object's data.
+        """
+        raise NotImplementedError
+
+    @data.setter
+    def data(self, data):
+        raise NotImplementedError
+
+    def to_data(self):
+        """Returns the data dictionary that represents the object.
+
+        Returns
+        -------
+        dict
+            The object's data.
+        """
+        return self.data
 
     def copy(self):
         """Makes a copy of this primitive.
--- a/src/compas/geometry/_primitives/_primitive.py
+++ b/src/compas/geometry/_primitives/_primitive.py
@@ -2,47 +2,18 @@ from __future__ import absolute_import
 from __future__ import division
 from __future__ import print_function
 
+from compas.base import DataBaseClass
+
 __all__ = ['Primitive']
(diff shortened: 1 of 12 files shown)


tomvanmele:
i would prefer to remove the slots from the ABC rather than making a separate base class. otherwise it kind of defeats the purpose. the slots definition is actually not needed...

beverlylytle:
> i would prefer to remove the slots from the ABC rather than making a separate base class. otherwise it kind of defeats the purpose. the slots definition is actually not needed...

Ok, so then implementing `to_json` and `from_json` is also desired?

tomvanmele:
yes, but perhaps those don't have to be declared as abstract and could in fact be implemented directly in the base class. if this complicates things for you, perhaps remove the abstract decorators from the json methods for now...

src/compas/geometry/_primitives/_primitive.py
```diff
@@ -2,18 +2,48 @@
 from __future__ import division
 from __future__ import print_function
 
-from compas.base import DataBaseClass
 
 __all__ = ['Primitive']
 
 
-class Primitive(DataBaseClass):
+class Primitive(object):
```
brgcode: why not derive from `Base` here?

tomvanmele (commented):
LGTM in general. haven't looked in detail at the robot stuff though...
could you perhaps add `Base` to the geometry primitive as well, instead of `object`?

beverlylytle:
> could you perhaps add `Base` to the geometry primitive as well, instead of `object`?

That is a possibility, but I was returning it to the way it is on the master branch currently.  If that were to happen, that would mean implementing `to/from_json` for all things that inherit from `Primitive` which would be beyond the scope of this PR.

tomvanmele:
fair enough

tomvanmele:
will leave to @gonzalocasas to review the robot stuff and merge

beverlylytle:
> could you perhaps add `Base` to the geometry primitive as well, instead of `object`?

Wasn't so hard as I thought...

tomvanmele:
cool!
@gonzalocasas ...

src/compas/robots/model/link.py
```diff
@@ -175,6 +352,48 @@ def __init__(self, name, type=None, visual=[], collision=[], inertial=None, **kw
         self.joints = []
         self.parent_joint = None
 
+    @property
+    def data(self):
+        return {
+            'name': self.name,
+            'type': self.type,  # !!!
```
gonzalocasas: what's the `# !!!` for? because type is a reserved keyword and we're using it anyway? ;)

gonzalocasas (approved):
LGTM! If possible, I would like to add a few simple unit tests. They don't need to be fancy at all, you can copy&paste some of the ones already existing (ones that have a full urdf as input), do a back and forth conversion `to_data` and then `from_data` and then keep the existing checks, so that you don't need to spend time on the details of what to check inside the URDF.

src/compas/robots/model/link.py
beverlylytle: oh, no, i just use '!!!' as a flag for things i want to double check (easy to search for this string in the code).  originally i wasn't sure of the type of `type`, but now that i know it's a string i can just delete my flag.  not fussed about it being a reserved word.

gonzalocasas: 😉

gonzalocasas:
image

Great!!!

beverlylytle:
I'm waiting for @yck011522's  actual use case testing to be successful before I merge.

PR #609, opened 2020-09-08
skip gh-pages deploy on pull requests

Licini:
This is to create a separate workflow file for pull requests. This file will not include the step to deploy docs to gh-pages that according to @tomvanmele , has been kept failing due to user authentication issues. It is also not necessary to re-deploy to gh-pages every time on pull requests submits.

The original full workflow will now only be triggered by merging or commits directly to Master branch.


.github/workflows/test.yml
```diff
@@ -0,0 +1,59 @@
...
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install cython --install-option='--no-cython-compile'
+      - name: Install
+        run: |
+          pip install --no-cache-dir -r requirements-dev.txt
+      - name: Build docs with sphinx
+        run: |
+          rm -rf dist/docs && invoke docs
```
gonzalocasas: If we decide to remove doc deployment from PRs, then we should also remove doc building, because this is built and discarded.

.github/workflows/build.yml
```diff
@@ -4,9 +4,6 @@ on:
   push:
     branches:
       - master
-  pull_request:
-    branches:
-      - master
 
 jobs:
   build-packages:
```
gonzalocasas: To avoid the authentication issue, I believe we could simply use the `GH_PAT` env variable instead of the `GITHUB_TOKEN`. The PAT (personal access token) needs to be stored once in secrets, but then it will use that consistently instead of trying to impersonate the one pushing (at least, that's my understanding).

gonzalocasas (commented):
Just yesterday, we refactored our doc building workflow in COMPAS FAB. We kept the build at PR, but deploy to a different directory depending on whether the push is a pull request, it is a tag release or is a push to master:

Additionally, we added the versioning thing, which would be good to add here as well (the sphinx theme only has support for it on the packages theme, but it should be trivial to bring it over to the "main" compas theme"):

https://github.com/compas-dev/compas_fab/blob/master/.github/workflows/docs.yml

If possible, I would request you to match that implementation, so that we're aligned. But if you're short on time, maybe drop the versioning for now, but keep the building at PR and deploying to a different directory so that versioning is "almost ready" for the next iteration.

tomvanmele:
no, not super urgent. let's do it properly...

Licini:
> Just yesterday, we refactored our doc building workflow in COMPAS FAB. We kept the build at PR, but deploy to a different directory depending on whether the push is a pull request, it is a tag release or is a push to master:
> 
> Additionally, we added the versioning thing, which would be good to add here as well (the sphinx theme only has support for it on the packages theme, but it should be trivial to bring it over to the "main" compas theme"):
> 
> https://github.com/compas-dev/compas_fab/blob/master/.github/workflows/docs.yml
> 
> If possible, I would request you to match that implementation, so that we're aligned. But if you're short on time, maybe drop the versioning for now, but keep the building at PR and deploying to a different directory so that versioning is "almost ready" for the next iteration.

@gonzalocasas Thanks for the review! Didn't know you can deploy to a subfolder. Will modify to align with your implementation.  I have a quick question regarding to this implementation:
Once a PR is merged or closed, will the subfolder be deleted automatically, or does this need to be handled somehow manually?

.github/workflows/test.yml
Licini: The idea was to still test if the docs can be build successfully in case changes are made on the docs part in a PR. But this might not be relevant if we go ahead to align with compas_fab to deploy to different directories, right?

gonzalocasas: exactly

gonzalocasas:
> Once a PR is merged or closed, will the subfolder be deleted automatically, or does this need to be handled somehow manually?

@Licini no, it's not deleted automatically. @beverlylytle was asking me the same thing. It would be nice, but I am not sure how exactly to do. What we could do is have a workflow that trims closed PRs from the built doc repo on a regular basis (i believe it's possible to have cron-like github action workflows that trigger automatically on an interval).

That said, the extra folder doesn't do any damage, so we can tackle that problem at a later stage.

Licini:
Just implemented the strategy from @gonzalocasas to deploy into different folders according to push type. And similar to compas_fab, the docs part are separately put into `docs.yml` Please let me know if it looks alright.
 
@brgcode Could you add a PAT(Personal Access Token) to the repo secrets so this will work? Also another thing for you to confirm is that do we want to make a new subfolder at each submit of pull request?

Licini:
> > Once a PR is merged or closed, will the subfolder be deleted automatically, or does this need to be handled somehow manually?
> 
> @Licini no, it's not deleted automatically. @beverlylytle was asking me the same thing. It would be nice, but I am not sure how exactly to do. What we could do is have a workflow that trims closed PRs from the built doc repo on a regular basis (i believe it's possible to have cron-like github action workflows that trigger automatically on an interval).
> 
> That said, the extra folder doesn't do any damage, so we can tackle that problem at a later stage.

Another option could be triggering a workflow when a PR is merged or closed. Found this thread discussing how to do that https://github.community/t/trigger-workflow-only-on-pull-request-merge/17359/7
For the part to actually delete the subfolder, only way I can think of is  to check out gh-pages branch, delete the folder and do a force push, any other good way?
And of course these are not relevant if we eventually decide it's not necessary to deploy docs at submission of PRs.

gonzalocasas:
looks good! I just added a `GH_PAT` secret and triggered the workflow to run again

gonzalocasas:
> I just added a GH_PAT secret and triggered the workflow to run again

it still complains, I guess it has to be @brgcode how adds the `GH_PAT`...

tomvanmele:
i added my PAT as GH_PAT under [secrets](https://github.com/compas-dev/compas/settings/secrets).
is that what was meant by 

> it still complains, I guess it has to be @brgcode how adds the `GH_PAT`

:)

gonzalocasas:
> i added my PAT as GH_PAT under [secrets](https://github.com/compas-dev/compas/settings/secrets).
> is that what was meant by
> 
> > it still complains, I guess it has to be @brgcode how adds the `GH_PAT`
> 
> :)

Yes. And yet, it still fails. Is there a typo somewhere that we're not seeing? I don't get what's going on

tomvanmele:
could it have to do with the level of access given to the token? i only gave it repo and workflow access...

tomvanmele:
don't really understand this PAT thing though. every explanation i find only provides info about how to make one and that it can be used as authentication in for example workflows. the secrets are supposed to make it available organisation-wide to other members with admin privileges, but how are others prevented from using it?

gonzalocasas:
> could it have to do with the level of access given to the token? i only gave it repo and workflow access...

only `repo` access should be required. I already did this for compas_fab, and my PAT only has repo access and it works. 

Now it seems it's not even finding the assigned PAT, right?

gonzalocasas:
> don't really understand this PAT thing though. every explanation i find only provides info about how to make one and that it can be used as authentication in for example workflows. the secrets are supposed to make it available organisation-wide to other members with admin privileges, but how are others prevented from using it?

you can assign them to the organization, but also just to this repo, which would be recommendable to limit usage. Nothing limits 'others' to use it, but the only way to use it is to have access to the repository already. Nobody can see the token also, so, it's not possible to manually "take it" and use it somewhere else.

tomvanmele:
> > don't really understand this PAT thing though. every explanation i find only provides info about how to make one and that it can be used as authentication in for example workflows. the secrets are supposed to make it available organisation-wide to other members with admin privileges, but how are others prevented from using it?
> 
> you can assign them to the organization, but also just to this repo, which would be recommendable to limit usage. Nothing limits 'others' to use it, but the only way to use it is to have access to the repository already. Nobody can see the token also, so, it's not possible to manually "take it" and use it somewhere else.

so but does that mean that i lend my PAT to all users with access to the repo to do certain things as configured in my PAT?

just to confirm the procedure

* made a PAT called `GH_PAT` (does it have to be this name or can it be anything?) on my account (brgcode)
* added a secret called `GH_PAT` in the `compas` secrets with the value of my PAT (the long scrambled string)

also i just read that "Secrets are not passed to workflows that are triggered by a pull request from a fork." perhaps that is the reason it doesn't work, yet...

gonzalocasas:
> so but does that mean that i lend my PAT to all users with access to the repo to do certain things as configured in my PAT?

yes, but there is only ONE way to access the PAT. And that is from a github actions workflow. So the only "users with access" that can use it, would need to submit a pull request (or push, if they have direct push access) a workflow that uses it. No other access is possible.



> just to confirm the procedure

yes, makes sense. On your account, you can name it however. The important part is to copy the scrambled thingy and paste it into the secret called `GH_PAT` of the repo.


> Secrets are not passed to workflows that are triggered by a pull request from a fork

Holy! That is the problem then. So this will work once it is merged.

tomvanmele:
okay will try to merge then.

one thing to consider though is that perhaps it is not necessary to deploy the docs during the checks of a PR, but rather only when the PR is merged, which if i understand correctly doe then not need `GH_PAT` because it can only be done by people who have the right access to begin with...

gonzalocasas:
> one thing to consider though is that perhaps it is not necessary to deploy the docs during the checks of a PR

sometimes, if the PR contains a good amount of docs, it's good it we can check the final output directly instead of having to check out, build locally and then check,

tomvanmele:
okay. fair enough...

PR #614, opened 2020-09-11
Use plugins to define rhino installable packages

gonzalocasas:
The main goal of this PR is to change the way other packages install themselves into Rhino to make it a more straightforward experience for the user. 

Currently, non-core packages (eg. `compas_fab`) need to be explicitly installed using something like `python -m compas_rhino.install -p compas_fab`. This works fine for simple packages, but as soon as packages have other dependencies, the thing gets messy. In the case of `compas_fab`, we provide a wrapped installer that takes care of both compas core and compas_fab+dependencies (`python -m compas_fab.rhino.install`) but this means extra code on each package that needs this behavior, and a confusing situation with similar but not entirely identical commands to install compas stuff.

So this PR leverages the new plugin system, defines a new `pluggable` extension point for packages to announce which packages they need and changes the install process to discover all existing plugins on that extension point and retrieve all the required rhino-installable packages.

For this purpose, the plugin system now has two types of plugins (`first_match` (default) and `collect_all`), and as an almost side note, I create a doc page to list all available extension points, otherwise it's very hard to discover them (eventually, this page generation should be automated):

image

There are already 3 examples of using this new end-point, two are in this PR: `compas_rhino` and `compas_ghpython`, instead of hard-coding the package names that need to go to Rhino, they define them using the same extension point. And for `compas_fab`, just [opened a pr](https://github.com/compas-dev/compas_fab/pull/200) to do the same.

### Checklist


gonzalocasas:
ping

src/compas_rhino/install.py
```diff
@@ -109,6 +111,38 @@ def install(version=None, packages=None):
...
+    >>> import compas.plugins
+    >>> @compas.plugins.plugin(category='install')
+    ... def installable_rhino_packages():
+    ...    return ['compas_fab']
+
+    Returns
+    -------
+    :obj:`list` of :obj:`str`
+        List of package names to make available inside Rhino.
+    """
```
tomvanmele, line 142: so in some other package you would basically write a plugin for this function that adds itself to the list of installable packages?

tomvanmele (approved):
cool, let's just try it :)

src/compas_rhino/install.py
gonzalocasas: yep, eg. compas_fab: https://github.com/compas-dev/compas_fab/pull/200

PR #623, opened 2020-10-02
Move RobotModelArtist

### LINKED ISSUE #484: Add artist for RobotModel
# Feature Request

As a user, I want artist that can visualize RobotModel.

## Details
The newly added RobotModel class is nice. But without artists that can paint it, there is nothing much I can do with it. 

In particular I'm interested in a ghpython artist implementation.

beverlylytle:
Closes #484 

### Checklist


beverlylytle:
There are some elements that are up for discussion.  I made a new class `ToolModel` in compas with a corresponding `Tool` class in compas_fab.  `ToolModel` inherits from `RobotModel` (rather than is composed of) because, as I see it, a `ToolModel` "is a" `RobotModel`.  The `Tool` class in compas_fab, similar to `Robot`, has a `tool_model` attribute along with some `AttachedCollisionMesh`es, which are created on initialization.  I also put some effort into how tools are attached and how they are transformed under `update`'s.  This could use some double checking.  I will litter the code with other comments.

src/compas/robots/base_artist/_artist.py
```diff
@@ -0,0 +1,309 @@
...
+        initial_transformation = Transformation.from_frame_to_frame(Frame.worldXY(), ee_frame)
+        # but if you attach a tool after applying some update to the robot, then it appears
+        # disconnected, being displayed at the original location of the end effector frame.
+        # I feel it shouldn't matter the order of calling attach_tool(_model) and update
+        # so I suggest the alternative, but maybe it isn't safe.
+
+        if link.collision:
+            relative_transformation = link.collision[0].current_transformation
+        else:
+            relative_transformation = link.visual[0].current_transformation
```
beverlylytle: As noted in the comment above, is this a terrible idea?

src/compas/robots/base_artist/_artist.py
```diff
@@ -0,0 +1,309 @@
...
+        This method delegates the geometry drawing to the :meth:`draw_geometry`
+        method. It transforms the geometry based on the saved initial
+        transformation from the robot model.
+
+        Parameters
+        ----------
+        link : :class:`compas.robots.Link`, optional
+            Link instance to create. Defaults to the robot model's root.
+        name : :obj:`str`, optional
+            Subdomain identifier. # !!! there's got to be a better explanation.
```
beverlylytle: Suggestions for a better docstring?

src/compas/robots/base_artist/_artist.py
```diff
@@ -0,0 +1,309 @@
...
+        for item in itertools.chain(link.visual, link.collision):
+            # Some links have only collision geometry, not visual. These meshes
+            # have not been loaded.
+            if item.native_geometry:
+                for geometry in item.native_geometry:
+                    self.transform(geometry, transformation)
+
+        for child_joint in link.joints:
+            # Recursive call
+            self.scale_link(child_joint.child_link, transformation)
```
beverlylytle: Should an attached_tool_model also be scaled?

src/compas/robots/model/robot.py
```diff
@@ -507,6 +516,27 @@ def get_base_link_name(self):
         joints = self.get_configurable_joints()
         return joints[0].parent.link
 
+    @staticmethod
+    def _get_item_meshes(item):
```
beverlylytle: This function doesn't really belong here, but where instead?

src/compas/robots/model/tool.py
```diff
@@ -0,0 +1,225 @@
...
+        in conjunction with the :meth:`to_data` method.
+
+        Parameters
+        ----------
+        data : :obj:`dict`
+            The data dictionary.
+
+        Returns
+        -------
+        :class:`Tool`
```
beverlylytle:
```suggestion
        :class:`ToolModel`
```

beverlylytle:
Finally, what's the best way to handle the simultaneous changes to compas and compas_fab?

beverlylytle:
And I think I've addressed https://github.com/compas-dev/compas_fab/issues/216.

src/compas/robots/base_artist/_artist.py
tomvanmele: is `subdomain` referring to the fact that the name has to be unique within a certain context? if that is the case i would say something like "Human-readable identifier for the link. Note that the identifier is unique within ..."

tomvanmele:
> Finally, what's the best way to handle the simultaneous changes to compas and compas_fab?

we could do a synchronized release...

gonzalocasas:
> > Finally, what's the best way to handle the simultaneous changes to compas and compas_fab?
> 
> we could do a synchronized release...

We don't need to rush to sync, because `compas_fab` defines an lower and upper bounded dependency to `compas`. The latest `compas_fab` depends on `compas>=0.15,<0.16` and the current main branch depends on `compas>=0.16.4,<0.17`. So, to be the safest, we should release this bumping the minor (eg. `0.17.0`) instead of only patching `0.16.x`

src/compas/robots/base_artist/_artist.py
```diff
@@ -0,0 +1,309 @@
...
+
+from compas.geometry import Frame
+from compas.geometry import Scale
+from compas.geometry import Transformation
+
+ABC = abc.ABCMeta('ABC', (object,), {'__slots__': ()})
+
+
+__all__ = [
+    'BaseRobotModelArtist'
```
gonzalocasas, line 17: is `AbstractRobotModelArtist` missing here? or intentionally not public?

src/compas/robots/base_artist/_artist.py
gonzalocasas: Is that `namespace` perhaps?

gonzalocasas: Or maybe `context`?  If I understand this right, it would take a something like `['ur5', 'visual', 'link5', '3']` and inject this additional context in there: `['ur5', 'visual', 'tool', 'link5', '3']`. Correct?

src/compas/robots/base_artist/_artist.py
```diff
@@ -0,0 +1,309 @@
...
+                is_visual = hasattr(item, 'get_color')
+                color = item.get_color() if is_visual else None
+
+                native_geometry = []
+                for i, mesh in enumerate(meshes):
+                    # create native geometry
+                    mesh_type = 'visual' if is_visual else 'collision'
+                    mesh_name_components = [self.model.name, mesh_type, link.name, str(i)]
+                    if name:
+                        mesh_name_components.insert(2, name)
```
gonzalocasas: When I read `insert(2, name)` I thought this was hard-coded somewhat. Then I understood that it relates to the fix structure of the list defined right above. However, it might be just better to have a tiny bit of duplication and just do:
```suggestion
                    if not name:
                        mesh_name_components = [self.model.name, mesh_type, link.name, str(i)]
                    else:
                        mesh_name_components = [self.model.name, mesh_type, name, link.name, str(i)]
```

which is more verbose, but also instantly clear.

src/compas/robots/base_artist/_artist.py
gonzalocasas: Yes, i think so. @romanarust ? @yck011522 ? what do you think?

src/compas/robots/model/robot.py
gonzalocasas: how about moving it to the `Geometry` class of the robot model `compas/robots/model/geometry.py`? (probably staying as a `staticmethod`)

src/compas_blender/artists/robotmodelartist.py
```diff
@@ -0,0 +1,50 @@
...
+            color = [1., 1., 1.]
+
+        if self.layer:
+            collection = bpy.data.collections.new(self.layer)
+            bpy.context.scene.collection.children.link(collection)
+
+        v, f = geometry.to_vertices_and_faces()
+        return compas_blender.draw_mesh(vertices=v, faces=f, name=name, color=color, centroid=False, layer=self.layer)
+
+    def redbpyraw(self, timeout=None):
```
gonzalocasas: `redbpyraw` -> `redraw`?

src/compas_ghpython/artists/robotmodelartist.py
```diff
@@ -0,0 +1,51 @@
...
+            mesh.FillHoles()
+
+        return mesh
+
+    def draw(self):
+        self.draw_visual()
+
+    @staticmethod
+    def draw_collection(collection):
+        pass
```
gonzalocasas: general question, but should this `pass` or `raise NotImplementedError`?

gonzalocasas (approved):
It's looking good! I have not yet tested it thou, but code looks fine, I only added a bunch of minor comments.

I do have have one very general question/curiosity, since the `ToolModel` derives from `RobotModel`, is it possible now to really chain robots one after the other without limits, or does this only work for that one level of 'nesting' between 1 robot model and its tool model? (This is a question only, if the answer is "No chaining, only one level", it is fine, it would be out of scope of this change anyway)

src/compas/robots/base_artist/_artist.py
beverlylytle: Intentionally not public.  I separated the class for legibility more than for functionality, putting the abstract methods in the abstract class, and the universal utilities in the base class.  But the intention is that any flavor of  RobotModelArtist would inherit from BaseRobotModelArtist.

src/compas_blender/artists/robotmodelartist.py
beverlylytle: Oopsies.

src/compas_ghpython/artists/robotmodelartist.py
beverlylytle: Yeah, I suppose an error makes more sense.

src/compas/robots/base_artist/_artist.py
beverlylytle: So... this is a terrible idea? (The thumbs up response is very confusing.)

gonzalocasas: no, I think it's fine. At least the explanation of the issue makes sense to me. Not sure if @romanarust has a different idea here...

gonzalocasas (approved):
LGTM! @romanarust do you want to comment on this before merging...?

beverlylytle:
oh, are we expecting to have chains of attached robots?  I think in theory one could manage to do this with what's in this PR but it would be ugly.  But, yeah, I had only envisioned one robot with one robot attached as a tool.

romanarust:
> LGTM! @romanarust do you want to comment on this before merging...?

I'll have a look later today.

gonzalocasas:
> oh, are we expecting to have chains of attached robots? I think in theory one could manage to do this with what's in this PR but it would be ugly. But, yeah, I had only envisioned one robot with one robot attached as a tool.

Not really expecting it. At some point, it feels like it would be better for the code base if the stuff is so general that these can be composed in arbitrary chains, but by no means, it is in scope of this PR. ;)

src/compas/robots/base_artist/_artist.py
romanarust: agree, it should be scaled.

PR #629, opened 2020-10-08
add auto release on new tags

Licini:
Create release on new tags so the latest version can be tracked by Zenodo


gonzalocasas (approved):
Nice! 🤩

Issue #631, opened 2020-10-12
Binary STL writer


beverlylytle:
# Feature Request

As a user of compas (in Rhino/Grasshopper), I want to easily export STLs so that they are small and honor the origin they came with.

## Details

**Is your feature request related to a problem? Please describe.**
Ascii STLs require the geometry of the mesh to be in the positive octant, while the binary encoding for STLs doesn't have such a restriction.  Currently the compas STL writer only writes in ascii format, and so must move a mesh's geometry if the mesh is not already in the positive octant, but in some cases, it is important to maintain the origin.  Moreover, ascii STL files are unwieldy for large meshes.

**Describe the solution you'd like**
I would like the STL writer to optionally write STLs to binary, and the associated `to_stl` methods have an added argument flag `binary` defaulting to `False` resulting in an ascii STL, whereas a value of `True` would result in a binary STL.

**Describe alternatives you've considered**
As an alternative, I've written a Grasshopper script which converts a compas mesh to a rhino object, bakes it, selects it and exports it as STL.  But it's a bit of a hassle to copy/paste this script every time I need it.

PR #634, opened 2020-10-14
Add binary STL writer

beverlylytle:
Addresses #631 

### Checklist


gonzalocasas (approved):
Looks good! I'm surprised the binary protocol is so simple.
One thing I would really suggest here is to add a unit test. It can be a trivial one simply writes to a file and the file is compared against a known version of it.

tomvanmele:
well or perhaps we can take a simple triangulated shape (box or sphere) and convert it from one file format to another and intermediately always back to the mesh data structure. without any additional attributes those intermediate meshes should always be identical to the original one...

gonzalocasas:
> well or perhaps we can take a simple triangulated shape (box or sphere) and convert it from one file format to another and intermediately always back to the mesh data structure. without any additional attributes those intermediate meshes should always be identical to the original one...

true

PR #638, opened 2020-10-18
Mesh thicken fix

### LINKED ISSUE #637: mesh_thicken() doesnt close all boundaries if the initial mesh has more than one boundary
**Describe the bug**
in the function `mesh_thicken()` , if the initial mesh forms a loop, it will return two offset mesh but only one side of boundary is closed. Think this is because the initial mesh has more than one boundary and the function doesnt take this into consiteration. 

**To Reproduce**

mesh = mesh_thicken(mesh, 5)
artist.draw()

**Screenshots**
image

wenqian157:
Fixes #637 

### Checklist

### DIFF
--- a/src/compas/datastructures/mesh/offset.py
+++ b/src/compas/datastructures/mesh/offset.py
@@ -84,7 +84,10 @@ def mesh_thicken(mesh, thickness=1.0, cls=None):
 
     # close boundaries
     n = thickened_mesh.number_of_vertices() / 2
-    for u, v in list(thickened_mesh.edges_on_boundary()):
+
+    edges_on_boundary = [edge for boundary in list(thickened_mesh.edges_on_boundaries()) for edge in boundary]
+
+    for u, v in edges_on_boundary:
         if u < n and v < n:
             thickened_mesh.add_face([u, v, v + n, u + n])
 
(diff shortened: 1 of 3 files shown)


src/compas/datastructures/mesh/offset.py
```diff
@@ -84,7 +84,10 @@ def mesh_thicken(mesh, thickness=1.0, cls=None):
 
     # close boundaries
     n = thickened_mesh.number_of_vertices() / 2
-    for u, v in list(thickened_mesh.edges_on_boundary()):
+
+    edges_on_boundary = [edge for boundary in list(thickened_mesh.edges_on_boundaries()) for edge in boundary]
```
tomvanmele, line 88: this seems wrong. will this not create one face for all boundaries? should it not create a face per boundary?

wenqian157: yes this creates a face per edge on the boundary. but if a mesh has more than one closed boundary, it will just create faces for one boundary and ignore the rest.
image

tomvanmele: maybe i misunderstand, but it seems to me that you are making one flat list for all edges on all boundaries. and then you add a face for those edges.

which seems not correct...

wenqian157: yes that is what I was doing. but isn't what we want? to add a face to each of these edges(blue and red), so the thickened mesh can be closed? the test result looks correct to me. There are no overlapped faces here if that is what you are concerned. While using the previous version the red loop will be missing.
image

tomvanmele: ha sorry, that indeed means i misunderstood...

PR #640, opened 2020-10-20
add clean up workflow

Licini:
Adding documentation clean up workflow triggered when a PR is merged or closed:

1. The workflow will checkout the gh-pages. 
2. Delete the folder corresponding to the PR that is merged/closed.
3. Then push the change back to gh-pages.

I've tested this on another repo. Should work as expected.

### DIFF
--- a/.github/workflows/cleanup.yml
+++ b/.github/workflows/cleanup.yml
@@ -1,0 +1,37 @@
+name: cleanup
+
+# triggered whenever a PR submitted to master is merged or closed
+on:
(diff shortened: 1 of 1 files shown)


Licini:
docs fail because of the GH_PAT thing again 😂, I think it's because I'm submitting from a branch of my forked repo.

tomvanmele:
yes it is indeed a bit silly that this doesn't work that way, because it also means that it will be pointless for deploying docs for review based on PRs...

PR #642, opened 2020-10-22
Update compas_rhino artists

tomvanmele:
adding draw_collection to box artist

### Checklist


tomvanmele:
i just updated `BoxArtist` for now. perhaps you can review this one first until we fully agree on the format for the shape artists, and then i roll this out for all others on this PR...

tomvanmele:
just a few points for discussion:

* should the faces of the boxes in the collection be grouped into joined meshes?
* should the collection function also provide the option to show/hide other components than just the faces

gonzalocasas (approved):
Please add a line to `CHANGELOG`.
Otherwise it LGTM!

About your questions: boxes are always closed, right? so I'd tend to assume it would display it as one mesh rather than individual faces, but I realize the usage and assumptions on this point changes a lot between teams, so maybe add the `join_faces` option like the `MeshArtist` has. 

Regarding collection functions, having functional parity between single entity and collection would be nice (ie. the same `show_*` params on both ideally)

tomvanmele:
the reason behind the simplified functionality for the `draw_collection` function would be that the function is there for convenience and is meant to draw many objects fast, not necessarily withe all possible options, just the defaults. if more control is desired, the user can always use the `draw` function with artists for the individual boxes... what do you think @gonzalocasas ? otherwise, i am not sure there is much point to the collection variant.

gonzalocasas:
> the reason behind the simplified functionality for the `draw_collection` function would be that the function is there for convenience and is meant to draw many objects fast, not necessarily withe all possible options, just the defaults. if more control is desired, the user can always use the `draw` function with artists for the individual boxes... what do you think @gonzalocasas ? otherwise, i am not sure there is much point to the collection variant.

Ah! Understood. I personally have never used these `_collection` variants, perhaps out of laziness or ignorance thou. If you want to try it as it is now, sounds good to me. At some point, we could gauge how much usage they get and if not much, we could get rid of them.

src/compas_rhino/artists/capsuleartist.py
```diff
@@ -39,10 +70,19 @@ def draw(self, u=10, v=10):
...
-        points = [{'pos': point, 'color': self.color} for point in vertices]
-        polygons = [{'points': [vertices[index] for index in face], 'color': self.color} for face in faces]
-        guids = compas_rhino.draw_points(points, layer=self.layer, clear=False, redraw=False)
-        guids += compas_rhino.draw_faces(polygons, layer=self.layer, clear=False, redraw=False)
+        guids = []
+        if show_vertices:
+            points = [{'pos': point, 'color': self.color} for point in vertices]
+            guids += compas_rhino.draw_points(points, layer=self.layer, clear=False, redraw=False)
+        if show_edges:
+            raise NotImplementedError
```
tomvanmele: i have left this open for now. the capsule doesn't actually have any edges. they could be generated to be able to visualize the wireframe. in general, the question is what the most relevant information is. i would argue that even showing the vertices is questionable...

src/compas/geometry/shapes/capsule.py
```diff
@@ -86,7 +86,7 @@ def line(self):
 
     @line.setter
     def line(self, line):
-        self._line = line
+        self._line = Line(*line)
```
tomvanmele, line 89: while fixing this, i noticed that the capsule has a very different logic than for example the box. it doesn't have a local frame and is defined directly through a line defined by two points. perhaps that should become an alternative constructor. the default, like with the box, should accept a frame and the parameters that define the capsule in that frame. ideally this is the default for all shapes: a frame and parameters. alternative constructors allow to use direct geometrical inputs, but would then have to compute a frame from those internally...

@gonzalocasas @beverlylytle @Licini

src/compas_rhino/artists/boxartist.py
```diff
@@ -18,28 +24,152 @@ class BoxArtist(ShapeArtist):
...
+                guid = compas_rhino.draw_mesh(vertices, faces, layer=self.layer, name=self.name, color=self.color, disjoint=True)
+                guids.append(guid)
+            else:
+                polygons = [{'points': [vertices[index] for index in face], 'color': self.color, 'name': self.name} for face in faces]
+                guids += compas_rhino.draw_faces(polygons, layer=self.layer, clear=False, redraw=False)
         self._guids = guids
         return guids
 
+    @staticmethod
+    def draw_collection(collection,
```
tomvanmele: the collection methods offer very little advantage over the individual draw methods. since static methods are somewhat frowned upon anyway, perhaps the collection methods are not really necessary...

src/compas/geometry/shapes/cone.py
```diff
@@ -96,7 +96,7 @@ def circle(self):
 
     @circle.setter
     def circle(self, circle):
-        self._circle = circle
+        self._circle = Circle(*circle)
```
tomvanmele: while fixing this bug, i noticed that the cone does not have a bottom face. should it?

src/compas_rhino/artists/boxartist.py
beverlylytle: i tend to agree.  it seems that really the only added functionality of `draw_collection` is the grouping, and it seems like that function should be pulled out, perhaps as a static method to the BaseArtist or maybe somewhere else.  it seems reasonable that one may want to group objects of mixed type, ie a box and a mesh or two.

src/compas_rhino/artists/capsuleartist.py
beverlylytle: Leaving this as an exception permanently results in kind of a strange user experience though.  I would tend toward letting the user decide if he wants to see edges and vertices, even if it doesn't make sense.  I find pretty senselessness preferable to runtime errors.  But if you do choose to leave it this way, please mention in the documentation that it's not implemented.

src/compas/geometry/__init__.py
```diff
@@ -36,9 +37,59 @@
     Vector
 
 
+The following representations of primitives can be used
+interchangeably as input in methods and functions.
+The represenations using native Python objects also correspond to the required input
```
beverlylytle:
```suggestion
The representations using native Python objects also correspond to the required input
```

src/compas/geometry/shapes/capsule.py
gonzalocasas: the suggestion makes sense to me

tomvanmele: will submit a proposal for this via a different PR

tomvanmele:
@gonzalocasas this thing is final now, so please let me know if it can be merged...

PR #645, opened 2020-10-26
Simplified circle constructor

tomvanmele:

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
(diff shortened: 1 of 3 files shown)


tomvanmele:
This is not yet fully implemented since i have not yet made the necessary changes to the related algorithms. Just wanted to have the discussion first. The same would also apply to the ellipse. One could even question if it makes sense to provide a normal. By just providing a normal the orientation of the circle and the ellipse is completely random. Perhaps it is better to define entirely in 2D and change to a spatial geometry through transformations. Furthermore, like with shapes, one could argue that the default constructor should be entirely parametric with a frame as optional parameter. Geometry-based definitions then move to the alternative constructors...

PR #651, opened 2020-11-04
closes #649: remove usage of ABCMeta to prevent performance issues on…

### LINKED ISSUE #649: Performance issue for create object in RhinoPython
**Describe the bug**
Creating object is very slow in RhinoPython. There seems to be a memory leak.
compas.version = '0.17.1' pulled from source e5b1e23484f6951c1c2013dfa58ae16349ddfdd4

**To Reproduce**
This function test3() creates 3 lines, the test runs this 100 times:
```

def test3():
    line1 = Line([-100, 0, 0], [100, 0, 0])
    line2 = Line([0, -100, 0], [0, 100, 0])
    line3 = Line([0, -100, 10], [0, 100, 10])

if __name__ == "__main__":
    import datetime

    first_time = datetime.datetime.now()
    for i in range(100):
        test3()
    later_time = datetime.datetime.now()
    difference = later_time - first_time
    print("Time took: %s" % difference)
```
**Running this is cPython 3.7** 
Time took: 0:00:00.004987

**Running this in Rhino (fresh after ResetScriptEngine)**
Time took: 0:00:00.051000
**Subsequent runs:**
Time took: 0:00:00.319000
Time took: 0:00:00.490000
Time took: 0:00:01.155000
Time took: 0:00:01.173000
Time took: 0:00:01.154000
Time took: 0:00:02.093000
Time took: 0:00:02.136000
Time took: 0:00:02.133000
Time took: 0:00:02.648000
you get the idea, it keeps going up... until Rhino never returns (aka crash)

gonzalocasas:
… IronPython

### Checklist

### DIFF
--- a/src/compas/base.py
+++ b/src/compas/base.py
@@ -1,21 +1,28 @@
+"""
+If you ever feel tempted to use ABCMeta in your code: don't, just DON'T.
+Assigning __metaclass__ = ABCMeta to a class causes a severe memory leak/performance
+degradation on IronPython 2.7.
+
+See these issues for more details:
+ - https://github.com/compas-dev/compas/issues/562
+ - https://github.com/compas-dev/compas/issues/649
+"""
 from __future__ import print_function
 from __future__ import absolute_import
 from __future__ import division
 
-import abc
 import json
 from uuid import uuid4
 
 from compas.utilities import DataEncoder
 from compas.utilities import DataDecoder
-from compas.utilities import abstractclassmethod
-
-ABC = abc.ABCMeta('ABC', (object,), {'__slots__': ()})
 
 
 __all__ = [
     'Base',
 ]
+import abc
+ABC = abc.ABCMeta('ABC', (object,), {'__slots__': ()})
 
 
 class Base(ABC):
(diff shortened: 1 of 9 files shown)


src/compas/base.py
```diff
@@ -1,21 +1,28 @@
...
 
 
 __all__ = [
     'Base',
 ]
+import abc
+ABC = abc.ABCMeta('ABC', (object,), {'__slots__': ()})
 
 
 class Base(ABC):
```
beverlylytle: Don't you think the other classes will be jealous that `Base` still inherits from `ABC`?

beverlylytle:
Should there also be a PR on compas_fab to remove ABCMeta from the backend features?

src/compas/base.py
tomvanmele: they were indeed

gonzalocasas:
Thank you both for the quick review!
@beverlylytle yes, we should do the same for fab.
@brgcode could we please push a new patch release 0.17.2 with this fix?
Thx!

tomvanmele:
done...

PR #674, opened 2020-11-23
json convenience functions

tomvanmele:

### Checklist

### DIFF
--- a/src/compas/_json.py
+++ b/src/compas/_json.py
@@ -1,0 +1,135 @@
+from __future__ import print_function
+from __future__ import absolute_import
+from __future__ import division
+
+import json
+from compas.utilities import DataEncoder
+from compas.utilities import DataDecoder
+
+
+__all__ = [
+    'json_dump',
+    'json_dumps',
+    'json_load',
+    'json_loads'
+]
+
+
+def json_dump(data, fp):
+    """Write a collection of COMPAS object data to a JSON file.
+
+    Parameters
+    ----------
+    data : list of :class:`compas.base.Base`
+        A list of COMPAS objects.
+    fp : file-like object or path
+        A writeable file-like object or the path to a file.
+
+    Returns
+    -------
+    None
+
+    Examples
+    --------
+    >>> import compas
+    >>> from compas.geometry import Point, Vector
+    >>> data1 = [Point(0, 0, 0), Vector(0, 0, 0)]
+    >>> compas.json_dump(data1, 'data.json')
+    >>> data2 = compas.json_load('data.json')
+    >>> data1 == data2
+    True
+    """
+    if hasattr(fp, 'write'):
+        return json.dump(data, fp, cls=DataEncoder)
+    with open(fp, 'w') as fp:
+        return json.dump(data, fp, cls=DataEncoder)
+
+
+def json_dumps(data):
+    """Write a collection of COMPAS objects to a JSON string.
+
+    Parameters
+    ----------
+    data : list of :class:`compas.base.Base`
+        A list of COMPAS objects.
+
+    Returns
+    -------
+    str
+
+    Examples
+    --------
+    >>> import compas
+    >>> from compas.geometry import Point, Vector
+    >>> data1 = [Point(0, 0, 0), Vector(0, 0, 0)]
+    >>> s = compas.json_dumps(data1)
+    >>> data2 compas.json_loads(s)
+    >>> data1 == data2
+    True
+    """
+    return json.dumps(data, cls=DataEncoder)
+
+
+def json_load(fp):
+    """Read COMPAS object data from a JSON file.
+
+    Parameters
+    ----------
+    fp : file-like object or path
+        A writeable file-like object or the path to a file.
+
+    Returns
+    -------
+    data
+        The (COMPAS) data contained in the file.
+
+    Examples
+    --------
+    >>> import compas
+    >>> from compas.geometry import Point, Vector
+    >>> data1 = [Point(0, 0, 0), Vector(0, 0, 0)]
+    >>> compas.json_dump(data1, 'data.json')
+    >>> data2 = compas.json_load('data.json')
+    >>> data1 == data2
+    True
+    """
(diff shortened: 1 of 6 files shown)


src/compas/_json.py
```diff
@@ -0,0 +1,135 @@
...
+    'json_loads'
+]
+
+
+def json_dump(data, fp):
+    """Write a collection of COMPAS object data to a JSON file.
+
+    Parameters
+    ----------
+    data : list of :class:`compas.base.Base`
```
gonzalocasas: Are really only `list`s supported? isn't it also valid to pass a single instance of a class implementing anything that can be encoded/decoded by the `DataEncoder/DataDecoder` pair and it will work?

src/compas/_json.py
```diff
@@ -0,0 +1,135 @@
...
+
+
+# ==============================================================================
+# Main
+# ==============================================================================
+
+if __name__ == '__main__':
+    import doctest
+
+    doctest.testmod(globs=globals())
```
gonzalocasas, line 137: I would really really add unit testing for this, the tests in the doctests are not covering a lot of edge cases, and this is one of those features that really should always work correctly.

gonzalocasas (commented):
Generally lgtm, except one question about the input and the need for some unit tests (besides the doctests)

src/compas/_json.py
tomvanmele: yes indeed. will update...

src/compas/_json.py
```diff
@@ -0,0 +1,135 @@
...
+    >>> from compas.geometry import Point, Vector
+    >>> data1 = [Point(0, 0, 0), Vector(0, 0, 0)]
+    >>> compas.json_dump(data1, 'data.json')
+    >>> data2 = compas.json_load('data.json')
+    >>> data1 == data2
+    True
+    """
+    if hasattr(fp, 'write'):
+        return json.dump(data, fp, cls=DataEncoder)
+    with open(fp, 'w') as fp:
```
beverlylytle, line 45: If the plan is to fix up all of the file readers so that they take file-like objects and paths, I think these lines of code will be used quite a few times.  Maybe there should be a utility function which does this.  Some sort of context manager like 
```
class open_file(object):
    def __init__(self, fp, *args, **kwargs):
        if hasattr(fp, 'write'):
            self.f = fp
        else:
            self.f = open(fp, *args, **kwargs)

    def __enter__(self):
        return self.f

    def __close__(self):
        self.f.close()
        # but maybe you don't always want to close file-like objects...
```
And I suppose this could be generalized to handle reading as well.

src/compas/_json.py
Licini: would be nice to support dictionaries as well something like {obj_a: ..., obj_b: ...}

tomvanmele: just to be clear. you can in fact pass it anything you want. it is just the documentation that suggests the input is limited to lists...

tests/compas/test_json.py
```diff
@@ -0,0 +1,59 @@
...
+    assert s == '[[1, 2, 3], [1.0, 2.0, 3.0], 1.0, 1, NaN, Infinity]'
+
+
+def test_dumps_primitive():
+    d1 = Point(0, 0, 0)
+    s1 = compas.json_dumps(d1)
+    d2 = Point(np.float64(0.0), np.float64(0.0), np.float64(0.0))
+    s2 = compas.json_dumps(d2)
+    assert s1 == '{"dtype": "compas.geometry/Point", "value": [0.0, 0.0, 0.0]}'
+    assert s2 == '{"dtype": "compas.geometry/Point", "value": [0.0, 0.0, 0.0]}'
```
gonzalocasas: These tests fail on ironpython due to the numpy import. One option is to wrap them around a big `if` to skip the numpy parts if on ipy:
```suggestion
import compas

from compas.geometry import Point, Vector, Frame
from compas.geometry import Box
from compas.geometry import Transformation
from compas.datastructures import Mesh


def test_dumps_native():
    data = [[], (), {}, '', 1, 1.0, True, None, float('inf')]
    s = compas.json_dumps(data)
    assert s == '[[], [], {}, "", 1, 1.0, true, null, Infinity]'


if compas.IPY:
    import numpy as np

        def test_dumps_numpy():
        data = [np.array([1, 2, 3]), np.array([1.0, 2.0, 3.0]), np.float64(1.0), np.int32(1), np.nan, np.inf]
        s = compas.json_dumps(data)
        assert s == '[[1, 2, 3], [1.0, 2.0, 3.0], 1.0, 1, NaN, Infinity]'

    def test_dumps_primitive():
        d1 = Point(0, 0, 0)
        s1 = compas.json_dumps(d1)
        d2 = Point(np.float64(0.0), np.float64(0.0), np.float64(0.0))
        s2 = compas.json_dumps(d2)
        assert s1 == '{"dtype": "compas.geometry/Point", "value": [0.0, 0.0, 0.0]}'
        assert s2 == '{"dtype": "compas.geometry/Point", "value": [0.0, 0.0, 0.0]}'
```

gonzalocasas (changes requested):

tomvanmele:
@gonzalocasas @Licini @beverlylytle tests pass now. can we merge this? or do we need more?

tests/compas/test_json.py
```diff
@@ -0,0 +1,71 @@
...
+        after = compas.json_loads(s)
+        assert after == [[1, 2, 3], [1.0, 2.0, 3.0], 1.0, 1]
+
+
+def test_json_primitive():
+    before = Point(0, 0, 0)
+    s = compas.json_dumps(before)
+    after = compas.json_loads(s)
+    if not compas.IPY:
+        assert type(before) == type(after)
```
beverlylytle: I'm not a fan of skipping tests based on the environment they are run on.  With the numpy situation, it's clear that that has to happen, but here, it's not so clear.  What about `assert before.__class__ == after.__class__` or ` assert isinstance(after, before.__class__)`?

PR #676, opened 2020-11-24
deal with str keys after json serialization

Licini:
After json-serialization, All keys of dicts will be converted from `int` to `str` ,  this is to deal with this situation in data setter of `Halfface` class

### Checklist

### DIFF
--- a/src/compas/datastructures/volmesh/core/halfface.py
+++ b/src/compas/datastructures/volmesh/core/halfface.py
@@ -268,10 +268,7 @@ class HalfFace(Datastructure):
             for u in cell[c]:
                 for v in cell[c][u]:
                     f = cell[c][u][v]
-                    if f in halfface:
-                        faces.append(halfface[f])
-                    else:
-                        faces.append(halfface[str(f)])
+                    faces.append(self._halfface[f])
             self.add_cell(faces, ckey=int(c), attr_dict=attr)
 
         for e in edge_data:
--- a/src/compas/datastructures/volmesh/core/halfface.py
+++ b/src/compas/datastructures/volmesh/core/halfface.py
@@ -268,7 +268,10 @@ class HalfFace(Datastructure):
             for u in cell[c]:
                 for v in cell[c][u]:
                     f = cell[c][u][v]
-                    faces.append(halfface[f])
+                    if f in halfface:
+                        faces.append(halfface[f])
+                    else:
+                        faces.append(halfface[str(f)])
             self.add_cell(faces, ckey=int(c), attr_dict=attr)
 
         for e in edge_data:
--- a/src/compas/datastructures/volmesh/core/halfface.py
+++ b/src/compas/datastructures/volmesh/core/halfface.py
@@ -264,12 +264,15 @@ class HalfFace(Datastructure):
 
         for c in cell:
             attr = cell_data.get(c) or {}
-            faces = []
+            fkeys = set()
             for u in cell[c]:
                 for v in cell[c][u]:
                     f = cell[c][u][v]
-                    faces.append(self._halfface[f])
-            self.add_cell(faces, ckey=int(c), attr_dict=attr)
+                    fkeys.add(f)
+            fkeys = list(fkeys)
+            faces = [self._halfface[fkey] for fkey in fkeys]
+
+            self.add_cell(faces, ckey=int(c), attr_dict=attr, fkeys=fkeys)
 
         for e in edge_data:
             self._edge_data[literal_eval(e)] = edge_data[e] or {}
@@ -476,7 +479,7 @@ class HalfFace(Datastructure):
                 self._plane[w][v][u] = None
         return fkey
 
-    def add_cell(self, faces, ckey=None, attr_dict=None, **kwattr):
+    def add_cell(self, faces, ckey=None, attr_dict=None, fkeys=None, **kwattr):
         """Add a cell to the volmesh object.
 
         Parameters
@@ -525,9 +528,13 @@ class HalfFace(Datastructure):
         self._cell[ckey] = {}
         for name, value in attr.items():
             self.cell_attribute(ckey, name, value)
-        for vertices in faces:
-            fkey = self.add_halfface(vertices)
-            vertices = self.halfface_vertices(fkey)
+
+        for fkey_index, vertices in enumerate(faces):
+            if fkeys:
+                fkey = fkeys[fkey_index]
+            else:
+                fkey = self.add_halfface(vertices)
+                vertices = self.halfface_vertices(fkey)
             for i in range(-2, len(vertices) - 2):
                 u = vertices[i]
                 v = vertices[i + 1]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 ### Changed
+* Fixed bug in `compas.datastructures.Halfface.data(setter)`  in reading json serialized data
 
 ### Removed
 
--- a/tests/compas/datastructures/test_halfface.py
+++ b/tests/compas/datastructures/test_halfface.py
@@ -1,0 +1,17 @@
+import compas
+from compas.datastructures import VolMesh
+from compas.utilities import DataEncoder
+import json
+
+
+
+def test_data_setter():
+
+    v1 = VolMesh.from_obj(compas.get("boxes.obj"))
+    data1 = v1.to_data()
+    json_str = json.dumps(data1, cls=DataEncoder)
+
+    v2 = VolMesh.from_data(json.loads(json_str))
+    data2 = v2.to_data()
+
+    assert data1 == data2


gonzalocasas (changes requested):
I think this would benefit from a couple of test cases that exercise the two newly added code paths. Other than that, lgtm 👍

src/compas/datastructures/volmesh/core/halfface.py
```diff
@@ -268,7 +268,10 @@ def data(self, data):
             for u in cell[c]:
                 for v in cell[c][u]:
                     f = cell[c][u][v]
-                    faces.append(halfface[f])
+                    if f in halfface:
+                        faces.append(halfface[f])
+                    else:
+                        faces.append(halfface[str(f)])
```
tomvanmele: the if statement doesn't make sense to me. either it is supposed to be a string by construction or it is not. `f` is not a key in `cell`. therefore it will still be an `int` and should always be converted to `str` before using it in `halfface`. alternatively we could refer to the already processed `self._halfface` and just keep it as `int`

Licini: I agree this is not a very good solution there, I m thinking about some better way.... Not sure if I understand all of it. But do you think we can just force all of the keys to be str? because once it's encoded to json and parsed back, it becomes very difficult to determine a key used to be a str or something else. Currently to_data() function produce a lot of int keys. This create a lot of ambiguity on how to parse them back from a json string.

Licini: Or if a dictionary only suppose to have integer keys, we should consider use list instead

src/compas/datastructures/volmesh/core/halfface.py
```diff
@@ -268,7 +268,7 @@ def data(self, data):
             for u in cell[c]:
                 for v in cell[c][u]:
                     f = cell[c][u][v]
-                    faces.append(halfface[f])
+                    faces.append(self._halfface[f])
```
Licini: @brgcode This would be better no?

Licini:
> I think this would benefit from a couple of test cases that exercise the two newly added code paths. Other than that, lgtm 👍

Sounds good, gonna add a unit test before we can merge

src/compas/datastructures/volmesh/core/halfface.py
tomvanmele: yes that should be okay

src/compas/datastructures/volmesh/core/halfface.py
tomvanmele: we should just force all of the dict keys to `int` in the setter, because only `int` identifiers are supported by the data structure. using lists is not an option because then data management is a nightmare...

there should not be an issue. the mesh works in exactly the same way and there has never been a problem.

src/compas/datastructures/volmesh/core/halfface.py
```diff
@@ -476,7 +479,7 @@ def add_halfface(self, vertices, fkey=None, attr_dict=None, **kwattr):
                 self._plane[w][v][u] = None
         return fkey
 
-    def add_cell(self, faces, ckey=None, attr_dict=None, **kwattr):
+    def add_cell(self, faces, ckey=None, attr_dict=None, fkeys=None, **kwattr):
```
Licini, line 482: Hey @tomvanmele Could you take a look at this change here: `add_cell` used to append new halffaces to the class, however in the case of `data setter` , the halffaces are already added initially, calling `add_cell` results in having duplicated halffaces.  So there needs to be an optional argument to supply an existing fkey instead of generate a new one.

brgcode:
in my opinion, things can be much simpler
```python
    @property
    def data(self):
        """dict: A data dict representing the volmesh data structure for serialisation.
        """
        # edge_data = {}
        # for e in self._edge_data:
        #     edge_data[repr(e)] = self._edge_data[e]
        data = {
            'attributes': self.attributes,
            'dva': self.default_vertex_attributes,
            'dea': self.default_edge_attributes,
            'dfa': self.default_face_attributes,
            'dca': self.default_cell_attributes,
            'vertex': self._vertex,
            'halfface': self._halfface,
            'cell': self._cell,
            'plane': self._plane,
            'edge_data': self.edge_data,
            'face_data': self._face_data,
            'cell_data': self._cell_data,
            'max_vertex': self._max_vertex,
            'max_face': self._max_face,
            'max_cell': self._max_cell}
        return data

    @data.setter
    def data(self, data):
        attributes = data.get('attributes') or {}
        dva = data.get('dva') or {}
        dea = data.get('dea') or {}
        dfa = data.get('dfa') or {}
        dca = data.get('dca') or {}
        vertex = data.get('vertex') or {}
        halfface = data.get('halfface') or {}
        cell = data.get('cell') or {}
        # plane = data.get('plane') or {}
        edge_data = data.get('edge_data') or {}
        face_data = data.get('face_data') or {}
        cell_data = data.get('cell_data') or {}
        max_vertex = data.get('max_vertex', -1)
        max_face = data.get('max_face', -1)
        max_cell = data.get('max_cell', -1)

        if not vertex or not halfface or not cell:
            return

        self.attributes.update(attributes)
        self.default_vertex_attributes.update(dva)
        self.default_edge_attributes.update(dea)
        self.default_face_attributes.update(dfa)
        self.default_cell_attributes.update(dca)

        self._vertex = {}
        self._halfface = {}
        self._cell = {}
        self._plane = {}
        self._edge_data = {}
        self._face_data = {}
        self._cell_data = {}

        for v in vertex:
            attr = vertex[v] or {}
            self.add_vertex(int(v), attr_dict=attr)

        # for f in halfface:
        #     attr = face_data.get(f) or {}
        #     self.add_halfface(halfface[f], fkey=int(f), attr_dict=attr)

        for c in cell:
            attr = cell_data.get(c) or {}
            faces = []
            for u in cell[c]:
                for v in cell[c][u]:
                    f = cell[c][u][v]
                    faces.append(halfface[f])
            self.add_cell(faces, ckey=int(c), attr_dict=attr)

        for e in edge_data:
            self._edge_data[e] = edge_data[e] or {}

        for f in face_data:
            self._face_data[f] = face_data[f] or {}

        self._max_vertex = max_vertex
        self._max_face = max_face
        self._max_cell = max_cell
```

src/compas/datastructures/volmesh/core/halfface.py
tomvanmele: in my opinion things can be much simpler
```python
    @property
    def data(self):
        """dict: A data dict representing the volmesh data structure for serialisation.
        """
        # edge_data = {}
        # for e in self._edge_data:
        #     edge_data[repr(e)] = self._edge_data[e]
        cell = {}
        for c in self._cell:
            cell[c] = {}
            for u in self._cell[c]:
                cell[c].setdefault(u, {})
                for v in self._cell[c][u]:
                    cell[c][u][v] = self._halfface[self._cell[c][u][v]]
        data = {
            'attributes': self.attributes,
            'dva': self.default_vertex_attributes,
            'dea': self.default_edge_attributes,
            'dfa': self.default_face_attributes,
            'dca': self.default_cell_attributes,
            'vertex': self._vertex,
            # 'halfface': self._halfface,
            'cell': cell,
            # 'plane': self._plane,
            'edge_data': self._edge_data,
            'face_data': self._face_data,
            'cell_data': self._cell_data,
            'max_vertex': self._max_vertex,
            'max_face': self._max_face,
            'max_cell': self._max_cell}
        return data

    @data.setter
    def data(self, data):
        attributes = data.get('attributes') or {}
        dva = data.get('dva') or {}
        dea = data.get('dea') or {}
        dfa = data.get('dfa') or {}
        dca = data.get('dca') or {}
        vertex = data.get('vertex') or {}
        # halfface = data.get('halfface') or {}
        cell = data.get('cell') or {}
        # plane = data.get('plane') or {}
        edge_data = data.get('edge_data') or {}
        face_data = data.get('face_data') or {}
        cell_data = data.get('cell_data') or {}
        max_vertex = data.get('max_vertex', -1)
        max_face = data.get('max_face', -1)
        max_cell = data.get('max_cell', -1)

        if not vertex or not cell:
            return

        self.attributes.update(attributes)
        self.default_vertex_attributes.update(dva)
        self.default_edge_attributes.update(dea)
        self.default_face_attributes.update(dfa)
        self.default_cell_attributes.update(dca)

        self._vertex = {}
        self._halfface = {}
        self._cell = {}
        self._plane = {}
        self._edge_data = {}
        self._face_data = {}
        self._cell_data = {}

        for v in vertex:
            attr = vertex[v] or {}
            self.add_vertex(int(v), attr_dict=attr)

        # for f in halfface:
        #     attr = face_data.get(f) or {}
        #     self.add_halfface(halfface[f], fkey=int(f), attr_dict=attr)

        for c in cell:
            attr = cell_data.get(c) or {}
            faces = []
            for u in cell[c]:
                for v in cell[c][u]:
                    faces.append(cell[c][u][v])
            self.add_cell(faces, ckey=int(c), attr_dict=attr)

        for e in edge_data:
            self._edge_data[e] = edge_data[e] or {}

        for f in face_data:
            self._face_data[f] = face_data[f] or {}

        # self._edge_data = edge_data
        # self._face_data = face_data
        self._max_vertex = max_vertex
        self._max_face = max_face
        self._max_cell = max_cell
```

brgcode: in the data getter edge data conversion can be commented out because in the new implementation all `edge_data` dict keys are strings by default.

this means the `literal_eval` on the setter side can also be omitted, but perhaps both can be left if for backward compatibility...

brgcode: the half-faces should indeed be created implicitly by the `add_cell` method, and the face identifiers from `f = cell[c][u][v]` have to be converted to strings before looking up the vertices in the half-face dict because of the implicit string conversion by JSON => `faces.append(halfface[str(f)])`.

other than that i don't see the need for additional conversions. all data not stored as dict keys will just retain its original type (int, mostly)

brgcode: the edge data and face data loops at the end could be avoided entirely if not for the potential empty dicts having been converted to lists by JSON

tomvanmele: the problem just lies in the difference between the data dict and the JSON serialized data dict. in the former halfface keys are still `int` and in the latter they will have been converted to `str`. however, we can simply solve this by storing the halffaces directly in a cell dict before serializing rather than in a cell dict that points at a halfface dict...

tomvanmele:
```python
import compas
from compas.datastructures import VolMesh

vmesh1 = VolMesh.from_obj(compas.get('boxes.obj'))

data1 = vmesh1.to_data()

vmesh1.to_json('vmesh1.json')
vmesh1.from_json('vmesh1.json')

vmesh1.validate_data()
vmesh1.validate_json()

data1_ = vmesh1.to_data()

print(data1 == data1_)

vmesh2 = VolMesh.from_data(data1_)

vmesh2.validate_data()
vmesh2.validate_json()

data2 = vmesh2.to_data()

vmesh2.to_json('vmesh2.json')
vmesh2.from_json('vmesh2.json')

data2_ = vmesh2.to_data()

print(data2 == data2_)

print(data1 == data2)
```

tomvanmele: you also have to update the schemas for this to work, though...

tomvanmele:
@Licini ping

tomvanmele:
@Licini pong

Licini:
Since some new changes has already been pushed to master, and not much need to remain the commits of this one. Will make a new PR for adding test.

PR #677, opened 2020-11-24
take into account special vertices for boundary detection

### LINKED ISSUE #548: Compas non-manifold but valid mesh crashes
**Describe the bug**
I am creating some meshes by splitting a bigger mesh. The resulting meshes are valid, but non-manifold. When I try to operate on these meshes (for example to iterate through the boundary edges), the application freezes and slowly eats up all the memory, until it crashes a few minutes later with a 'Bad memory allocation'.  

**To Reproduce**
I created a sample script that reproduces the crash. When you run it, open the task manager to observe the memory usage. You can download it here.
https://www.dropbox.com/sh/uipx8gl234p9nmi/AAA2tKOcy25DR0gDJJ85Af5na?dl=0

**Expected behavior**
It should print out the edges that are on the boundary. Instead it crashes after using all the available memory. 

 - OS:  Windows 10
 - Python version: Python 3.7.6
 - Python package manager: conda

Compas version :  0.15.6 (latest)

### LINKED ISSUE #675: mesh.vertices_on_boundary() for N-gon mesh
**Describe the bug**
The function `mesh.vertices_on_boundary()` never arrives to the exit condition of the while loop, appending infinitely to the list `vertices`, and eventually causing a `Runtime error (OutOfMemoryException): Array dimensions exceeded supported range`.
The definition of the `start` key might be the problem.

**To Reproduce**
Script to reproduce the behavior in Python 3.7:
```
mesh_json = {
    "compas": "0.17.2",
    "datatype": "compas.datastructures/Mesh",
    "data": {
        "attributes": {
            "name": "Mesh"
        },
        "dva": {
            "x": 0.0,
            "y": 0.0,
            "z": 0.0
        },
        "dea": {},
        "dfa": {},
        "vertex": {
            "0": {
                "x": 17.56081771850586,
                "y": 4.622284412384033,
                "z": 1.2479242086410522
            },
            "1": {
                "x": 18.21266746520996,
                "y": 5.342650413513184,
                "z": 2.371530055999756
            },
            "2": {
                "x": 20.60313606262207,
                "y": 5.956735610961914,
                "z": 2.9836647510528564
            },
            "3": {
                "x": 17.613462448120117,
                "y": 5.7198333740234375,
                "z": 2.9479243755340576
            },
            "4": {
                "x": 19.751209259033203,
                "y": 7.413698196411133,
                "z": 2.355659008026123
            },
            "5": {
                "x": 20.739913940429688,
                "y": 6.193633556365967,
                "z": 2.9836647510528564
            },
            "6": {
                "x": 17.75023651123047,
                "y": 5.956733703613281,
                "z": 2.9479243755340576
            },
            "7": {
                "x": 21.414566040039062,
                "y": 6.00639533996582,
                "z": 2.5479743480682373
            },
            "8": {
                "x": 17.33991241455078,
                "y": 5.719833850860596,
                "z": 2.9479243755340576
            },
            "9": {
                "x": 21.61810302734375,
                "y": 7.367147445678711,
                "z": 1.2479242086410522
            },
            "10": {
                "x": 16.98848533630371,
                "y": 4.607467174530029,
                "z": 1.2479242086410522
            },
            "11": {
                "x": 20.739912033081055,
                "y": 5.7198357582092285,
                "z": 2.9836647510528564
            },
            "12": {
                "x": 21.01346206665039,
                "y": 5.719833850860596,
                "z": 2.9836647510528564
            },
            "13": {
                "x": 17.02552032470703,
                "y": 7.527175426483154,
                "z": 2.3149938583374023
            },
            "14": {
                "x": 17.33991241455078,
                "y": 6.193633556365967,
                "z": 2.9479243755340576
            },
            "15": {
                "x": 21.013463973999023,
                "y": 6.19363260269165,
                "z": 2.9836647510528564
            },
            "16": {
                "x": 21.0572452545166,
                "y": 4.767782211303711,
                "z": 1.2479242086410522
            },
            "17": {
                "x": 19.797176361083984,
                "y": 6.417858123779297,
                "z": 2.5859525203704834
            },
            "18": {
                "x": 17.187955856323242,
                "y": 5.521986484527588,
                "z": 2.645582914352417
            },
            "19": {
                "x": 22.073240280151367,
                "y": 6.379892826080322,
                "z": 1.2479242086410522
            },
            "20": {
                "x": 21.1451473236084,
                "y": 5.397745609283447,
                "z": 2.3964271545410156
            },
            "21": {
                "x": 21.800334930419922,
                "y": 5.456541538238525,
                "z": 1.2479242086410522
            },
            "22": {
                "x": 18.767168045043945,
                "y": 6.26754903793335,
                "z": 2.357071876525879
            },
            "23": {
                "x": 15.935832023620605,
                "y": 5.5590925216674805,
                "z": 1.247817873954773
            },
            "24": {
                "x": 17.203136444091797,
                "y": 5.956733703613281,
                "z": 2.9479243755340576
            },
            "25": {
                "x": 18.05254554748535,
                "y": 7.338461399078369,
                "z": 2.4046356678009033
            },
            "26": {
                "x": 19.065616607666016,
                "y": 4.965931415557861,
                "z": 1.7346264123916626
            },
            "27": {
                "x": 21.150230407714844,
                "y": 5.9567365646362305,
                "z": 2.9836647510528564
            },
            "28": {
                "x": 17.613460540771484,
                "y": 6.193634510040283,
                "z": 2.9479243755340576
            },
            "29": {
                "x": 16.063350677490234,
                "y": 6.945024013519287,
                "z": 1.2477844953536987
            },
            "30": {
                "x": 20.833829879760742,
                "y": 6.970712661743164,
                "z": 2.58367657661438
            },
            "31": {
                "x": 20.284622192382812,
                "y": 4.733113765716553,
                "z": 1.2479242086410522
            },
            "32": {
                "x": 16.275901794433594,
                "y": 7.384394645690918,
                "z": 1.2477434873580933
            },
            "33": {
                "x": 21.28115463256836,
                "y": 7.4636454582214355,
                "z": 1.6519025564193726
            },
            "34": {
                "x": 20.048507690429688,
                "y": 5.355278491973877,
                "z": 2.319051742553711
            },
            "35": {
                "x": 16.516220092773438,
                "y": 6.338499546051025,
                "z": 2.418168544769287
            }
        },
        "face": {
            "0": [
                34,
                16,
                20,
                12,
                11
            ],
            "1": [
                20,
                21,
                7,
                27,
                12
            ],
            "2": [
                20,
                16,
                21
            ],
            "3": [
                21,
                19,
                7
            ],
            "4": [
                27,
                7,
                19,
                30,
                15
            ],
            "5": [
                4,
                30,
                33
            ],
            "6": [
                26,
                31,
                16,
                34
            ],
            "7": [
                9,
                33,
                30,
                19
            ],
            "8": [
                10,
                1,
                3,
                8,
                18
            ],
            "9": [
                11,
                2,
                17,
                34
            ],
            "10": [
                26,
                1,
                10,
                0
            ],
            "11": [
                0,
                31,
                26
            ],
            "12": [
                35,
                23,
                18,
                8,
                24
            ],
            "13": [
                30,
                4,
                5,
                15
            ],
            "14": [
                34,
                17,
                26
            ],
            "15": [
                18,
                23,
                10
            ],
            "16": [
                22,
                6,
                3,
                1
            ],
            "17": [
                2,
                5,
                4,
                17
            ],
            "18": [
                17,
                22,
                26
            ],
            "19": [
                1,
                26,
                22
            ],
            "20": [
                24,
                14,
                13,
                35
            ],
            "21": [
                13,
                14,
                28,
                25
            ],
            "22": [
                4,
                25,
                17
            ],
            "23": [
                25,
                22,
                17
            ]
        },
        "facedata": {
            "0": {},
            "1": {},
            "2": {},
            "3": {},
            "4": {},
            "5": {},
            "6": {},
            "7": {},
            "8": {},
            "9": {},
            "10": {},
            "11": {},
            "12": {},
            "13": {},
            "14": {},
            "15": {},
            "16": {},
            "17": {},
            "18": {},
            "19": {},
            "20": {},
            "21": {},
            "22": {},
            "23": {}
        },
        "edgedata": {},
        "max_vertex": 35,
        "max_face": 23
    }
}

built_mesh = Mesh.from_data(mesh_json)
built_mesh.edges_on_boundary()
```

**Expected behavior**
Returning the boundary edges of the mesh.

**Desktop:**
 - OS: Windows 10
 - Python version: 3.7
 - Python package manager: conda
 - compas version: 0.17.2

**Additional context**
Are N-gon meshes generally supported/accepted in compas?
What other functions could lead to problems for N-gon meshes?

tomvanmele:

### Checklist

### DIFF
--- a/src/compas/datastructures/mesh/core/mesh.py
+++ b/src/compas/datastructures/mesh/core/mesh.py
@@ -1262,95 +1262,45 @@ class BaseMesh(HalfEdge):
     # boundary
     # --------------------------------------------------------------------------
 
-    def vertices_on_boundary(self, ordered=False):
-        """Find the vertices on the boundary.
-
-        Parameters
-        ----------
-        ordered : bool, optional
-            If ``True``, Return the vertices in the same order as they are found on the boundary.
-            Default is ``False``.
+    def vertices_on_boundary(self):
+        """Find the vertices on the longest boundary.
 
         Returns
         -------
         list
-            The vertices of the boundary.
-
-        Warnings
-        --------
-        If the vertices are requested in order, and the mesh has multiple borders,
-        currently only the vertices of one of the borders will be returned.
-
-        Examples
-        --------
-        >>>
+            The vertices of the longest boundary.
 
         """
-        # vertices_set = set()
-        # for key, nbrs in iter(self.halfedge.items()):
-        #     for nbr, face in iter(nbrs.items()):
-        #         if face is None:
-        #             vertices_set.add(key)
-        #             vertices_set.add(nbr)
-        # vertices = list(vertices_set)
-
-        # if not vertices:
-        #     return vertices
-        # if not ordered:
-        #     return vertices
-
-        # key = sorted([(key, self.vertex_coordinates(key)) for key in vertices], key=lambda x: (x[1][1], x[1][0]))[0][0]
-
-        # vertices = []
-        # start = key
-        # while True:
-        #     for nbr, fkey in iter(self.halfedge[key].items()):
-        #         if fkey is None:
-        #             vertices.append(nbr)
-        #             key = nbr
-        #             break
-        #     if key == start:
-        #         break
-        # return vertices
         boundaries = self.vertices_on_boundaries()
         if boundaries:
             return boundaries[0]
         return []
 
     def edges_on_boundary(self, oriented=False):
-        """Find the edges on the boundary.
-
-        Parameters
-        ----------
-        oriented : bool, optional
-            If ``False`` (default) the edges are aligned head-to-tail along the boundary.
-            If ``True`` the edges have the same orientation as in the mesh.
+        """Find the edges on the longest boundary.
 
         Returns
         -------
         edges : list
-            The boundary edges.
+            The edges of the longest boundary.
         """
-        vertices = self.vertices_on_boundary()
-        edges = list(pairwise(vertices + vertices[:1]))
-        if not oriented:
-            return edges
-        return [(u, v) if self.has_edge((u, v)) is None else (v, u) for u, v in edges]
+        boundaries = self.edges_on_boundaries()
+        if boundaries:
+            return boundaries[0]
+        return []
 
     def faces_on_boundary(self):
-        """Find the faces on the boundary.
+        """Find the faces on the longest boundary.
 
         Returns
         -------
         list
-            The faces on the boundary.
+            The faces on the longest boundary.
         """
-        faces = []
-        for u, v in self.edges_on_boundary():
-            fkey = self.halfedge[v][u]
-            if fkey not in faces:
-                faces.append(fkey)
-        return faces
+        boundaries = self.faces_on_boundaries()
+        if boundaries:
+            return boundaries[0]
+        return []
 
     def vertices_on_boundaries(self):
         """Find the vertices on all boundaries of the mesh.
@@ -1364,6 +1314,7 @@ class BaseMesh(HalfEdge):
         --------
         >>>
         """
+        # all boundary vertices
         vertices_set = set()
         for key, nbrs in iter(self.halfedge.items()):
             for nbr, face in iter(nbrs.items()):
@@ -1372,9 +1323,16 @@ class BaseMesh(HalfEdge):
                     vertices_set.add(nbr)
         vertices_all = list(vertices_set)
 
+        # return an empty list if there are no boundary vertices
         if not vertices_all:
-            return vertices_all
+            return []
 
+        # init container for boundary groups
+        boundaries = []
+
+        # identify *special* vertices
+        # these vertices are non-manifold
+        # and should be processed differently
         special = []
         for key in vertices_all:
             count = 0
@@ -1382,38 +1340,61 @@ class BaseMesh(HalfEdge):
                 face = self.halfedge_face(key, nbr)
                 if face is None:
                     count += 1
-            if count > 1:
-                special.append(key)
+                    if count > 1:
+                        if key not in special:
+                            special.append(key)
 
-        boundaries = []
+        superspecial = special[:]
 
-        if special:
-            for start in special:
-                for key in self.vertex_neighbors(start):
-                    face = self.halfedge_face(start, key)
-                    if face is None:
-                        vertices = [start, key]
-                        while True:
-                            for nbr in self.vertex_neighbors(key):
-                                face = self.halfedge_face(key, nbr)
-                                if face is None:
-                                    if nbr != start:
-                                        vertices.append(nbr)
-                                    key = nbr
-                                    break
-                            if key == start:
-                                boundaries.append(vertices)
-                                vertices_all = [x for x in vertices_all if x not in vertices]
-                                break
+        # process the special vertices first
+        while special:
+            start = special.pop()
+            nbrs = []
+            # find all neighbors of the current spacial vertex
+            # that are on the mesh boundary
+            for nbr in self.vertex_neighbors(start):
+                face = self.halfedge_face(start, nbr)
+                if face is None:
+                    nbrs.append(nbr)
+            # for normal mesh vertices
+            # there should be only 1 boundary neighbor
+            # for special vertices there are more and they all have to be processed
+            while nbrs:
+                vertex = nbrs.pop()
+                vertices = [start, vertex]
+                while True:
+                    # this is a *super* special case
+                    if vertex in superspecial:
+                        boundaries.append(vertices)
+                        break
+                    # find the boundary loop for the current starting halfedge
+                    for nbr in self.vertex_neighbors(vertex):
+                        if nbr == vertices[-2]:
+                            continue
+                        face = self.halfedge_face(vertex, nbr)
+                        if face is None:
+                            vertices.append(nbr)
+                            vertex = nbr
+                            break
+                    if vertex == start:
+                        boundaries.append(vertices)
+                        break
+                # remove any neighbors that might be part of an already identified boundary
+                nbrs = [vertex for vertex in nbrs if vertex not in vertices]
 
+        # remove all boundary vertices that were already identified
+        vertices_all = [vertex for vertex in vertices_all if all(vertex not in vertices for vertices in boundaries)]
+
+        # process the remaining boundary vertices if any
         if vertices_all:
             key = vertices_all[0]
             while vertices_all:
-                vertices = []
+                vertices = [key]
                 start = key
                 while True:
-                    for nbr, fkey in iter(self.halfedge[key].items()):
-                        if fkey is None:
+                    for nbr in self.vertex_neighbors(key):
+                        face = self.halfedge_face(key, nbr)
+                        if face is None:
                             vertices.append(nbr)
                             key = nbr
                             break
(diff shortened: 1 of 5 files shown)


tomvanmele:
not sure if it is entirely backwards compatible though. in some cases it will detect the "outer boundary" differently...

tomvanmele:
we should also reconsider the checks `is_valid` and `is_manifold`. the mesh in the linked issue passes `is_valid` but not `is_manifold`, which is misleading to say the least...

src/compas/datastructures/mesh/core/mesh.py
```diff
@@ -1286,32 +1286,36 @@ def vertices_on_boundary(self, ordered=False):
...
+        #             vertices.append(nbr)
+        #             key = nbr
+        #             break
+        #     if key == start:
+        #         break
+        # return vertices
+        boundaries = self.vertices_on_boundaries()
+        if boundaries:
+            return boundaries[0]
+        return []
```
tomvanmele: if we agree on the new implementation, this should be deleted...

src/compas/datastructures/mesh/core/mesh.py
```diff
@@ -1371,25 +1375,56 @@ def vertices_on_boundaries(self):
...
+                            for nbr in self.vertex_neighbors(key):
+                                face = self.halfedge_face(key, nbr)
+                                if face is None:
+                                    if nbr != start:
+                                        vertices.append(nbr)
+                                    key = nbr
+                                    break
+                            if key == start:
+                                boundaries.append(vertices)
+                                vertices_all = [x for x in vertices_all if x not in vertices]
```
beverlylytle: Since `vertices_all` isn't used  in the lines 1390-1407, it would be more efficient to update this value outside of the loops. 
```
processed_vertices = set().union(*boundaries)
vertices_all = [x for x in vertices_all if x not in processed_vertices]

File: src/compas/datastructures/mesh/core/mesh.py [hunk 3]
```diff
@@ -1371,25 +1375,56 @@ def vertices_on_boundaries(self):
...
+                        vertices = [start, key]
+                        while True:
+                            for nbr in self.vertex_neighbors(key):
+                                face = self.halfedge_face(key, nbr)
+                                if face is None:
+                                    if nbr != start:
+                                        vertices.append(nbr)
+                                    key = nbr
+                                    break
+                            if key == start:
```
beverlylytle: Do you suppose there are non-manifold meshes for which this condition wouldn't be met, eg a cylinder with a base the shape of a "p" with the starting special vertex somewhere on the tail of the "p", and so this loop would be infinite?  Or is that not an edge case we care about/is valid? (Pun unavoidable.)

src/compas/datastructures/mesh/core/mesh.py
```diff
@@ -1371,25 +1375,56 @@ def vertices_on_boundaries(self):
...
-        key = sorted([(key, self.vertex_coordinates(key)) for key in vertices_all], key=lambda x: (x[1][1], x[1][0]))[0][0]
+        special = []
+        for key in vertices_all:
+            count = 0
+            for nbr in self.vertex_neighbors(key):
+                face = self.halfedge_face(key, nbr)
+                if face is None:
+                    count += 1
+            if count > 1:
+                special.append(key)
```
beverlylytle: If this were included in the `for nbr in self.vertex_neighrbors(key)` loop and ended with a `break`, it might fail faster.  Well, not "fail", but...

src/compas/datastructures/mesh/core/mesh.py
beverlylytle: Or maybe consider the more concrete case of a mesh consisting of three triangles, with the middle triangle sharing one vertex with the triangle to the left and one vertex with the triangle to the right, giving two special vertices in the mesh, and all edges are boundary edges.  I think the loops in this case might result in "random" walks around the boundary.

tomvanmele: yes there are still (many) uncovered cases. perhaps it is better to limit the function to manifold meshes? we could add a separate algorithm that still tries to find boundaries but with the disclaimer that it is to be used "at your own discretion"...

beverlylytle: Would it makes sense to turn the logic around for these boundary methods?  First one finds all of the faces on the boundary by taking a reference count for each unoriented edge on each face.  The faces attached to edges that have reference count 1 are the boundary faces.  This avoids the non-manifold issues because there are no special faces, and special edges (those that are coincident to three or more faces) are thrown out in the process.  This also finds the collection of boundary edges, from which we want the connected components.  For this, we can start playing dominoes with the edges, by which I mean, we can initialize a boundary component list with a randomly chosen boundary edge and choose a vertex from that edge, perform a breadth-first search, looking through the remaining edges to find ones coincident with that vertex, append those to our current boundary component (and remove them from the "remaining edges"), and then move on to the next edge in the component until we reach the end, and then start again with a new randomly chosen boundary edge.  The order of the edges (and hence vertices) within a component for those without special vertices is correct.  And for those with... well, I don't think order is well-defined.

tomvanmele: that is actually a very good idea. will try to implement something along those lines...

src/compas/datastructures/mesh/core/mesh.py
gonzalocasas: maybe debatable opinion but here I feel the one-liner actually improves readability:

```suggestion
        return boundaries[0] if boundaries else []
```

gonzalocasas (commented):
i don't have much to add except perhaps it would make sense to add some unit tests based on the failing examples of the linked issue.

tests/compas/datastructures/test_mesh.py
```diff
@@ -235,16 +264,10 @@ def test_is_regular():
...
     pass
-    # TODO: this test needs to be passed!
-    # mesh = Mesh.from_off(compas.get('cube.off'))
-    # assert mesh.is_manifold()
-
-    # mesh.add_face([0, 1, 3])
-    # assert not mesh.is_manifold()
 
 
 def test_is_orientable():
```
gonzalocasas, line 275: you could mark these as `skip` instead of keeping empty functions (https://docs.pytest.org/en/latest/skipping.html)

src/compas/datastructures/mesh/core/mesh.py
tomvanmele: so this didn't actually work but i extended the algorithm based on the separate processing of non-manifold vertices. it now always seems to work (see the tests) ...

tests/compas/datastructures/test_mesh.py
tomvanmele: i will add an implementation but as part of separate PR. just took out the commented code. they were already empty...

tomvanmele:
@gonzalocasas @beverlylytle how do we feel about this one?

gonzalocasas:
i think it's good to go. I would delete the commented out code, because OCD, but not a big deal.

gonzalocasas:
looks good to me! I changed the type to breaking change because there are removed methods, signature changes and so on.

tomvanmele:
so perhaps a patch release before merging this one and then i bump the minor?

tomvanmele:
if the ironpython agrees of course

PR #681, opened 2020-11-26
Urdf writer

### LINKED ISSUE #670: Add Urdf writers
Currently, the URDF support can only read/parse urdf files, but cannot write them back to disk if they are modified.
Adding urdf writer support would help in other areas, such as mix&match different robot clients in compas_fab

beverlylytle:
Closes #670. 

### Checklist

- [X] I added a line to the `CHANGELOG.md` file in the `Unreleased` section under the most fitting heading (e.g. `Added`, `Changed`, `Removed`).
- [X] I ran all tests on my computer and it's all green (i.e. `invoke test`).
- [X] I ran lint on my computer and there are no errors (i.e. `invoke lint`).
- [X] I added new functions/classes and made them available on a second-level import, e.g. `compas.datastructures.Mesh`.
- [X] I have added tests that prove my fix is effective or that my feature works.
- [X] I have added necessary documentation (if appropriate)


beverlylytle:
I tinkered quite a lot with `compas.files.XML` to make it look more like other things in `compas.files`, but maybe now it's just harder to read.  I can change it back if desired.

gonzalocasas:
I just went through this one superficially and I will do a more proper review tomorrow, but in the meantime, it's looking real good! Very clean! 💯

CHANGELOG.md
```diff
@@ -9,7 +9,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+* Added URDF and XML writers.
+* Added `compas.files.URDF.from_robot`.
```
tomvanmele, line 14: should it not be the other way round? `compas.robots.RobotModel.from/to_urdf`

tests/compas/robots/test_model.py
```diff
@@ -85,6 +97,15 @@ def test_root_urdf_attributes():
     assert r.name == 'panda'
 
 
+def test_root_urdf_attributes_to_string():
+    r = RobotModel.from_urdf_string(
+        """<?xml version="1.0" encoding="UTF-8"?><robot xmlns:xacro="http://www.ros.org/wiki/xacro" name="panda"></robot>""")
+    urdf = URDF.from_robot(r)
+    assert urdf.to_string(prettify=True) == b"""<?xml version="1.0" encoding="utf-8"?>\n<robot name="panda"/>\n"""
+    # Note: the Processing Instruction 'xmlns:xacro="http://www.ros.org/wiki/xacro"'
+    # is lost on `from_urdf_string` due to limitations of `xml.etree.ElementTree`
```
beverlylytle: This is wrong.  I'm working on fixing this.

src/compas/files/xml_cli.py
```diff
@@ -146,11 +193,11 @@ def _parse_xml_declaration(self, xml_decl):
     def _start_element(self, reader):
         """Notify the tree builder that a start element has been
         encountered."""
-        name = reader.LocalName
+        name = reader.Name
```
beverlylytle: I'm terribly confused about this.  When I run the code from the tests in rhino on my machine, it works.  I have no idea why all the ironpython tests are failing now.

beverlylytle: I replaced all references to `LocalName` to `Name` and all tests pass.  I looked briefly at the source code of `aglyph.compat.ipyetree` and did not see a good reason why `LocalName` is used there.  I've included two (nearly identical) tests involving namespaces.  One for xml and one for urdf.  But probably there should be more.  Ideas?

src/compas/files/xml.py
```diff
@@ -100,16 +178,83 @@ def __init__(self, root):
...
+            XML string
+        """
+        reparsed = minidom.parseString(rough_string)
+        return reparsed.toprettyxml(indent="  ", encoding='utf-8')
+
+    def attach_namespaces(root, source):
+        """Parse and find the namespaces declared, and add them to the root's attributes."""
+        namespaces = [node for _, node in ET.iterparse(source, events=['start-ns'])]
+        attrib = {'xmlns:' + ns: uri for ns, uri in namespaces}
+        root.attrib.update(attrib)
```
tomvanmele: should this not go to the top of the file?

beverlylytle: Not sure.  @gonzalocasas is there a reason you put it at bottom of the file?

src/compas/robots/model/geometry.py
```diff
@@ -501,6 +535,10 @@ def __init__(self, filename):
         super(Texture, self).__init__()
         self.filename = filename
 
+    def get_urdf_element(self):
```
tomvanmele, line 542: shouldn't all these be named `to_urdf` or `to_urdf_element`?

src/compas/robots/model/robot.py
```diff
@@ -178,6 +185,21 @@ def from_urdf_file(cls, file):
         urdf = URDF.from_file(file)
         return urdf.robot
 
+    def to_urdf_file(self, file, prettify=False):
```
tomvanmele, line 188: should we here do the same as what was discussed in relation to the file readers in general? just do `to_urdf` and differentiate between file or string based on what is passed as parameter to the function?

tomvanmele (commented):
generally looks good. some minor stuff here and there. perhaps for the tests it would be useful if we could identify and include some boundary cases...

src/compas/files/xml.py
gonzalocasas: No real reason that I remember. Probably leftovers of an early draft.

src/compas/robots/model/robot.py
beverlylytle: I thought we discussed something a little different from this.  That `read` should take either a file-like object or a file path.  I guess one could argue that `to_urdf_file` should take a file path or a file-like object and write to those places, but the `to_string` methods are different. They return a string.  I don't think the two should be combined.

tomvanmele: sounds good

src/compas/files/xml.py
```diff
@@ -2,14 +2,49 @@
 from __future__ import division
 from __future__ import print_function
 
+
+from io import BytesIO
```
gonzalocasas: totally minor styling comment, but, sometimes (this being one of those times) I feel it reads nicer to the package and use fully-qualified class names, ie `import io` and then use `io.BytesIO` and `io.StringIO` rather than explicitly importing each globally

src/compas/robots/model/geometry.py
```diff
@@ -403,6 +427,12 @@ def __init__(self, filename, scale='1.0 1.0 1.0'):
         self.filename = filename
         self.scale = _parse_floats(scale)
 
+    def get_urdf_element(self):
+        attributes = {'filename': self.filename}
+        if self.scale != [1.0, 1.0, 1.0]:
```
gonzalocasas, line 436: I would add a little inlined comment here to explain why this is a special case. (I think it's just because that's the default, right?)

tests/compas/files/test_xml.py
```diff
@@ -31,3 +32,20 @@ def test_xml_to_string(sample_xml):
     xml = XML.from_string(sample_xml)
     strxml = xml.to_string('utf-8')
     assert strxml.startswith(b'<Tests>')
+
+
+def test_xml_to_pretty_string(sample_xml):
+    xml = XML.from_string(sample_xml)
+    prettyxml = xml.to_string(prettify=True)
+    print(b"\n  " in prettyxml)
```
gonzalocasas: this test is a bit superfluous because it does not assert anything and pytest swallows all stdout so the print statement is never visible. Maybe you can try to assert the presence of a partial part of the sample in the pretty printed version like you do on other tests?

src/compas/files/xml.py
```diff
@@ -2,14 +2,49 @@
...
+        """
+        reparsed = minidom.parseString(rough_string)
+        return reparsed.toprettyxml(indent="  ", encoding='utf-8')
+
+    def attach_namespaces(root, source):
+        """Parse and find the namespaces declared, and add them to the root's attributes."""
+        if hasattr(source, 'seek'):
+            source.seek(0)
+        namespaces = [node for _, node in ET.iterparse(source, events=['start-ns'])]
+        attrib = {'xmlns:' + ns: uri for ns, uri in namespaces}
```
gonzalocasas: I found one more case that the current tests don't cover and is failing on cpython. If there's a "default" namespace, eg, one that doesn't define a prefix, the parser is unable to parse correctly. Here's one example of a file that fails (from the Drake collection):
```
<?xml version="1.0"?>
<robot xmlns="https://drake.mit.edu" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" name="Acrobot">
  <link name="base_link">
    <visual>
      <geometry>
        <box size="0.2 0.2 0.2"/>
      </geometry>
      <material name="green">
        <color rgba="0 1 0 1"/>
      </material>
    </visual>
    <collision name="base_link_collision">
      <geometry>
        <box size="0.2 0.2 0.2"/>
      </geometry>
    </collision>
  </link>
  <link name="upper_link">
    <inertial>
      <origin xyz="0 0 -.5" rpy="0 0 0"/>
      <mass value="1"/>
      <inertia ixx="1" ixy="0" ixz="0" iyy="0.083" iyz="0" izz="1"/>
    </inertial>
    <visual>
      <origin xyz="0 0 -0.5" rpy="0 0 0"/>
      <geometry>
        <cylinder length="1.1" radius="0.05"/>
      </geometry>
      <material name="red">
        <color rgba="1 0 0 1"/>
      </material>
    </visual>
    <collision name="upper_link_collision">
      <origin xyz="0 0 -0.5" rpy="0 0 0"/>
      <geometry>
        <cylinder length="1.1" radius="0.05"/>
      </geometry>
    </collision>
  </link>
  <link name="lower_link">
    <inertial>
      <origin xyz="0 0 -1" rpy="0 0 0"/>
      <mass value="1"/>
      <inertia ixx="1" ixy="0" ixz="0" iyy="0.33" iyz="0" izz="1"/>
    </inertial>
    <visual>
      <origin xyz="0 0 -1" rpy="0 0 0"/>
      <geometry>
        <cylinder length="2.1" radius=".05"/>
      </geometry>
      <material name="blue">
        <color rgba="0 0 1 1"/>
      </material>
    </visual>
    <collision name="lower_link_collision">
      <origin xyz="0 0 -1" rpy="0 0 0"/>
      <geometry>
        <cylinder length="2.1" radius=".05"/>
      </geometry>
    </collision>
  </link>
  <joint name="base_weld" type="fixed">
    <parent link="world"/>
    <child link="base_link"/>
  </joint>
  <joint name="shoulder" type="continuous">
    <parent link="base_link"/>
    <child link="upper_link"/>
    <origin xyz="0 0.15 0"/>
    <axis xyz="0 1 0"/>
    <dynamics damping="0.1"/>
  </joint>
  <joint name="elbow" type="continuous">
    <parent link="upper_link"/>
    <child link="lower_link"/>
    <origin xyz="0 0.1 -1"/>
    <axis xyz="0 1 0"/>
    <dynamics damping="0.1"/>
  </joint>
  <transmission type="SimpleTransmission" name="elbow_trans">
    <actuator name="elbow"/>
    <joint name="elbow"/>
    <mechanicalReduction>1</mechanicalReduction>
  </transmission>
  <frame name="hand" link="lower_link" xyz="0 0 -2.1" rpy="0 0 0"/>
</robot>
```

I found this when running the `python  tests/compas/robots/test_model.py`.
As an additional note: I have only tested this on python 3.8 and I read that (at least part of) this behavior changed on 3.8, so it would be good to add the test and see how this behaves on 3.6/3.7.

gonzalocasas (changes requested):

tests/compas/files/test_xml.py
beverlylytle: Ha, that was supposed to be `assert b"\n  " in prettyxml`, which would assert that at least a new line and a (perhaps partial) tab was injected into the string.  Between all of the different behaviors of the implementations of the parsers, it's hard to say exactly what spacing will be where and what the indentation will be.  Hence the loose test.

src/compas/files/xml.py
beverlylytle: Regarding Python versions, my dev environment is python 3.7, which has informed many of the choices made regarding namespaces.

beverlylytle: I've added a change that should support default namespaces.  Now let's hope the new tests work for ipy on windows.

tests/compas/files/test_xml.py
```diff
@@ -49,3 +49,12 @@ def test_namespaces_to_string():
     # Note: Minidom does some funny things to namespaces.  First, if a namespace isn't used, it will be stripped out.
     # Second, it will include the original namespace declaration, but also repeat that declaration with another name,
     # and replace all references to the original with the new.
+
+def test_default_namespace_to_string():
+    xml = XML.from_string(
+        """<?xml version="1.0" encoding="UTF-8"?><robot xmlns="https://default.org/namespace" xmlns:xacro="http://www.ros.org/wiki/xacro" name="panda"><xacro:bamboo/></robot>"""
+    )
+    xml_string = xml.to_string(prettify=True)
```
gonzalocasas, line 58: Interesting that this test passes! That means we have another issue somewhere else, because my `acrobot.urdf` test still fails. The main difference being that I don't test full round trip, but just parse and try to access values:

```python
@pytest.fixture
def acrobot_file():
    return os.path.join(BASE_FOLDER, 'fixtures', 'acrobot.urdf')

def test_acrobot_urdf(acrobot_file):
    r = RobotModel.from_urdf_file(acrobot_file)
    assert r.name == 'Acrobot'

```

beverlylytle: 🤔

beverlylytle: Oh, dear, something is very broken indeed.  Your `r` isn't even a `RobotModel` instance. (Duck python typing.)

beverlylytle: When there is a default namespace, all of the tags are prepended with `{<default xmlns value>}` in the python xml element tree parser. So when creating the `URDF`, `parse_element` is passed `{https://drake.mit.edu}robot` as a path rather than simply `robot`.  (This behavior exists on the main branch currently.). Since the `URDFParser` isn't registered with such a path, it creates a `URDFGenericElement` instead of a `RobotModel`.  There is a way to interrupt the parse and inject our own tags stripped of the namespace, but I don't think that's wise.  Maybe a drake robot is indeed a different creature than an unmodified robot.  What do you think?

gonzalocasas: Yep, it is a pretty crappy situation. The only "solution" I found on the interwebz was doing a search/replace to strip that empty default namespace from the text. I tested it on the `XmlReader.from_string()` and it works, but I did not find a way to do it for the `XmlReader.from_file()`. A first naive approach was to read all file content and treat it as string, but that failed miserably when feeding it the `test_model.py` examples that does in-memory io from a zip file.

We could leave it as it is (since it is already like this on master), however, I don't know when it broke because it used to work and it should be fixed at some point. I suspect there's likely a way to hijack the XML parser and ignore the default  namespace.

beverlylytle: I've been digging around and not found anything acceptable.  So should I make an issue of this and merge this PR as is?

src/compas/robots/model/geometry.py
beverlylytle: I wouldn't want a `to_urdf_element` or `to_urdf` method on `RobotModel` to be confused by a user with something that produces a urdf file.  So I prefer to leave this largely internal method name as is.

tests/compas/files/test_xml.py
gonzalocasas: Yes, as discussed, this is good to go, especially now that you added the extra (marked as skip) test for this situation.

PR #682, opened 2020-11-27
improved surface to mesh conversion for trimmed surfaces

tomvanmele:

### Checklist

### DIFF
--- a/src/compas_rhino/geometry/surface.py
+++ b/src/compas_rhino/geometry/surface.py
@@ -40,9 +40,9 @@ class RhinoSurface(BaseRhinoGeometry):
             The type of COMPAS mesh.
         facefilter : callable, optional
             A filter for selection which Brep faces to include.
-            Default is ``None`` in which case all faces are included.
-            If provided, the filter should return ``True``/``False`` per faces.
-            A simple filter that includes all faces is ``def facefilter(face): return True``.
+            If provided, the filter should return ``True`` or ``False`` per face.
+            A very simple filter that includes all faces is ``def facefilter(face): return True``.
+            Default parameter value is ``None`` in which case all faces are included.
         cleanup : bool, optional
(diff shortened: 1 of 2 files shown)


tomvanmele:
Screenshot 2020-11-27 at 09 05 41

for handling surfaces like the one on the left...

```python
import compas_rhino

from compas_rhino.geometry import RhinoSurface
from compas_rhino.artists import MeshArtist


def facefilter(face):
    success, w, h = face.GetSurfaceSize()
    if success:
        if w > 10 and h > 10:
            return True
    return False


guid = compas_rhino.select_surface()
surf = RhinoSurface.from_guid(guid)
mesh = surf.to_compas(facefilter=facefilter)

artist = MeshArtist(mesh, layer="Blocks")
artist.clear_layer()
artist.draw()
```

gonzalocasas (commented):
I can't comment much on the actual functionality. I do find the `_a_` and `_b_` to be very weird names for variables, it's like they would have special meaning but they don't, so I would actually prefer a more regular name, like `a_key`, `b_key`. And then, there's the missing changelog line.

src/compas_rhino/geometry/surface.py
```diff
@@ -31,13 +31,20 @@ def from_selection(cls):
...
+    def to_compas(self, cls=None, facefilter=None, cleanup=True):
         """Convert the surface b-rep loops to a COMPAS mesh.
 
         Parameters
         ----------
         cls : :class:`compas.datastructures.Mesh`, optional
             The type of COMPAS mesh.
+        facefilter : callable, optional
+            A filter for selection which Brep faces to include.
+            Default is ``None``.
```
beverlylytle: This might lead to some confusion.  The default behavior is to include all faces, right?  This could be read to say that no faces are included.

Licini (commented):
Look all good to me!

PR #683, opened 2020-11-30
typo in normal polygon: cross vectors being added need to be halved

juney-lee:

### DIFF
--- a/src/compas/geometry/_core/normals.py
+++ b/src/compas/geometry/_core/normals.py
@@ -9,6 +9,7 @@ from compas.geometry._core import cross_vectors_xy
 from compas.geometry._core import length_vector
 from compas.geometry._core import length_vector_xy
 from compas.geometry._core import normalize_vector
+from compas.geometry._core import scale_vector
 
 from compas.geometry._core import centroid_points
 
@@ -59,7 +60,7 @@ def normal_polygon(polygon, unitized=True):
     for i in range(p):
         b = polygon[i]
         ob = subtract_vectors(b, o)
-        n = cross_vectors(oa, ob)
+        n = scale_vector(cross_vectors(oa, ob), 0.5)
         oa = ob
 
         nx += n[0]
--- a/src/compas/geometry/_core/normals.py
+++ b/src/compas/geometry/_core/normals.py
@@ -9,7 +9,6 @@ from compas.geometry._core import cross_vectors_xy
 from compas.geometry._core import length_vector
 from compas.geometry._core import length_vector_xy
 from compas.geometry._core import normalize_vector
-from compas.geometry._core import scale_vector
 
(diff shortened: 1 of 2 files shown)


juney-lee:
Not sure how we overlooked this... 

`area_polygon` is basically `length_vector(normal_polygon(polygon, unitized=False)`. But since they are presumably used for very different purposes (area for mesh faces, and unitized normals for cells and volmesh where the weighted normals with directions are more convenient), I suppose this slight redundancy is okay?

src/compas/geometry/_core/normals.py
```diff
@@ -59,7 +60,7 @@ def normal_polygon(polygon, unitized=True):
     for i in range(p):
         b = polygon[i]
         ob = subtract_vectors(b, o)
-        n = cross_vectors(oa, ob)
+        n = scale_vector(cross_vectors(oa, ob), 0.5)
```
tomvanmele: would be better to add this as part of the return statement instead of doing this multiplication over and over again

tomvanmele:
```python
if not unitized:
    return 0.5 * nx, 0.5 * ny, 0.5 * nz
```

Licini:
@juney-lee the unit test is not passing on this one due to a very small numerical difference, perhaps it's better to apply a precision in the test.

tomvanmele:
yes, indeed. all tests based on geometry should use `allclose` rather than absolute comparisons...

juney-lee:
yeah... that's a lot of decimal points! why would it result in ever-so-slightly different answers from before though? exactly the same calculation... only the last digit is different.

tomvanmele:
@juney-lee doesn't matter we need to change this everywhere in any case...

tests/compas/datastructures/test_mesh.py
```diff
@@ -509,7 +509,7 @@ def test_face_coordinates():
 
 def test_face_normal():
     mesh = Mesh.from_obj(compas.get('quadmesh.obj'))
-    assert mesh.face_normal(0) == [0.5435358481001584, -0.16248515023849733, 0.8235091728584537]
+    assert mesh.face_normal(0) == [0.54353584810015843, -0.1624851502384973, 0.8235091728584536]
```
tomvanmele, line 512: this should be
```python
assert allclose(mesh.face_normal(0), [0.543535, -0.162485, 0.823509])
```

tomvanmele: with the import somewhere at the top of the file of course.

tomvanmele: we should also stop using these hardcoded identifiers that assume some kind of specific ordering of the internal data. better to use a mesh of which we actually know the normals, like a cube...

PR #692, opened 2020-12-08
Unit test for halfface data getter/setter

Licini:
Following the conversation in #676
Since the fix is already pushed to master. This is to just add the unit test. 
PS: I also tested the change in compas_3gs and it works well.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+* Fix bug in `Halfface`'s data getter/setter for json serializing.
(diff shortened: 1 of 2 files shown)


tomvanmele:
i think you are missing a few things...
perhaps i just push the update and the related tests?

Licini:
> i think you are missing a few things...
> perhaps i just push the update and the related tests?

Sure might be better that way.... btw I just found a more annoying issue now, the format of dict output by data getter and accepted by data setter of `halfface` is not yet aligned with others like `halfedge`. Basically still the same with versions prior to 0.17. And weirdly data/json validation on my windows machine ignores all of those. 

 We' better coordinate this a bit with @juney-lee , cuz updating to latest data schema could affect the loading of 3gs session files made by students.

PR #696, opened 2020-12-10
Network neighbourhood bug

duchaoyu:
no need to remove the key in the last iteration of the while loop

### Checklist

### DIFF
--- a/src/compas/datastructures/network/core/graph.py
+++ b/src/compas/datastructures/network/core/graph.py
@@ -1275,8 +1275,6 @@ class Graph(Datastructure):
                 temp += self.neighbors(key)
             nbrs.update(temp)
             i += 1
-        if key in nbrs:
-            nbrs.remove(key)
         return list(nbrs)
 
     def neighbors_out(self, key):
--- a/src/compas/datastructures/network/core/graph.py
+++ b/src/compas/datastructures/network/core/graph.py
@@ -1271,10 +1271,12 @@ class Graph(Datastructure):
             if i == ring:
                 break
             temp = []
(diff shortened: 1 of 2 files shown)


src/compas/datastructures/network/core/graph.py
```diff
@@ -1275,8 +1275,6 @@ def neighborhood(self, key, ring=1):
                 temp += self.neighbors(key)
             nbrs.update(temp)
             i += 1
-        if key in nbrs:
-            nbrs.remove(key)
```
tomvanmele: this line does not remove potential duplicates, but rather the original node which should not appear in the list of its own neighbours...

however, there is another problem related to that that is indeed an actual bug (will let you figure it out :)

duchaoyu: oh... yes... It removes the key in the last iteration of the while loop... the last neighbour key in the (i-1)ring.

tomvanmele: the change should actually be
```python
def neighborhood(self, vertex, ring=1):
    nbrs = set(self.neighbors(vertex))
    ...
    if vertex in nbrs:
        nbrs.remove(vertex)
```

tomvanmele: @duchaoyu ping

tomvanmele: please fix so i can merge...

tomvanmele (approved):
only an entry in the changelog is still missing...

duchaoyu:
hope now it's fine...

Issue #703, opened 2020-12-11
Online documentation contains html dump


Achillx:
**Describe the bug**
The online API documentation page displays HTML code.

**To Reproduce**
Visit [https://compas.dev/compas/latest/api/compas.datastructures.html](https://compas.dev/compas/latest/api/compas.datastructures.html)

**Expected behavior**
Well...no HTML dumps :p

**Screenshots**
![Screenshot 2020-12-11 at 14 48 29](https://user-images.githubusercontent.com/38357732/101910941-f62f0100-3bbf-11eb-8732-5288f33044f8.png)

**Desktop (please complete the following information):**
 - OS: macOS Catalina 10.15.7
 - Chrome: Version 87.0.4280.88 (Official Build) (x86_64)
 - Safari: Version 14.0.1 (15610.2.11.51.10, 15610)

beverlylytle:
Same with geometry https://compas.dev/compas/latest/api/compas.geometry.html

tomvanmele:
this is the inheritance diagram that is somehow not converted to an image by `graphviz`. will have a look...

tetov:
I was looking at this a bit yesterday.

Two things:
* You need to install both python-graphviz and graphviz on the github runnner. `sudo apt-get install graphviz`. 
* Uncomment graphviz extension in `docs/conf.py`. 

Tested locally but not on github action-like environment.

Licini:
> I was looking at this a bit yesterday.
> 
> Two things:
> 
> * You need to install both python-graphviz and graphviz on the github runnner. `sudo apt-get install graphviz`.
> * Uncomment graphviz extension in `docs/conf.py`.
> 
> Tested locally but not on github action-like environment.

Thanks! gonna give a try now on github action

PR #713, opened 2020-12-14
enable graphviz

Licini:
Fixing issue #703


Licini:
should be fixed now @brgcode https://compas.dev/compas/pull_713/api/compas.datastructures.html

PR #717, opened 2020-12-16
packages argument used as addition to discovered plugins

tetov:
Currently giving a list of packages to install to `compas_rhino.install` is mutually exclusive to install discovered `installable_rhino_packages`. I would like to be able to install packages that are currently not able to advertise themselves as plugins without also specifying packages that could be discovered.

A backwards-compatible one.

### Checklist

- ~~I added a line to the `CHANGELOG.md` file in the `Unreleased` section under the most fitting heading (e.g. `Added`, `Changed`, `Removed`).~~

### DIFF
--- a/src/compas_rhino/install.py
+++ b/src/compas_rhino/install.py
@@ -24,8 +24,8 @@ def install(version=None, packages=None):
         The version number of Rhino.
         Default is ``'6.0'``.
     packages : list of str, optional
-        List of packages to install or None to use default package list.
-        Default is ``['compas', 'compas_rhino', 'compas_ghpython']``.
+        List of packages to install in addition to packages available as
+        ``installable_rhino_packages`` plugins.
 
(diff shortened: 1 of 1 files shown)


gonzalocasas:
I'm not sure if this is correct. `python -m compas_rhino.install` is one thing, it is supposed to install all `compas*` plugins that work in Rhino. Calling `python -m compas_rhino.install -p <package>` is a very different thing, and it is supposed to install only those that you pass on, not the union of the parameters plus all discoverable. 

If you need both the discoverable ones plus manually selected ones, you can call the two commands in sequence: `python -m compas_rhino.install && python -m compas_rhino.install -p <package>`

Am I missing something?

tetov:
I somehow didn't think of that. Sorry.

gonzalocasas:
;)

PR #721, opened 2020-12-17
Default to python and not pythonw because it is not automatically installed on Mac

tomvanmele:
…talled on Mac

### Checklist

### DIFF
--- a/src/compas/_os.py
+++ b/src/compas/_os.py
@@ -195,7 +195,7 @@ def select_python(python_executable):
                 return python
 
     # Assume a system-wide install exists
-    return python_executable or 'pythonw'
+    return python_executable or 'python'
 
 
 def prepare_environment(env=None):
--- a/src/compas/_os.py
+++ b/src/compas/_os.py
@@ -194,8 +194,10 @@ def select_python(python_executable):
             if os.path.exists(python):
                 return python
 
+    default_exe = 'pythonw' if is_windows() else 'python'
+
     # Assume a system-wide install exists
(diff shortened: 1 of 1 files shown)


tetov:
Last night I was thinking of opening a issue about this since `pythonw.exe` is also missing on linux. Haha!

I'll test this out a bit too.

gonzalocasas:
I'm not really sure, this is something that we have tested quite a bit now, changing it right before the release feels wrong. Is `pythonw` installed with `python.app` package?
Would it be safer (or possible) to try to detect if pythonw is installed and default to it and only default to python isle it's not detected?

tomvanmele:
yes but then we have to make more changes. for example `PYTHON_DIRECTORY` would have to be populated with `os.getenv('CONDA_PREFIX')` if the bootstrapper is not available. otherwise it will default to the system python...

gonzalocasas:
An alternative would be to default to `pythonw` on Windows only, and default to `python` for any other platform.

EDIT: the downside of using `python` on Windows is that a black window will pop up while the call is executing, which in the case of the RPC server is basically up until you close Rhino or explicitly shutdown the server.

tomvanmele:
ok will push updates accordingly...

tetov:
Now I can start proxy implicitly in WSL too.

[Example](https://github.com/gramaziokohler/compas_mobile_robot_reloc/blob/bump/tests/test_proxy.py)

gonzalocasas (approved):
LGTM!

PR #731, opened 2021-01-07
separate origin from current_origin

beverlylytle:
`compas.robots.RobotModel` uses and modifies the attribute `origin` of its `Joint`s, losing the information necessary for recreating the associated URDF.  This PR is to separate `origin` (that which should be included in a URDF) from `current_origin` (that which is used for calculating and visualizing forward kinematics).  This is somewhere between a bug fix and a new feature.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 * Fix Rhino7 Mac installation path
-* Separate `compas.robots.Joint.origin` into the parent-relative origin `origin` and the world-relative origin `current_origin`.
+* Separate `compas.robots.Joint.origin` into the static parent-relative `origin` and the dynamic world-relative `current_origin`.
+* Separate `compas.robots.Joint.axis` into the static parent-relative `axis` and the dynamic world-relative `current_axis`.
 
 ### Removed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+* Separate `compas.robots.Joint.origin` into the parent-relative origin `origin` and the world-relative origin `current_origin`.
+
 ### Removed
 
 
--- a/src/compas/robots/model/joint.py
+++ b/src/compas/robots/model/joint.py
@@ -181,7 +181,6 @@ class Dynamics(Base):
             'friction': self.friction,
         }
         attributes.update(self.attr)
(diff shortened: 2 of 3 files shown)


src/compas/robots/model/joint.py
```diff
@@ -181,7 +181,6 @@ def get_urdf_element(self):
             'friction': self.friction,
         }
         attributes.update(self.attr)
-        attributes = dict(filter(lambda x: x[1], attributes.items()))
```
beverlylytle, line 184: This line would result in many joints have attribute-less dynamics tags `<dynamics />` which is not allowed, at least by pybullet.  So now they look like `<dynamics  friction="0.0" damping="0.0" />`.  Much easier than trying to eliminate the tag completely if it doesn't have non-default attributes.

src/compas/robots/model/joint.py
```diff
@@ -609,6 +608,7 @@ def __init__(self, name, type, parent, child, origin=None, axis=None, calibratio
         self.attr = kwargs
         self.child_link = None
         self.position = 0
+        self.current_origin = Origin(self.origin.point, self.origin.xaxis, self.origin.yaxis)
```
gonzalocasas: any reason why not use `self.origin.copy()`?

src/compas/robots/model/joint.py
```diff
@@ -698,10 +695,8 @@ def transform(self, transformation):
         -------
         None
         """
-        if self.origin:
-            self.origin.transform(transformation)
-        if self.axis:
-            self.axis.transform(transformation)
+        self.current_origin.transform(transformation)
+        self.axis.transform(transformation)
```
gonzalocasas: Shouldn't we do the same for `self.axis`?

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 * Fix Rhino7 Mac installation path
+* Separate `compas.robots.Joint.origin` into the parent-relative origin `origin` and the world-relative origin `current_origin`.
```
gonzalocasas: I like the `parent-relative vs world-relative` wording, maybe add as inlined comment at its first occurrence.

src/compas/robots/model/joint.py
gonzalocasas: I was wondering what this line was doing :P

gonzalocasas (commented):
looks good! only added a couple of comments

src/compas/robots/model/joint.py
beverlylytle: no

src/compas/robots/model/joint.py
beverlylytle: yes

PR #735, opened 2021-01-15
add methods remove_link, remove_joint

beverlylytle:
These new methods on RobotModel are necessary for the fix to the `add_attached_collision_mesh` method of the pybullet client ([#259](https://github.com/compas-dev/compas_fab/pull/259)) .  This PR will remain as a draft until after the release of version 1.0.  Remind me to update the CHANGELOG after that.

### Checklist

### DIFF
--- a/src/compas/robots/model/robot.py
+++ b/src/compas/robots/model/robot.py
@@ -591,7 +591,6 @@ class RobotModel(Base):
 
                     if not shape.geometry:
                         raise Exception('Unable to load geometry for {}'.format(shape.filename))
-        self._geometry_loaded = True
 
     def ensure_geometry(self):
         """Check if geometry has been loaded.
@@ -600,9 +599,12 @@ class RobotModel(Base):
         :exc:`Exception`
             If geometry has not been loaded.
         """
-        if not self._geometry_loaded:
-            raise Exception(
-                'This method is only callable once the geometry has been loaded.')
+        for link in self.links:
+            for element in itertools.chain(link.collision, link.visual):
+                shape = element.geometry.shape
+                if not shape.geometry:
+                    raise Exception(
+                        'This method is only callable once the geometry has been loaded.')
 
     @property
     def frames(self):
--- a/src/compas/robots/model/robot.py
+++ b/src/compas/robots/model/robot.py
@@ -1012,10 +1012,10 @@ class RobotModel(Base):
             The name of the joint
         """
         joint = self.get_joint_by_name(name)
-        self.joints = [joint for joint in self.joints if joint.name != name]
+        self.joints = [j for j in self.joints if j.name != name]
         parent_link = self.get_link_by_name(joint.parent.link)
-        parent_link.joints = [joint for joint in parent_link.joints if joint.name != name]
-        self._adjacency[parent_link.name] = [joint.name for joint in parent_link.joints]
+        parent_link.joints = [j for j in parent_link.joints if j.name != name]
+        self._adjacency[parent_link.name] = [j.name for j in parent_link.joints]
         del self._links[joint.child.link]
         del self._joints[name]
         del self._adjacency[name]
--- a/src/compas/robots/model/robot.py
+++ b/src/compas/robots/model/robot.py
@@ -68,6 +68,7 @@ class RobotModel(Base):
         self._rebuild_tree()
         self._create(self.root, Transformation())
         self._scale_factor = 1.
+        self._geometry_loaded = False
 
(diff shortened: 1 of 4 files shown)


gonzalocasas:
> Remind me to update the CHANGELOG after that.

@beverlylytle ping ;)

tests/compas/robots/test_model.py
```diff
@@ -69,6 +69,9 @@ def test_programmatic_robot_model():
     urdf = URDF.from_robot(robot)
     robot_reincarnated = RobotModel.from_urdf_string(urdf.to_string())
     assert(['link0', 'joint1', 'link1', 'joint2', 'link2'] == list(robot_reincarnated.iter_chain()))
+    robot.remove_link('link2')
+    robot.remove_joint('joint2')
+    assert ['link0', 'joint1', 'link1'] == list(robot.iter_chain())
```
gonzalocasas, line 79: it would be good to add another unit tests that removes a link/joint from the middle of the chain as well.

gonzalocasas (approved):
Looks good, but I have one important question: don't we need to re-calculate the chain of transforms (as done in `_create`) when a joint in removed?

src/compas/robots/model/robot.py
```diff
@@ -575,6 +591,18 @@ def load_geometry(self, *resource_loaders, **kwargs):
 
                     if not shape.geometry:
                         raise Exception('Unable to load geometry for {}'.format(shape.filename))
+        self._geometry_loaded = True
```
beverlylytle: Using this flag isn't a good idea in a couple of cases.  First if the robot is defined purely using primitives, then `load_geometry` wouldn't need to be called.  Secondly, if a user adds a link manually (that is, not with `add_link`, but with creating a `Link` instance, etc) and gives its collision and visual elements a `filename` rather than a mesh (as unlikely as this is), this could lead to mysterious errors being raised.  So I think I need to essentially copy the nested for loops into `ensure_geometry`.

gonzalocasas: Yes, agree with the inconvenience of the flag. The `ensure_*` methods are basically asserts, so, if there's the risk of stale state, then the nested loop suddenly feels like a good option ;)

beverlylytle:
> don't we need to re-calculate the chain of transforms (as done in `_create`) when a joint in removed?

`_create` only effects descendants of the root, and one joint's effect on the transforms of a robot in the recursive nature of `_create` is only felt by the joint's descendants, right?  so if a joint is removed, all of its descendants are no longer connected to the root, and the transforms of the ancestors of the removed joint remain the same.  as far as i understand there's no harm in calling `_create` again on the root except for the repeated calculations, so do you mean to call `_create` on these possibly existing disconnected bits of robot (for example in the case of the removal the fixed joint in an attached robotool)?

gonzalocasas:
> all of its descendants are no longer connected to the root

ah! sorry! you're right. It's only removing the intermediate link, but not reconnecting the chain. Then it is fine as it is.

beverlylytle:
> > all of its descendants are no longer connected to the root
> 
> ah! sorry! you're right. It's only removing the intermediate link, but not reconnecting the chain. Then it is fine as it is.

well, this does suggest one question.  i implemented `remove_joint` to do the opposite of `add_joint` in that it removes the last references in a `RobotModel` to the joint's child link.  should it do this recursively to all descendants of the joint? that is, if you remove a joint in the middle of the tree, should it remove the entire subtree? or do we leave it to the user to maintain a proper connected tree structure?  EDIT: of course `iter_links` and `iter_joints` don't care about the disconnected bits, but there could be some dangling links and joints in `_links` and `_joints`, resp.  So `to_urdf` would produce a probably invalid file. And `get_link_by_name` could return a disconnected link.  However, if the user "irresponsibly" uses `add_link` without the appropriate `add_joint` followup, the robotmodel might end up in the same sort of disconnected state.  So, I'm disinclined to make `remove_joint` and `remove_link` anything more than the opposite of `add_joint` and `add_link`.

src/compas/robots/model/robot.py
```diff
@@ -68,6 +68,7 @@ def __init__(self, name, joints=[], links=[], materials=[], **kwargs):
         self._rebuild_tree()
         self._create(self.root, Transformation())
         self._scale_factor = 1.
+        self._geometry_loaded = False
```
gonzalocasas: This is not used anymore, is it?

beverlylytle: ah, nope

PR #736, opened 2021-01-18
Pylance support through type stubs

### LINKED ISSUE #621: compas imports - work with pylance intellisense
# Feature Request

[Pylance ](https://marketplace.visualstudio.com/items?itemName=ms-python.vscode-pylance)is a feature rich python language plugin for VSCode. It provide code analysis and [intellisense](https://code.visualstudio.com/docs/editor/intellisense). 

The intellisense function works with python default functions and types and almost all other libraries I used (such as numpy). However it does not work with compas.

I suspect this is caused by the 2-layer-deep import routing that compas uses. If I write the full path, or in fact keep the default full path that Pylance automatically creates, the intellisense works, See below where `intersection_line_plane` is yellow, that one works.

I think @Achillx has faced similar problem.

tomvanmele:

### Checklist


tomvanmele:
Adding type annotations will have to happen at some point anyway and it helps solve the PyLance problem (at least partially). There are however a few things to be aware of.

* Adding type annotations in a way that is compatible with Python 2.7 (Rhino) is more elegant with stub files than by adding them in code comments.
* Adding the stub files at the "second import level" helps strengthen the definition of an API.
* Currently "Goto Definition" in VS Code will only load the stub file and not the original source file.
* Auto-import on the other hand will still use the fully qualified path (so not the path defined in the API)
* There seems to be no way to automatically copy the docstring into the sub files in a way that PyLance can read and understand.

All of that to say that this is still very much a WIP PR...

tomvanmele:
@yck011522

yck011522:
@brgcode 
Thanks a lot for addressing byhe pylance problem together with the type annotation issue. I'll try it out later. 

May I ask from the developer's workflow, is the pyi file manually put together? In another words. When a function is added, a corresponding entry have to be added in the pyi file? Thanks.

tomvanmele:
yes at the moment there doesn't seem to be a good way to autogenerate stub files with meaningful annotations. autogenerated files basically accept `Any` and return `Any`.

this will be a bit painful at first because we have to start from scratch.
however, adding the type annotations directly in the code or in stub files instead seems like the same amount of work.

also, currently the "second level" API has to be defined manually by listing API items in the `__all__` variables of the modules. so we just have to find a way to not have to do it twice...

tomvanmele:
@gonzalocasas @beverlylytle this is now ready for review. not implemented for the cad packages yet though...

PR #744, opened 2021-01-20
Robot update fix

beverlylytle:
Say you have a `RobotModel` that has part of its geometry defined by primitives (box, etc) and part of its geometry defined by meshes as in [this urdf](https://github.com/ros/urdf_tutorial/blob/master/urdf/05-visual.urdf).  Then, without loading the geometry, the artist can draw the robot (that is, draws the primitives but ignores the unloaded meshes), but if `update` is called, an error is raised.  This PR is to allow error-free calls to `update`  to robots without the geometry loaded.  

### Checklist


src/compas/robots/base_artist/_artist.py
```diff
@@ -217,8 +217,11 @@ def _apply_transformation_on_transformed_link(self, item, transformation):
         -------
         None
         """
-        relative_transformation = transformation * item.current_transformation.inverse()
-        for native_geometry in item.native_geometry:
+        if getattr(item, 'current_transformation'):
+            relative_transformation = transformation * item.current_transformation.inverse()
+        else:
+            relative_transformation = transformation
```
gonzalocasas, line 223: I don't totally get the effect of this change. Is this an "no links have geometry loaded vs all links have it" situation? Or could it happen that some have it and some others not? If that is the case, does it mean that the link that is not yet loaded will be skipped in the transformation chain? Doesn't that lead to inconsistent results?

beverlylytle: I'm not sure I totally follow what you are asking.  I found myself in a situation where some links had geometry given by primitives and others by meshes, and these meshes had not been loaded.  I don't know what you mean by transformation chain... I will step through the relevant part of the life cycle of a `current_transformation` not to man-splain, but to make sure we are both on the same page.

 A visual or collision item's `current_transformation` is something that is initalized to `None`, and then when `create` is called in an artist, it is set to the identity transformation if there happens to be a mesh loaded or primitive geometry present.  When `update` is called,  transformations are calculated based on a joint state and the joint `current_origin`s, and then these transformations are applied to the geometry of a joint's child link, here with `_apply_transformation_on_transformed_link`.  In summary, the transformation chain (am I using this term correctly?) is maintained by the joint's `current_origin/transformation`, and the `current_transformation` of the collision and visual items of a link is computed from that.  So, actually, instead of what this PR has currently, it might make more sense just to initialize `Collision/Visual.current_transformation` to `Transformation()`, because I'm not sure that there is ever this distinction made between `None` and something for this attribute.

actually, what about the case where the visual meshes are complex meshes, and the collisions are just bounding boxes (or a collection of boxes that bound) and you only want to visualize the collision geometry and not load the complicated stuff?  i think this would be a good use case that would break the current implementation.

gonzalocasas (approved):

PR #749, opened 2021-01-21
Correct exception type

gonzalocasas:
Changed the exception type from `NotImplementedError` to `ValueError` since that the method is actually implemented, but the error mode is that the argument is an invalid one (which fits the description of what `ValueError` is).

### Checklist


gonzalocasas:
We have not decided (afaik) whether we consider a changing exception type breaking change or not. In a typed language, this would definitely be. In python, and in this particular case, I'd say it is non breaking.

tomvanmele (approved):

PR #762, opened 2021-02-02
Allow to set a robot's base (RCF)

gonzalocasas:
This adds a property RCF to the robot model (`model.rcf`) to allow changing the base frame of the robot. Since operation is potentially quite involved once the model is used in `COMPAS FAB`, I will need to add an event emitter or some way in which this model can be observed from an external class and hook into changes (like rcf change) so that those are propagated to backends. I'm still not sure how to do this best so, I'm opening this a draft for comments already now and I will update soon.

### Checklist


src/compas/robots/model/robot.py
```diff
@@ -550,8 +565,78 @@ def get_base_link_name(self):
...
+        # 2) or independently of the model of joints, as a reference frame
+        # The 1) takes precedence over the 2) if both are used.
+        if not self.root or not self._joints:
+            self._rcf = frame.copy()
+            return
+
+        base_link = self.get_link_by_name(self.get_base_link_name())
+
+        # if a fixed joint exists, set its origin based on the specified frame
+        if base_link and base_link.parent_joint and base_link.parent_joint.type == Joint.FIXED:
```
beverlylytle: This check is used a couple of times.  Maybe it should be pulled out to `Joint` and named something better than `is_fixed_child`.

src/compas/robots/model/joint.py
```diff
@@ -608,10 +613,28 @@ def __init__(self, name, type, parent, child, origin=None, axis=None, calibratio
...
+        return self._origin
+
+    @property
+    def axis(self):
+        return self._axis
+
+    @origin.setter
+    def origin(self, value):
+        self._origin = value
+        # The following is a world-relative frame representing the origin, which change with
```
beverlylytle:
```suggestion
        # The following is a world-relative frame representing the origin, which changes with
```

src/compas/robots/model/joint.py
```diff
@@ -608,10 +613,28 @@ def __init__(self, name, type, parent, child, origin=None, axis=None, calibratio
...
+        # the joint state, while `origin` is parent-relative and static.
+        self.current_origin = value.copy()
+
+    @axis.setter
+    def axis(self, value):
+        self._axis = value
+        # The following is a world-relative axis which change with
```
beverlylytle:
```suggestion
        # The following is a world-relative axis which changes with
```

src/compas/robots/model/joint.py
```diff
@@ -608,10 +613,28 @@ def __init__(self, name, type, parent, child, origin=None, axis=None, calibratio
...
+    @property
+    def axis(self):
+        return self._axis
+
+    @origin.setter
+    def origin(self, value):
+        self._origin = value
+        # The following is a world-relative frame representing the origin, which change with
+        # the joint state, while `origin` is parent-relative and static.
+        self.current_origin = value.copy()
```
beverlylytle, line 498: maybe i haven't properly thought this through, but should this be 
`self.current_origin = self.current_origin @ self.origin.inverse() @ value`
where i'm ignoring the distinction between a frame and its transformation

src/compas/robots/model/joint.py
```diff
@@ -608,10 +613,28 @@ def __init__(self, name, type, parent, child, origin=None, axis=None, calibratio
...
+
+    @axis.setter
+    def axis(self, value):
+        self._axis = value
+        # The following is a world-relative axis which change with
+        # the joint state, while `axis` is parent-relative and static.
+        self.current_axis = value.copy()
```
beverlylytle, line 505: And similarly here, should we not 'remember' the transformations that have already been applied to the `current_axis`?

src/compas/robots/model/robot.py
```diff
@@ -550,8 +565,78 @@ def get_base_link_name(self):
...
+            A transformation describing the root of the chain.
+        """
+        # Uninitialized robot, default to identity matrix
+        if not self.root or not self._joints:
+            return Transformation()
+
+        base_link = self.get_link_by_name(self.get_base_link_name())
+
+        # if there's no fixed joint and there is an RCF is explicitly set, return it as a transform
+        ref_frame = Frame.worldXY()
```
beverlylytle: I don't understand.  Say you've got a robot with a proper base link.  Then you set `robot.rfc = <some frame>`.  This will change the origin of the base link.  But if then you access `robot.rcf` you'll get the identity transformation in return.  Does that make sense?  Is that supposed to make sense?

src/compas/robots/model/robot.py
```diff
@@ -550,8 +565,78 @@ def get_base_link_name(self):
...
+
+        Returns
+        -------
+        :class:`compas.geometry.Frame`
+            Robot coordinate frame, usually placed at the base of the robot and attached to a ``world`` link.
+        """
+        return Frame.from_transformation(self._root_transformation)
+
+    @rcf.setter
+    def rcf(self, frame):
```
beverlylytle, line 598: There generally seems to be a lot of repeated code between the `rcf` setter and the `_root_transformation`.  Is there a nice way to compress these things?

src/compas/robots/model/robot.py
gonzalocasas: nope, the comment might be somewhat misleading if it's right above this line, but `Frame.worldXY()` is the default if the `if` right after is not true. In the case where you set the `rcf`, then the `if` goes in, and the `ref_frame` is that one instead.

beverlylytle: i think i didn't express myself well.  consider the following code:
```
import math
import compas_fab
from compas.robots import LocalPackageMeshLoader, RobotModel, Frame


urdf_filepath = compas_fab.get('universal_robot/ur_description/urdf/ur5.urdf')
loader = LocalPackageMeshLoader(compas_fab.get('universal_robot'), 'ur_description')
model = RobotModel.from_urdf_file(urdf_filepath)

assert model.rcf == Frame.worldXY()
model.rcf = Frame.from_euler_angles([math.pi, 0, 0], point=[1,0,0])
assert model.rcf == Frame.worldXY()
```
Without knowing the internals of `rcf`, that doesn't make any sense to the user.

beverlylytle:
> I will need to add an event emitter or some way in which this model can be observed from an external class and hook into changes (like rcf change) so that those are propagated to backends.

It will also need to be propagated to the `RobotModelArtist`.

src/compas/robots/model/robot.py
gonzalocasas: You're totally right. I'll fix.

gonzalocasas: This should be now fixed (https://github.com/compas-dev/compas/pull/762/commits/a4a3e8f86bfd3d99c4e14f781cfb39ccebd00bab) along with simpler code path to determine what happens where (and a bit less duplication)

src/compas/robots/model/robot.py
gonzalocasas: also done at https://github.com/compas-dev/compas/pull/762/commits/a4a3e8f86bfd3d99c4e14f781cfb39ccebd00bab (to some extent at least)

src/compas/utilities/decorators.py
```diff
@@ -136,9 +136,129 @@ def wrapper(*args, **kwargs):
...
+    if not hasattr(event_source, '__event_observers__'):
+        return
+
+    event_key = _get_event_key(event_source, event_name)
+    event_observers = event_source.__event_observers__.get(event_key)
+
+    if event_observers:
+        event_observers.remove(callback)
+    if not event_observers:
+        del event_source.__event_observers__[event_key]
```
beverlylytle: I see a lot of potential for exceptions being raised with this function.  If there is a typo in `event_name`, there will be a `KeyError`.  If the wrong callback is passed there could be a `ValueError`.  Does it make sense to raise custom exceptions for this, rather than the generic and perhaps mysterious ones?

src/compas/utilities/decorators.py
```diff
@@ -136,9 +136,129 @@ def wrapper(*args, **kwargs):
...
+
+    Notes
+    -----
+    This decorator should only be applied to methods and properties
+    of a class, not to stand-alone functions.
+
+    Examples
+    --------
+    >>> class Test(object):
+    ...   @observable(event_name='init')
```
beverlylytle: I trust myself to type 'init' correctly, but I do not trust myself to type 'updated_transformations' correctly.  I know this is what tests are for, but somethings are difficult to test (say, did the image of the robot  in blender update correctly when the rcf changed?). Is there a nice way of throwing an exception somewhere if the observable event names for a class don't match up with the observed event names or vice versa? or is that something that shouldn't be protected against?

beverlylytle:
I just noticed that `compas_fab.robots.Robot` has the methods `set_RCF` (throwing a NotImplementedError) and `get_RCF`.  I suppose those should be updated in a further PR.

jf---:
@gonzalocasas should this PR move to `compas_robots` or be closed? Seems valuable work, but quite stalled

yck011522:
@gonzalocasas I guess we have decided to shelf this idea for the time being. How about we move it to compas_robots first and manage it from there?

yck011522:
I know we said that we will not work on this issue soon, but just to collect one more argument for developing this function, is that the ability to set the robot base (to be different from World frame), can easily fix the orientation of a UR base. 
https://github.com/ros-industrial/universal_robot/issues/676

PR #764, opened 2021-02-04
hide meshes on instantiation of RobotModelArtist

### LINKED ISSUE #760: RobotModelArtist.draw_visual for blender draws visual and collision meshes
**Describe the bug**
When using `compas_blender.artists.RobotModelArtist.draw_visual`, all meshes, both visual and collision, are added to the scene.  Same is true for `draw_collision`.  However, this is not the case for the corresponding methods in `compas_rhino`.  

**To Reproduce**
1. In Blender, run the following script: 
 ```
import compas
from compas.robots import GithubPackageMeshLoader
from compas.robots import RobotModel
import compas_blender
from compas_blender.artists import RobotModelArtist, BaseArtist

compas_blender.clear()

compas.PRECISION = '12f'
github = GithubPackageMeshLoader('ros-industrial/abb', 'abb_irb6600_support', 'kinetic-devel')
model = RobotModel.from_urdf_file(github.load_urdf('irb6640.urdf'))
model.load_geometry(github)

artist = RobotModelArtist(model, layer='COMPAS FAB::Example')

artist.draw_visual()
```

**Expected behavior**
I expect only visual meshes to be added to the scene, when in fact collision meshes are also added.

 - OS: MacOSX
 - Python version 3.7.8
 - Python package manager conda

beverlylytle:
Closes #760.  I also renamed one horribly named internal method as well as a little tidying.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Added ability of `RobotModel.add_link` to accept primitives in addition to meshes.
 * Fixed bug regarding the computation of `Joint.current_origin`.
 * Fixed bug regarding a repeated call to `RobotModel.add_joint`.
+* Fixed bug in :class:`compas_blender.RobotModelArtist.update`.
 
 ### Removed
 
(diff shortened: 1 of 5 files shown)


beverlylytle:
I was tempted to change the name of the methods `draw_geometry` to something like `construct_native_geometry` because in neither the rhino nor the gh robot artist do the methods have a visual result, and if this pr is accepted, then that would be true of the blender artist too.

gonzalocasas (approved):
I think it is good to go 👍   

About the larger method renaming you mentioned: go for it, it sounds good! Either `construct_geometry` or `create_geometry` (which maps nicely to the `_create` methods. But if you rather leave it for a different PR, that's alright as well.

Eventually, we should aim at reshuffling things around in the Blender artists so that the creation of the meshes doesn't happen until it's actually needed (rather than doing it always and then only showing some of it) but that definitely doesn't need to happen now.

beverlylytle:
I will hold off on merging this until #765 is finished.

Issue #759, opened 2021-01-25
blender Robotmodelartist does not create collections correctly


robin-gdwl:
The function [draw_geometry](https://github.com/compas-dev/compas/blob/416e490caf4bfb11a47d8e208d19f04e1d76eb16/src/compas_blender/artists/robotmodelartist.py#L30) from the [compas_blender.artists.RobotModelArtist ](https://github.com/compas-dev/compas/blob/master/src/compas_blender/artists/robotmodelartist.py) creates a new collection every time it is called. 
Since the method [create()](https://github.com/compas-dev/compas/blob/416e490caf4bfb11a47d8e208d19f04e1d76eb16/src/compas/robots/base_artist/_artist.py#L117) of the base_artist calls draw_geometry for every mesh in the urdf file the blender artist creates a new collection each time (at least thats how I understand it).
In addition: the mesh is also not put into the collection that was created. 

Location of the draw_geometry call in the base artist: 
https://github.com/compas-dev/compas/blob/416e490caf4bfb11a47d8e208d19f04e1d76eb16/src/compas/robots/base_artist/_artist.py#L153

blender Robotmodelartist method:
https://github.com/compas-dev/compas/blob/416e490caf4bfb11a47d8e208d19f04e1d76eb16/src/compas_blender/artists/robotmodelartist.py#L42-L44

A possible solution could be to simply check if a collection with the name defined in [self.layer](https://github.com/compas-dev/compas/blob/416e490caf4bfb11a47d8e208d19f04e1d76eb16/src/compas_blender/artists/robotmodelartist.py#L24) already exists and use that one instead of creating a new one. 

It appears to me that the blender robotartist is at an early stage in the developement. I might implement a fix for this and create a pull request if a rewrite of the robotartist is not already under way. 
![image](https://user-images.githubusercontent.com/25942644/105737935-cb61f780-5f36-11eb-8c93-73de3f25a9c2.png)

example script tested in Blender 2.9.1 and compas 1.0.0 on Windows 10: 
```
import compas
from compas.robots import GithubPackageMeshLoader
from compas.robots import RobotModel
import compas_blender
from compas_blender.artists import RobotModelArtist, BaseArtist

compas_blender.clear()
#compas_blender.draw()

compas.PRECISION = '12f'
github = GithubPackageMeshLoader('ros-industrial/abb', 'abb_irb6600_support', 'kinetic-devel')
model = RobotModel.from_urdf_file(github.load_urdf('irb6640.urdf'))
[...]

PR #765, opened 2021-02-05
Fix for Multiple collections created by Blender RobotmodelArtist. 

robin-gdwl:
Simple fix for the blender RobotmodelArtist. 
Responds to this Issue:  #759 

## Changed:

* Fixed bug where initialising a `compas_blender.artists.Robotmodelartist` would create a new collection for each mesh and then also not put the mesh iton the created collection. 
* Changed the initialisation of `compas_blender.artists.Robotmodelartist` to include a `collection`-parameter instead of a `layer`-parameter to be more consistent with Blender's nomenclature. 
* Used a utility function from `compas_blender.utilities` to create the collection if none exists instead of using a new call to a bpy-method. 

- [ ] Breaking change: bug fix or new feature that involve incompatible API changes. **not entirely sure about this one**

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,24 +5,13 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## Unreleased 2021-02-05
-
-### Added
-* Added Blender Python-example to the documentation section: Tutorials -> Robots
-
-### Changed
-
-* Fixed bug where initialising a `compas_blender.artists.Robotmodelartist` would create a new collection for each mesh and then also not put the mesh iton the created collection. 
-* Changed the initialisation of `compas_blender.artists.Robotmodelartist` to include a `collection`-parameter instead of a `layer`-parameter to be more consistent with Blender's nomenclature. 
-* Used a utility function from `compas_blender.utilities` to create the collection if none exists instead of using a new call to a bpy-method. 
-
-### Removed
 
 ## Unreleased
 
 ### Added
 
 * Added `RobotModel.remove_link`, `RobotModel.remove_joint`, `RobotModel.to_urdf_string`, and `RobotModel.ensure_geometry`.
+* Added Blender Python-example to the documentation section: Tutorials -> Robots
 
 ### Changed
 
@@ -34,6 +23,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Fixed bug regarding the computation of `Joint.current_origin`.
 * Fixed bug regarding a repeated call to `RobotModel.add_joint`.
 * Fixed bug in `compas.datastructures.mesh_slice_plane`.
+* Fixed bug where initialising a `compas_blender.artists.Robotmodelartist` would create a new collection for each mesh and then also not put the mesh iton the created collection. 
+* Changed the initialisation of `compas_blender.artists.Robotmodelartist` to include a `collection`-parameter instead of a `layer`-parameter to be more consistent with Blender's nomenclature. 
+* Used a utility function from `compas_blender.utilities` to create the collection if none exists instead of using a new call to a bpy-method. 
 
 ### Removed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,19 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased 2021-02-05
+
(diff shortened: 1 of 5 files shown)


src/compas_blender/artists/robotmodelartist.py
```diff
@@ -20,15 +20,17 @@ class RobotModelArtist(BaseRobotModelArtist):
         The name of the layer that will contain the robot meshes.
     """
 
-    def __init__(self, model, layer=None):
-        self.layer = layer
+    def __init__(self, model, collection=None, layer=None):
+        self.view_layer = layer
+        self.collection = collection
```
tomvanmele, line 25: the term `layer` was introduced here to be consistent with Rhino. internally it is supposed to be used as the name for the collection. if that is confusing we should change this to `collection` for all blender artists...

tomvanmele: actually i need to take that back. 

it seems we already removed this from the other artist in favor of implicitly named collections. perhaps this would be a good time to decide on a global strategy

beverlylytle: I think it makes more sense to use `collection` and that `layer` should be removed from `RobotModelArtist`.

robin-gdwl: It would be great if (in the future) one could port a script from Rhino to Blender just by changing the import statement, this would only work if all Robotmodelartists use the same parameter names. However using the word "layer" in the Blender context is generally confusing.

tomvanmele: agreed, but perhaps the solution to this is that we take a step back and use our own conventions, rather than transplanting the conventions from one system onto another. but i imagine that will be part of a lengthy discussion :)

CHANGELOG.md
```diff
@@ -5,6 +5,19 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased 2021-02-05
```
beverlylytle: I'm pretty sure `bump2version` will behave quite badly having 2 "Unreleased" sections, so I think it would be best if this section would be combined with the "Unreleased" section below.

docs/tutorial/robots.rst
```diff
@@ -167,8 +188,45 @@ it can be used as a scaffolding for code to be run in a Python script editor wit
...
+
+    compas.PRECISION = '12f'
+    # Load the urdf-file from Github 
+    github = GithubPackageMeshLoader('ros-industrial/abb', 'abb_irb6600_support', 'kinetic-devel')
+    model = RobotModel.from_urdf_file(github.load_urdf('irb6640.urdf'))
+    model.load_geometry(github)
+    
+    # Load the robot meshes into the blender scene
+    artist = RobotModelArtist(model, collection='COMPAS FAB::Example')
+
```
beverlylytle: I think the complexity of the `compas_blender` `RobotModelArtist` example should match the complexity of the `compas_rhino` one in this section.  Comparing the two, it makes the blender version seem way more complicated and a bit intimidating to complete beginners in compas/robotics/blender.  Maybe this complete example could be moved to later in the tutorial.

docs/tutorial/robots.rst
```diff
@@ -167,8 +188,45 @@ it can be used as a scaffolding for code to be run in a Python script editor wit
     artist.clear_layer()
     artist.draw_visual()
 
+.. raw:: html
+
+    </div>
+    <div class="tab-pane" id="visualise_robot_blender">
```
beverlylytle, line 194: I like the html tabs!

beverlylytle (changes requested):
Looks good!  The documentation is nice, especially the formatting.  And now the Blender artist will behave as expected.  Thanks!  Just one or two comments (in particular, with the `CHANGELOG.md`) to address, and we can merge.

beverlylytle:
Another review comment:  I noticed that only one check was run because the default branch name changed since your fork last pulled.  So it would be good if you could merge the main branch into your fork.

CHANGELOG.md
robin-gdwl: I misread how the CHANGELOG is supposed to be structured, will fix that.

docs/tutorial/robots.rst
robin-gdwl: Agreed, I will discard the Github import and make it more similar to the Rhino example.

tomvanmele:
this ready to be merged?

beverlylytle:
> this ready to be merged?

I think not.  @robin-gdwl said he would make one more change to the robot tutorial.

robin-gdwl:
Sorry for the wait, I was creating a section about importing urdf-files and got stuck because importing into Rhino had issues with the unit scale. 
Then I remembered that importing URDF-files is already covered quite well in the same tutorial a few lines down....

beverlylytle (approved):
Looks good!  Let's get this merged!

PR #768, opened 2021-02-08
add unload_modules to compas_blender

beverlylytle:
As a developer working in Blender I want a utility function similar to `compas_rhino.unload_modules` so that I can quickly iterate changes in the source code (without quitting and restarting Blender).  This pull request is to address that.  It is the same as the `compas_rhino` `unload_modules` except that the filter object is cast to a list because Blender complains that the dictionary `sys.modules` changes size while being iterated over.  While there is a docstring I didn't add it to the documentation, mimicking what exists for `compas_rhino`.

### Checklist


src/compas_blender/utilities/misc.py
```diff
@@ -0,0 +1,34 @@
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
```
tomvanmele: can be removed. only necessary for Rhino because of Python 2.7...

tomvanmele (approved):
LGTM. just the py27 stuff...

src/compas_blender/utilities/misc.py
beverlylytle: oh, of course.  old habits die hard

PR #770, opened 2021-02-12
Grasshopper GHPY libraries install

romanarust:
On COMPAS FAB we started to make Grasshopper components that are compiled in a library (.GHPY) (https://github.com/compas-dev/compas_fab/pull/249).
For installing that library, it needs to be copied into the Grasshopper library path. This action is ideally executed when calling 
`python -m compas_rhino.install`

As suggested by @gonzalocasas, compas core should know how to install GH components, and other packages only need to provide the component locations.

For the time being, we're providing two extension points (`pluggable` interfaces) to allow hooking up into the install/uninstall and do arbitrary stuff. This will be a good solution until we sort out all the details of the installation (and potentially compilation) of GH components on end-user machines.

### Checklist

### DIFF
--- a/src/compas_rhino/install.py
+++ b/src/compas_rhino/install.py
@@ -140,7 +140,7 @@ def _run_post_execution_steps(steps_generator):
                 status = 'OK' if success else 'ERROR'
                 print('   {} {}: {}'.format(package.ljust(20), status, message))
             except ValueError:
-                post_execution_errors.append(ValueError('Step successful, but result is wrongly formatted: {}'.format(str(result))))
+                post_execution_errors.append(ValueError('Step successful, but result is wrongly formatted: {}'.format(str(item))))
 
     if post_execution_errors:
         print()
--- a/src/compas_rhino/install.py
+++ b/src/compas_rhino/install.py
@@ -12,7 +12,12 @@ import compas_rhino
 import compas._os
 import compas.plugins
 
-__all__ = ['install']
+__all__ = [
+    'install',
+    'installable_rhino_packages',
+    'after_rhino_install',
+    'after_rhino_uninstall',
+]
 
 
 def install(version=None, packages=None):
@@ -84,8 +89,13 @@ def install(version=None, packages=None):
     symlinks = [(link['source_path'], link['link']) for link in symlinks_to_install]
     install_results = compas._os.create_symlinks(symlinks)
 
+    installed_packages = []
     for install_data, success in zip(symlinks_to_install, install_results):
-        result = 'OK' if success else 'ERROR: Cannot create symlink, try to run as administrator.'
+        if success:
+            installed_packages.append(install_data['name'])
+            result = 'OK'
+        else:
+            result = 'ERROR: Cannot create symlink, try to run as administrator.'
         results.append((install_data['name'], result))
 
     if not all(install_results):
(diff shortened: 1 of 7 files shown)


gonzalocasas:
@romanarust I made the method public, because we'll be using it from `compas_fab` already now, so if it's public it would be clearer if someone at some point wants to remove/refactor it, that there might be other code using it.

beverlylytle (approved):
The Extension Points documentation additions don't link anywhere, but that can be taken later.  Code looks good, but I admit there's been no functional testing.

src/compas_rhino/install.py
```diff
@@ -106,11 +115,40 @@ def install(version=None, packages=None):
...
+                print('   {} {}: {}'.format(package.ljust(20), status, message))
+            except ValueError:
+                post_execution_errors.append(ValueError('Step successful, but result is wrongly formatted: {}'.format(str(result))))
+
+    if post_execution_errors:
+        print()
+        print('One or more errors occurred:')
+        print()
+        for error in post_execution_errors:
+            print('   - {}'.format(repr(error)))
```
beverlylytle, line 149: should an error here result in an `exit_code = -1`?

src/compas_rhino/install.py
```diff
@@ -106,11 +115,40 @@ def install(version=None, packages=None):
...
+    post_execution_errors = []
+    for result in steps_generator:
+        if isinstance(result, Exception):
+            post_execution_errors.append(result)
+            continue
+
+        for item in result:
+            try:
+                package, message, success = item
+                status = 'OK' if success else 'ERROR'
```
beverlylytle, line 142: should an error here result in an `exit_code = -1`?

gonzalocasas:
just as fyi, this is how the thing runs with a simple (fake) implementation of the pluggable:

image

src/compas_rhino/install.py
```diff
@@ -106,11 +115,40 @@ def install(version=None, packages=None):
...
+                print('   {} {}: {}'.format(package.ljust(20), status, message))
+            except ValueError:
+                post_execution_errors.append(ValueError('Step successful, but result is wrongly formatted: {}'.format(str(result))))
```
beverlylytle: and here, should `result` be `item`?

src/compas_rhino/install.py
gonzalocasas: you're right, fixed at https://github.com/compas-dev/compas/pull/770/commits/7b6ef0dba37bc879b4c26c0781dd2e6e8cfee326

src/compas_rhino/install.py
gonzalocasas: ups! https://github.com/compas-dev/compas/pull/770/commits/6440b4d499e3b9e09552f09165a80090cfa40e8b

src/compas_rhino/install.py
beverlylytle: I think it's a little more complicated.  Shouldn't all the successes be collected into some list `successes` and then instead of returning True, `all(successes)` is returned?

src/compas_rhino/install.py
```diff
@@ -106,11 +115,45 @@ def install(version=None, packages=None):
...
+                post_execution_errors.append(ValueError('Step successful, but result is wrongly formatted: {}'.format(str(item))))
+
+    if post_execution_errors:
+        print()
+        print('One or more errors occurred:')
+        print()
+        for error in post_execution_errors:
+            print('   - {}'.format(repr(error)))
+
+        return False
```
beverlylytle: This is confusing to me.  If there are `post_execution_errors`, then it's been printed that the step was successful, but now we are getting an -1 exit code.  So is it a success or an error?

gonzalocasas: How about now? I fixed the issue you found [here](https://github.com/compas-dev/compas/pull/770#discussion_r575369400) and hopefully made it a bit clearer what the return value of this method is supposed to be: https://github.com/compas-dev/compas/pull/770/commits/ae01ee4c4c6ae4fd106ead408391a734abbe432f

beverlylytle: I guess it's that I don't like the message of the `ValueError`. If the result isn't formatted correctly how can it be assume that the step was successful?

gonzalocasas: ahh... mmm... fair point. how about: `ValueError('Step ran without errors but result is wrongly formatted: {}'.format(str(item))`?

beverlylytle: that is a less confusing message for me.

gonzalocasas: done! https://github.com/compas-dev/compas/pull/770/commits/4836330d3c59e811e31607c18dcdf92b23e319c7

PR #774, opened 2021-02-18
Polyline functions

mattiskoh:

### Checklist

- [ x] I ran lint on my computer and there are no errors (i.e. `invoke lint`).
- [ x] I added new functions/classes and made them available on a second-level import, e.g. `compas.datastructures.Mesh`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 
 ### Added
-* Added `Polyline.split_at_corners` and `.Polyline.tangent_at_point_on_polyline`.
+* Added `divide_polyline`, `divide_polyline_by_length`, `Polyline.split_at_corners` and `.Polyline.tangent_at_point_on_polyline`.
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 
 ### Added
-* Added `divide_polyline`, `divide_polyline_by_length`, `Polyline.split_at_corners` and `.Polyline.tangent_at_point_on_polyline`.
+* Added `divide_polyline`, `divide_polyline_by_length`, `Polyline.split_at_corners` and `Polyline.tangent_at_point_on_polyline`.
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 
 ### Added
+* Added `Polyline.split_at_corners` and `.Polyline.tangent_at_point_on_polyline`.
 
 ### Changed
 
--- a/src/compas/geometry/primitives/polyline.py
+++ b/src/compas/geometry/primitives/polyline.py
@@ -269,7 +269,7 @@ class Polyline(Primitive):
 
         Returns
         -------
-        list of polylines
+        list of :class:`compas.geometry.Polyline`
 
         """
 
--- a/src/compas/geometry/primitives/polyline.py
+++ b/src/compas/geometry/primitives/polyline.py
@@ -295,7 +295,7 @@ class Polyline(Primitive):
             if id1 < id2:
                 split_polylines.append(Polyline(points[id1:id2+1]))
             else:
-                looped_pts = [points[i] for i in range(id1, len(points))] + points[0:id2+1]
+                looped_pts = [points[i] for i in range(id1, len(points))] + points[1:id2+1]
                 split_polylines.append(Polyline(looped_pts))
 
         return split_polylines
--- a/src/compas/geometry/primitives/polyline.py
+++ b/src/compas/geometry/primitives/polyline.py
@@ -298,9 +298,10 @@ class Polyline(Primitive):
             else:
                 looped_pts = [points[i] for i in range(id1, len(points))] + points[1:id2+1]
                 split_polylines.append(Polyline(looped_pts))
-if self.is_closed() and not corner_ids:
-    return [Polyline(self.points)]
-    
+
+        if self.is_closed() and not corner_ids:
+            return [Polyline(self.points)]
+
         return split_polylines
 
     def tangent_at_point_on_polyline(self, point):
@@ -332,33 +333,7 @@ if self.is_closed() and not corner_ids:
         """
         segment_length = self.length/num_segments
 
-        num_pts = int(round(self.length/segment_length))
-        total_length = [0, 0]
-        if hasattr(self, 'points'):
-            division_pts = [self.points[0]]
-            new_polyline = self
-        else:
-            division_pts = [self.start]
-            new_polyline = Polyline([self.start, self.end])
-
-        for i in range(num_pts):
-            for i_ln, line in enumerate(new_polyline.lines):
-                total_length.append(total_length[-1] + line.length)
-                if total_length[-1] > segment_length:
-                    amp = (segment_length - total_length[-2]) / line.length
-                    new_pt = line.start + line.vector.scaled(amp)
-                    division_pts.append(new_pt)
-                    total_length = [0, 0]
-                    remaining_pts = new_polyline.points[i_ln+2:]
-                    new_polyline = Polyline([new_pt, line.end] + remaining_pts)
-                    break
-            if len(division_pts) == num_pts+1:
-                break
-
-        if len(division_pts) != num_pts+1:
-            division_pts.append(new_polyline.points[-1])
-
-        return division_pts
+        return self.divide_polyline_by_length(segment_length)
 
     def divide_polyline_by_length(self, length):
         """Splits a polyline in segments of a given length
@@ -374,7 +349,12 @@ if self.is_closed() and not corner_ids:
         """
         num_pts = int(round(self.length/length))
         total_length = [0, 0]
-        division_pts = [self.points[0]]
+        if hasattr(self, 'points'):
+            division_pts = [self.points[0]]
+            new_polyline = self
+        else:
+            division_pts = [self.start]
+            new_polyline = Polyline([self.start, self.end])
         new_polyline = self
         for i in range(num_pts):
             for i_ln, line in enumerate(new_polyline.lines):
--- a/src/compas/geometry/primitives/polyline.py
+++ b/src/compas/geometry/primitives/polyline.py
@@ -319,6 +319,7 @@ class Polyline(Primitive):
 # Main
 # ==============================================================================
 
+
 if __name__ == '__main__':
 
     import doctest
--- a/src/compas/geometry/primitives/polyline.py
+++ b/src/compas/geometry/primitives/polyline.py
@@ -305,7 +305,7 @@ class Polyline(Primitive):
 
         Parameters:
         -----------
-        point: :class:`compas.geometry.Primitives.Point`
+        point: :class:`compas.geometry.Point`
 
         Returns
         -------
--- a/src/compas/geometry/primitives/polyline.py
+++ b/src/compas/geometry/primitives/polyline.py
@@ -7,6 +7,7 @@ from compas.geometry import transform_points
 from compas.geometry.primitives import Primitive
 from compas.geometry.primitives import Point
 from compas.geometry.primitives import Line
+from compas.geometry.predicates import is_point_on_line
 
 from compas.utilities import pairwise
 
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 
 ### Added
+* Added `divide_polyline`, `divide_polyline_by_length`, `Polyline.split_at_corners` and `.Polyline.tangent_at_point_on_polyline`.
```
```suggestion
* Added `divide_polyline`, `divide_polyline_by_length`, `Polyline.split_at_corners` and `Polyline.tangent_at_point_on_polyline`.
```

src/compas/geometry/primitives/polyline.py
```diff
@@ -258,11 +259,134 @@ def transform(self, T):
             self.points[index].y = point[1]
             self.points[index].z = point[2]
 
+    def split_at_corners(self, angle_threshold):
+        """Splits a polyline at corners larger than the given angle_threshold
 
+        Parameters:
+        -----------
+        angle_threshold: radians
```
beverlylytle:
```suggestion
        angle_threshold : float
            In radians.
```

src/compas/geometry/primitives/polyline.py
```diff
@@ -258,11 +259,134 @@ def transform(self, T):
...
+
+        Returns
+        -------
+        list of polylines
```
beverlylytle:
```suggestion
        list of :class:`compas.geometry.Polyline`
```

src/compas/geometry/primitives/polyline.py
```diff
@@ -258,11 +259,134 @@ def transform(self, T):
...
+                split_polylines.append(Polyline(looped_pts))
+
+        return split_polylines
+
+    def tangent_at_point_on_polyline(self, point):
+        """Calculates the tangent vector of a point on a polyline
+
+        Parameters:
+        -----------
+        point: :class:`compas.geometry.Primitives.Point`
```
beverlylytle:
```suggestion
        point: :class:`compas.geometry.Point`
```

src/compas/geometry/primitives/polyline.py
```diff
@@ -258,11 +259,134 @@ def transform(self, T):
...
+
+        Returns
+        -------
+        Vector
```
beverlylytle:
```suggestion
        :class:`compas.geometry.Vector`
```

src/compas/geometry/primitives/polyline.py
```diff
@@ -258,11 +259,134 @@ def transform(self, T):
...
+        """Divide a polyline in equal segments
+
+        Parameters:
+        -----------
+        num_segments: int
+
+        Returns
+        -------
+        list
+            list of points :class:`compas.geometry.Primitives.Point`
```
beverlylytle:
```suggestion
            list of :class:`compas.geometry.Point`
```

src/compas/geometry/primitives/polyline.py
```diff
@@ -258,11 +259,134 @@ def transform(self, T):
...
+        for line in self.lines:
+            if is_point_on_line(point, line):
+                return line.direction
+
+    def divide_polyline(self, num_segments):
+        """Divide a polyline in equal segments
+
+        Parameters:
+        -----------
+        num_segments: int
```
beverlylytle:
```suggestion
        num_segments : int
```

src/compas/geometry/primitives/polyline.py
```diff
@@ -258,11 +259,134 @@ def transform(self, T):
...
+            division_pts.append(new_polyline.points[-1])
+
+        return division_pts
+
+    def divide_polyline_by_length(self, length):
+        """Splits a polyline in segments of a given length
+
+        Parameters:
+        -----------
+        length: float
```
beverlylytle:
```suggestion
        length : float
```

src/compas/geometry/primitives/polyline.py
```diff
@@ -258,11 +259,134 @@ def transform(self, T):
...
+
+        Returns
+        -------
+        list
+            list of points :class:`compas.geometry.Primitives.Point`
```
beverlylytle:
```suggestion
            list of :class:`compas.geometry.Point`
```

src/compas/geometry/primitives/polyline.py
```diff
@@ -258,11 +259,134 @@ def transform(self, T):
...
+        else:
+            corner_ids = [0] + corner_ids + [len(points)]
+
+        for id1, id2 in pairwise(corner_ids):
+            if id1 < id2:
+                split_polylines.append(Polyline(points[id1:id2+1]))
+            else:
+                looped_pts = [points[i] for i in range(id1, len(points))] + points[1:id2+1]
+                split_polylines.append(Polyline(looped_pts))
+
```
beverlylytle, line 301: If this is invoked on a closed Polyline that approximates, say, a circle with very many line segments and a sufficiently large `angle_threshold`, then this would return an empty list, and I think it makes sense to return a list containing only a copy of the original Polyline.
```suggestion
if self.is_closed() and not corner_ids:
    return [Polyline(self.points)]
    
```

src/compas/geometry/primitives/polyline.py
```diff
@@ -258,11 +259,134 @@ def transform(self, T):
...
+        if self.is_closed() and len(corner_ids) > 0:
+            if corner_ids[-1] != len(points):
+                corner_ids = [corner_ids[-1]] + corner_ids
+        else:
+            corner_ids = [0] + corner_ids + [len(points)]
+
+        for id1, id2 in pairwise(corner_ids):
+            if id1 < id2:
+                split_polylines.append(Polyline(points[id1:id2+1]))
```
beverlylytle, line 297: Do you really want that the new `Polyline`s use the same `Point`s, or should the points be copied?

src/compas/geometry/primitives/polyline.py
```diff
@@ -258,11 +259,134 @@ def transform(self, T):
...
+                    remaining_pts = new_polyline.points[i_ln+2:]
+                    new_polyline = Polyline([new_pt, line.end] + remaining_pts)
+                    break
+            if len(division_pts) == num_pts+1:
+                break
+
+        if len(division_pts) != num_pts+1:
+            division_pts.append(new_polyline.points[-1])
+
+        return division_pts
```
beverlylytle: This code is just the same as below, right?  so to avoid duplication, just call the function below.
```suggestion
        return self.divide_polyline_by_length(segement_length)
```

src/compas/geometry/primitives/polyline.py
```diff
@@ -258,11 +259,134 @@ def transform(self, T):
...
+        -----------
+        point: :class:`compas.geometry.Primitives.Point`
+
+        Returns
+        -------
+        Vector
+        """
+        for line in self.lines:
+            if is_point_on_line(point, line):
+                return line.direction
```
beverlylytle, line 320:
```suggestion
                return line.direction
            raise Exception('{} not found!'.format(point))
```

src/compas/geometry/primitives/polyline.py
```diff
@@ -258,11 +259,134 @@ def transform(self, T):
...
+        Parameters:
+        -----------
+        length: float
+
+        Returns
+        -------
+        list
+            list of points :class:`compas.geometry.Primitives.Point`
+        """
+        num_pts = int(round(self.length/length))
```
beverlylytle: This line I find a little confusing.  First, I think `round` returns an `int` when the precision is not specified.  Secondly, `round` behaves scientifically rather than computer-sciencely in that `round(4.5) == 4` while `round(5.5) == 6`.  So in `divide_polyline` I don't think you always get exactly the number of segments that you specify, but maybe that's ok.  Would `int(self.length / length)` make sense here (which always rounds down)?

beverlylytle (changes requested):
Looking pretty good.  I know it looks like a lot of comments, but mostly they are trivial things to do with docstring style. 

I think it would be a really good thing to add some tests for these new functions.  In particular, testing around some edge cases. For example `divide_polyline` on polylines with only two points or with many, on closed or not closed polylines, with few or many `num_segments`.  And similarly for `divide_polyline_by_length` with a length that divides (or not) the polyline's length, or exceeds it. Etc.

src/compas/geometry/primitives/polyline.py
mattiskoh: I have no strong opinion about this. I definitely understand the reason to copy these points to avoid any mishaps later on in the function, but as it seems to work I would just leave it as is. 

Performance-wise I thought this would be the most efficient ( but I'm definitely no expert 😂 )

If you feel that it would improve the code by copying feel free to add.

src/compas/geometry/primitives/polyline.py
mattiskoh: Correct, I will add the edge case for the Polyline existing of only 2pts (a Line..) to `self.divide_polyline_by_length` and simplify `self.divide_polyline` as above.

src/compas/geometry/primitives/polyline.py
mattiskoh: Yes, you're right. I will add the change

src/compas/geometry/primitives/polyline.py
beverlylytle: I am ambivalent about this, weighing performance against hard to spot bugs. So, I'd say leave it as it is until someone complains ;)

src/compas/geometry/primitives/polyline.py
```diff
@@ -258,11 +259,123 @@ def transform(self, T):
...
+        point: :class:`compas.geometry.Point`
+
+        Returns
+        -------
+        :class:`compas.geometry.Vector`
+        """
+        for line in self.lines:
+            if is_point_on_line(point, line):
+                return line.direction
+            raise Exception('{} not found!'.format(point))
```
beverlylytle: I think this needs to have one less level indentation.  Sorry!
```suggestion
        raise Exception('{} not found!'.format(point))
```

tests/compas/geometry/test_polyline.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+@pytest.mark.parametrize('coords,expected', [
+    ([[0.0, 0.0, 0.0], [100.0, 0.0, 0.0]], [[0.0, 0.0, 0.0], [80.0, 0.0, 0.0], [100.0, 0.0, 0.0]]),
+])
+def test_polyline_divide_length_strict2(coords, expected):
+    assert expected == Polyline(coords).divide_polyline_by_length(80, False)
+
+
+@pytest.mark.parametrize('coords, input, expected', [
```
beverlylytle:
```suggestion
@pytest.mark.parametrize('coords,input,expected', [
```

tests/compas/geometry/test_polyline.py
```diff
@@ -0,0 +1,54 @@
...
+    ([[0.0, 0.0, 0.0], [1.0, 0.0, 0.0], [1.0, 1.0, 0.0], [0.0, 1.0, 0.0], [0.0, 0.0, 0.0]], math.pi/2, [Polyline([[0.0, 0.0, 0.0], [1, 0.0, 0.0]]),
+                                                                                                        Polyline([[1, 0.0, 0.0], [1, 1, 0.0]]), Polyline([[1, 1, 0.0], [0.0, 1, 0.0]]), Polyline([[0.0, 1, 0.0], [0.0, 0.0, 0.0]])]),
+])
+def test_polyline_split_at_corners(coords, input, expected):
+    assert expected == Polyline(coords).split_at_corners(input)
+
+
+@pytest.mark.parametrize('coords, input, expected', [
```
beverlylytle:
```suggestion
@pytest.mark.parametrize('coords,input,expected', [
```

src/compas/geometry/primitives/polyline.py
```diff
@@ -258,11 +259,131 @@ def transform(self, T):
...
+        return self.divide_polyline_by_length(segment_length, False)
+
+    def divide_polyline_by_length(self, length, strict=True):
+        """Splits a polyline in segments of a given length
+
+        Parameters:
+        -----------
+        length: float
+
+        strit: boolean
```
beverlylytle:
```suggestion
        strict : bool
```

src/compas/geometry/primitives/polyline.py
```diff
@@ -258,11 +259,131 @@ def transform(self, T):
...
+            If set to False, the remainder segment will be added even if it is smaller than the desired length
```
beverlylytle:
```suggestion
            If set to ``False``, the remainder segment will be added even if it is smaller than the desired length
```

beverlylytle (approved):
Lookin good!  let's merge this!

PR #778, opened 2021-03-01
Update `Box` transform

romanarust:
Small fix on box transform to apply `Scale` to its width, height, length.

### Checklist


beverlylytle:
I would have expected that `Box.transform` behaves this way already, that is, that the dimensions of the box change according to the scaling of the transformation.  Indeed, `Sphere.transform` both moves the center point and scales the radius.  However, `Cylinder.transform` does not scale the height or the radius of the object.  `Capsule.transform` is confusing, in that it transforms the defining line, but not the radius.  All I mean to say is, I think this is a good change for `Box`, but for the other shapes and primitives, the behavior of `transform` is not consistent, and this maybe should be discussed.

tomvanmele:
just to add on to this, what happens if the transformation contains a shear and/or projection component?

romanarust:
This problem applies both to `Shape` and `Primitive`. 
If a `Box` is projected or sheared, is it then still a `Box`? If a `Circle` is scaled non-uniformly, is it then still a `Circle`?
We could ignore such "distortions" and simply raise a warning...

beverlylytle:
In the case of `Sphere`, the maximum of the scalars is applied to the radius...

tomvanmele:
> We could ignore such "distortions" and simply raise a warning...

so we allow the user to do what they want but point out in the docs that distortions might occur?
i would be in favour of that because the alternative is to do a lot of expensive checks in the background for every transformation...

tomvanmele:
what is the plan with this?

gonzalocasas:
Let's merge this as is.

PR #781, opened 2021-03-08
add __ne__ to primitives and transformations

beverlylytle:
Apparently python2 doesn't infer that `__ne__` is `not __eq__` resulting in things like
```

print(Point(1,1,1) == Point(1,1,1))
print(Point(1,1,1) != Point(1,1,1))
```
printing `True` twice.  

I discovered this while investigating why ironpython tests were failing for #774.   

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 * Fixed bug where mimic joints were considered configurable.
+* Fixed bug where `!=` gave incorrect results in Rhino for compas objects.
 
 ### Removed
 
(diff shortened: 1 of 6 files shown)


tomvanmele:
as discussed in one of the last dev meetings, it is not clear what the side effects are of pushing this all the way onto the base class, because, for example, it prevents COMPAS objects from being used as keys in dictionaries, which is a useful feature to create runtime maps.

perhaps we can for now limit this to primitives and shapes, where the meaning of this comparison is also more clear?

also, since we are talking specifically about `__ne__` i must have missed the part where we added `__eq__` for which the same argumets are true...

beverlylytle:
> perhaps we can for now limit this to primitives and shapes, where the meaning of this comparison is also more clear?
> 

sure, i'll move it.

beverlylytle:
Since shapes don't have `__eq__` implemented, I left them out.

beverlylytle:
Since `Transformation` has `__eq__` implemented, I added it there too.

tomvanmele (approved):

PR #782, opened 2021-03-08
stop deploy PRs and only keep latest patch

Licini:
Due to the limit 1GB limit of gh-pages, updating the docs workflow to:
- docs for PRs are still built for testing but no longer pushed to gh-pages.
- Only latest patch of each minor versions will kept.


.github/workflows/docs.yml
```diff
@@ -40,19 +44,23 @@ jobs:
...
                   FOLDER_NAME=latest
+                  echo "::set-output name=commit_type::main"
               fi;
           fi;
           echo "Docs will be deployed to https://compas.dev/compas/$FOLDER_NAME"
           mkdir -p deploy/$BRANCH_OR_TAG && mv -T dist/docs deploy/$FOLDER_NAME/
 
       - name: Deploy docs
-        if: success() && env.GH_PAT != ''
+        if: success() && env.GH_PAT != '' && steps.docs.outputs.commit_type != 'pull'
```
Licini, line 63: pull request will be skipped deploy here

.github/workflows/docs.yml
```diff
@@ -63,27 +71,36 @@ jobs:
...
+          echo current patch: $CURRENT_PATCH current minor: $CURRENT_MINOR
           for folder in $(ls -rd */ | tr -d '/')
           do
-              if [[ $folder =~ ^[0-9]{1,2}\.[0-9]{1,2}\.[0-9]{1,2}$|^latest$ ]]
-              then
-                  echo ${folder}  >> doc_versions.txt
+              if [[ $folder =~ ^[0-9]{1,2}\.[0-9]{1,2}\.[0-9]{1,2}$ ]]; then
+                  PATCH=$folder
+                  MINOR=${folder%.*}
+                  if [[ $PATCH != $CURRENT_PATCH && $MINOR = $CURRENT_MINOR ]]; then
```
Licini, line 95: The older patches of same minor version will be deleted here

PR #784, opened 2021-03-12
add __eq__ for circle and line

### LINKED ISSUE #776: Circle and Line do not implement builtin equivalence
The primitives should have equivalence with their builtins counterparts (eg `Point([0, 0, 1]) == [0, 0, 1]`) but `Circle` and `Line` are missing the implementation (`Line`  raises a `NotImplemented`, and `Circle` does not override `__eq__` so it is probably using identity equality).

**Expected behavior**
```
from compas.geometry import Circle

p1 = [0, 0, 0]
p2 = [1, 0, 0]
assert (p1, p2) == Line(p1, p2)

point = [0, 0, 0]
vector = [1, 0, 0]
plane = (point, vector)
assert (plane, 1.0) == Circle(plane, 1.0)
```

**Additional context**
Documented primitive equivalences: https://compas.dev/compas/latest/tutorial/geometry.html#other-primitives

beverlylytle:
Closes #776

### Checklist

### DIFF
--- a/src/compas/geometry/primitives/circle.py
+++ b/src/compas/geometry/primitives/circle.py
@@ -141,6 +141,14 @@ class Circle(Primitive):
     def __iter__(self):
         return iter([self.plane, self.radius])
 
+    def __eq__(self, other):
+        try:
+            other_plane = other[0]
+            other_radius = other[1]
+        except Exception:
+            return False
+        return self.plane == other_plane and self.radius == other_radius
+
     # ==========================================================================
     # constructors
     # ==========================================================================
--- a/src/compas/geometry/primitives/circle.py
+++ b/src/compas/geometry/primitives/circle.py
@@ -145,7 +145,7 @@ class Circle(Primitive):
         try:
             other_plane = other[0]
             other_radius = other[1]
-        except Exception:
+        except:  # noqa: E722
             return False
         return self.plane == other_plane and self.radius == other_radius
 
--- a/src/compas/geometry/primitives/line.py
+++ b/src/compas/geometry/primitives/line.py
@@ -147,7 +147,7 @@ class Line(Primitive):
         try:
             other_start = other[0]
             other_end = other[1]
-        except Exception:
+        except:  # noqa: E722
             return False
         return self.start == other_start and self.end == other_end
 
--- a/src/compas/geometry/primitives/line.py
+++ b/src/compas/geometry/primitives/line.py
@@ -144,7 +144,12 @@ class Line(Primitive):
         return iter([self.start, self.end])
 
     def __eq__(self, other):
-        raise NotImplementedError
+        try:
+            other_start = other[0]
+            other_end = other[1]
+        except Exception:
+            return False
+        return self.start == other_start and self.end == other_end
 
     # ==========================================================================
     # constructors
--- a/tests/compas/geometry/test_circle.py
+++ b/tests/compas/geometry/test_circle.py
@@ -1,0 +1,18 @@
+from compas.geometry import Circle
+from compas.geometry import Plane  # noqa: F401
+from compas.geometry import Point  # noqa: F401
+from compas.geometry import Vector  # noqa: F401
+
+
+def test_circle():
+    point = [0, 0, 0]
+    vector = [1, 0, 0]
+    plane = (point, vector)
+    c = Circle(plane, 1.0)
+    assert c.radius == 1.0
+    assert c.plane == plane
+    assert c == (plane, 1.0)
+    assert c == Circle(plane, 1.0)
+    assert c != 1.0
+    assert c != (plane, 2.0)
+    assert c == eval(repr(c))
--- a/tests/compas/geometry/test_circle.py
+++ b/tests/compas/geometry/test_circle.py
@@ -1,4 +1,5 @@
 from compas.geometry import Circle
+# These imports are used to check __repr__.
 from compas.geometry import Plane  # noqa: F401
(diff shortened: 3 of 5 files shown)


src/compas/geometry/primitives/circle.py
```diff
@@ -141,6 +141,14 @@ def __setitem__(self, key, value):
     def __iter__(self):
         return iter([self.plane, self.radius])
 
+    def __eq__(self, other):
+        try:
```
tomvanmele, line 145: why the try-except?
would just let this fail if the circle is not "properly" defined, no?

src/compas/geometry/primitives/line.py
```diff
@@ -144,7 +144,12 @@ def __iter__(self):
         return iter([self.start, self.end])
 
     def __eq__(self, other):
-        raise NotImplementedError
+        try:
```
tomvanmele, line 147: same comment

src/compas/geometry/primitives/circle.py
```diff
@@ -141,6 +141,14 @@ def __setitem__(self, key, value):
...
+            other_plane = other[0]
```
gonzalocasas, line 146: Could it be that other is an `iterable` and index accessors cannot be used?

The case that comes to mind is that of built-in equivalences of primitives; in principle, an iterable could also be a circle, but I am not sure if we care for iters when it comes to these things. If not, then ignore this point.

tests/compas/geometry/test_circle.py
```diff
@@ -0,0 +1,18 @@
...
+    vector = [1, 0, 0]
+    plane = (point, vector)
+    c = Circle(plane, 1.0)
+    assert c.radius == 1.0
+    assert c.plane == plane
+    assert c == (plane, 1.0)
+    assert c == Circle(plane, 1.0)
+    assert c != 1.0
+    assert c != (plane, 2.0)
+    assert c == eval(repr(c))
```
gonzalocasas, line 33: I would prefer more verbosity in the tests, and split them into narrower ones with meaningful names (e.g. `test_circle_equality`, `test_circle_repr`.

tests/compas/geometry/test_circle.py
```diff
@@ -0,0 +1,18 @@
+from compas.geometry import Circle
+from compas.geometry import Plane  # noqa: F401
+from compas.geometry import Point  # noqa: F401
+from compas.geometry import Vector  # noqa: F401
```
gonzalocasas, line 5: Why are these imports here if they are not used?

gonzalocasas (approved):
Looks good to me, however, I have a few of points that need clarification, but mostly minor.

src/compas/geometry/primitives/circle.py
beverlylytle: Because something like Circle(plane, radius) == 1 should return false and not raise an exception.

tests/compas/geometry/test_circle.py
beverlylytle: They are used in the `eval(repr())` construct.   I put the comments in to make my IDE not complain, but it may be that they aren't necessary for pylint... I didn't check.

tests/compas/geometry/test_circle.py
beverlylytle: Can do.

src/compas/geometry/primitives/circle.py
beverlylytle: I lean toward the side of "if `other` doesn't allow index accessing, then it doesn't quack sufficiently  like a `Circle` or a list". 🦆

gonzalocasas: fair enough! I just double checked the docs, and I believe there's nothing hinting at the possibility of using iterables, so I guess it's fine.

tests/compas/geometry/test_circle.py
gonzalocasas: aha! good! then I guess pylint also needs the `noqa`. I would add a comment about why these are needed here then

src/compas/geometry/primitives/circle.py
beverlylytle: ```Frame.worldXY() == Polyline([[0,0,0], [1,0,0], [0,1,0]])``` returns `True`. so that's kind of funny.

gonzalocasas: 🤔 maybe we need to be a bit more strict in that checking, if `other` is of a non-builtin type, stricter type checking should be in place, at least, it should be same class of sub-class of it

beverlylytle: ha, actually, it's a little weirder...
```Frame.worldXY() == Polyline([[0,0,0], [1,0,0], [0,1,0], [2,3,4])``` is true, and i was expecting that ```Polyline([[0,0,0], [1,0,0], [0,1,0], [2,3,4]) == Frame.worldXY()``` would return `False`, however there's a bug in  `Polyline.__eq__`.  so the following are both true..
```
Polyline([[0,0,0], [1,0,0], [0,1,0], [2,3,4]]) == Polyline([[0,0,0], [1,0,0], [0,1,0]])
Polyline([[0,0,0], [1,0,0], [0,1,0]]) == Polyline([[0,0,0], [1,0,0], [0,1,0], [2,3,4]])

beverlylytle: and i guess the same bug exists for polygon

PR #785, opened 2021-03-15
fix custom magic methods and add tests

beverlylytle:
I noticed some small bugs in the custom magic methods of a few primitives.  In particular
`Polyline([[0, 0, 0], [0, 0, 1]]) == Polyline([[0, 0, 0], [0, 0, 1], [0, 0, 2])` would return `True` , and a similar bug existed in a few other classes.  

### Checklist

### DIFF
--- a/src/compas/geometry/primitives/frame.py
+++ b/src/compas/geometry/primitives/frame.py
@@ -180,6 +180,8 @@ class Frame(Primitive):
         return iter([self.point, self.xaxis, self.yaxis])
 
     def __eq__(self, other, tol=1e-05):
+        if not hasattr(other, '__iter__') or not hasattr(other, '__len__') or len(self) != len(other):
+            return False
         for v1, v2 in zip(self, other):
             for a, b in zip(v1, v2):
                 if math.fabs(a - b) > tol:
--- a/src/compas/geometry/primitives/frame.py
+++ b/src/compas/geometry/primitives/frame.py
@@ -1,25 +1,24 @@
 from __future__ import print_function
 
-import math
-
-from compas.geometry import cross_vectors
-from compas.geometry import subtract_vectors
-from compas.geometry import matrix_from_basis_vectors
-from compas.geometry import basis_vectors_from_matrix
-from compas.geometry import quaternion_from_matrix
-from compas.geometry import matrix_from_quaternion
-from compas.geometry import axis_angle_vector_from_matrix
-from compas.geometry import matrix_from_axis_angle_vector
-from compas.geometry import euler_angles_from_matrix
-from compas.geometry import matrix_from_euler_angles
-from compas.geometry import decompose_matrix
-from compas.geometry import Transformation
+from compas.geometry import allclose
 from compas.geometry import argmax
+from compas.geometry import axis_angle_vector_from_matrix
+from compas.geometry import basis_vectors_from_matrix
+from compas.geometry import cross_vectors
+from compas.geometry import decompose_matrix
+from compas.geometry import euler_angles_from_matrix
+from compas.geometry import matrix_from_axis_angle_vector
+from compas.geometry import matrix_from_basis_vectors
+from compas.geometry import matrix_from_euler_angles
+from compas.geometry import matrix_from_quaternion
+from compas.geometry import quaternion_from_matrix
+from compas.geometry import subtract_vectors
+from compas.geometry import Transformation
 
-from compas.geometry.primitives import Primitive
 from compas.geometry.primitives import Point
-from compas.geometry.primitives import Vector
+from compas.geometry.primitives import Primitive
 from compas.geometry.primitives import Quaternion
+from compas.geometry.primitives import Vector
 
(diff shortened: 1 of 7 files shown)


src/compas/geometry/primitives/frame.py
```diff
@@ -180,6 +180,8 @@ def __iter__(self):
         return iter([self.point, self.xaxis, self.yaxis])
 
     def __eq__(self, other, tol=1e-05):
+        if not hasattr(other, '__iter__') or not hasattr(other, '__len__') or len(self) != len(other):
+            return False
```
tomvanmele, line 183: aren't `__iter__` and `__len__` automatically covered by `zip`?
also, you could use `zip_longest` instead of `zip` which will produce `None` for one of the iterables if the length is not the same...

src/compas/geometry/primitives/frame.py
```diff
@@ -180,6 +180,8 @@ def __iter__(self):
...
         for v1, v2 in zip(self, other):
             for a, b in zip(v1, v2):
                 if math.fabs(a - b) > tol:
```
tomvanmele: should we not replace this with the precision setting of `compas`?
noticed that we are not doing this for the other checks either...

src/compas/geometry/primitives/polygon.py
```diff
@@ -156,12 +156,15 @@ def __getitem__(self, key):
...
+        self.points[key] = Point(*value)
+        self._lines = None
 
     def __iter__(self):
         return iter(self.points)
 
     def __eq__(self, other):
+        if not hasattr(other, '__iter__') or not hasattr(other, '__len__') or len(self) != len(other):
+            return False
         return all(a == b for a, b in zip(self, other))
```
tomvanmele: we are not being consistent with the precision checks. perhaps we could use `allclose` everywhere?

tests/compas/geometry/test_polygon.py
```diff
@@ -0,0 +1,55 @@
...
+    polygon1 = Polygon(points1)
+    points2 = [[0, 0, x] for x in range(6)]
+    polygon2 = Polygon(points2)
+    points3 = [[0, 0, x] for x in range(5)] + [[0, 0, 0]]
+    polygon3 = Polygon(points3)
+    assert polygon1 == polygon1
+    assert polygon1 == points1
+    assert points1 == polygon1
+    assert polygon1 != polygon2
+    assert polygon2 != polygon1
```
tomvanmele, line 26: do the `__ne__` still have to be added for `ipy` as well?

src/compas/geometry/primitives/frame.py
beverlylytle: Yes and no.  If `other` doesn't have `__iter__`, `zip` will raise an error.  I feel that `__eq__` should never raise an error. For example `[0, 0, 0] == 0` would return `False` rather than raise an error, and if compas primitives are supposed to behave like python primitives, then we should mimic that. 

Using `zip_longest` would also throw an exception in the case you describe because you can't perform arithmetic on `None`.

src/compas/geometry/primitives/frame.py
beverlylytle: Yeah, that makes sense.

src/compas/geometry/primitives/polygon.py
beverlylytle: Can change this one, too.

tests/compas/geometry/test_polygon.py
beverlylytle: It was added in `Primitive`, so it is inherited. or do you mean something else?

src/compas/geometry/primitives/frame.py
tomvanmele: okay. fair enough...

tests/compas/geometry/test_polygon.py
tomvanmele: no. didn't think of that...

src/compas/geometry/primitives/frame.py
beverlylytle: actually, i'm a little confused by this comment.  isn't `compas.PRECISION` more about formatting the way objects are printed rather than prescribing tolerances on computations?

tomvanmele: yes and no. it is, for example, also used in the geometric maps to determine comparison precision...
but i agree that there is no "official policy" for this at the moment.

tomvanmele: but in the tests, when geometry comparisons are concerned, we tend to use (or should be using) `allclose` to avoid tests failing for lack of precision

beverlylytle: yeah, i've changed to using `allclose` now, but that also uses a hardcoded tolerance...

maybe once there is an "official policy" on this, such changes can be made more thoroughly.

tomvanmele (approved):

PR #786, opened 2021-03-15
Switch to static API

tomvanmele:


tomvanmele:
this replaces the `pylance` PR. it had gotten a bit messy and the original idea there was to switch to stubs, but this actually addresses a separate issue since switching to a static API already addresses the problem of Pylance not being able to find stuff...

tomvanmele:
@yck011522

tomvanmele:
@duchaoyu

src/compas/numerical/__init__.py
```diff
@@ -88,25 +88,103 @@
...
+        'stiffness_matrix',
+        'equilibrium_matrix',
+        'grad',
+        'div',
+        'curl',
+        'float_formatter',
+        'set_array_print_precision',
+        'unset_array_print_precision',
+    ]
+    __all__ += [
```
Licini: small thing... why is this not one list?

Licini (approved):
Wow that was a hell lot of refactoring.... Just curious did you do all of these manually or used some sort of tool??

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 
 ### Added
+
```
gonzalocasas, line 12: This one is big enough to deserve a line here, like `Make Pylance happy` ;)

src/compas/datastructures/mesh/planarisation_numpy.py
```diff
@@ -1,128 +0,0 @@
...
-from compas.utilities import window
-
-
-__all__ = [
-    'mesh_flatness',
-    'mesh_planarize_faces',
-]
-
-
-def mesh_flatness(mesh, maxdev=1.0):
```
gonzalocasas, line 20: I'm not sure here... these two functions in this file are deleted, but still in the imports, so, I guess they are somewhere else and this was duplicated?

src/compas/datastructures/network/planarity.py
```diff
@@ -27,18 +27,9 @@
     'network_is_planar',
     'network_is_planar_embedding',
     'network_embed_in_plane',
-
-    'network_embed_in_plane_proxy',
 ]
 
 
-def network_embed_in_plane_proxy(data, fixed=None, straightline=True):
```
gonzalocasas: Removing this function would make it a breaking change, unless this is duplicated and also provided by some other file? Otherwise I would suggest to add it back.

src/compas/datastructures/network/planarity_.py
```diff
@@ -1,326 +0,0 @@
-from __future__ import print_function
```
gonzalocasas, line 1: Same comment on this file, the file is deleted, but the functions are still listed on the API, so, are they coming from somewhere else?

src/compas/files/__init__.py
```diff
@@ -97,19 +97,26 @@
...
-from .off import *  # noqa: F401 F403
-from .ply import *  # noqa: F401 F403
-from .stl import *  # noqa: F401 F403
-from .urdf import *  # noqa: F401 F403
-from .xml import *  # noqa: F401 F403
-
-__all__ = [name for name in dir() if not name.startswith('_')]
+
+from .dxf import DXF, DXFReader, DXFParser
+from .gltf import GLTF, GLTFContent, GLTFExporter, GLTFMesh, GLTFParser, GLTFReader
```
gonzalocasas: don't we have `force_single_line` for imports? What I did for RRC was to do `isort: skip_file` for the `__init__` because I like the grid import style (which you used on `__init__` as well), but then everything else is single-line imports

src/compas/geometry/__init__.py
```diff
@@ -444,30 +444,666 @@
...
-from .pointclouds import *  # noqa: F401 F403
-from .quadmesh import *  # noqa: F401 F403
-from .triangulation import *  # noqa: F401 F403
-from .trimesh import *  # noqa: F401 F403
-
-
-__all__ = [name for name in dir() if not name.startswith('_')]
+import compas
+
+from ._core import (
```
gonzalocasas, line 449: style question: multiple grid imports from the same package for grouping, or a single grid import with interspersed comments?

src/compas/robots/__init__.py
```diff
@@ -113,12 +113,54 @@
...
+from .model import (
+    Geometry, Box, Cylinder, Sphere, Capsule, MeshDescriptor,
+    Color, Texture, Material, Origin
+)
+from .model import (
+    Joint, ParentLink, ChildLink, Calibration, Dynamics,
+    Limit, Axis, Mimic, SafetyController
+)
+from .model import Link, Inertial, Visual, Collision, Mass, Inertia
+from .model import RobotModel, ToolModel
```
gonzalocasas: here I'd go for a consistently single grid import from `.model`

src/compas/robots/__init__.py
```diff
@@ -113,12 +113,54 @@
...
+
+from .resources import AbstractMeshLoader, DefaultMeshLoader, LocalPackageMeshLoader
+from .resources import GithubPackageMeshLoader
```
gonzalocasas: and here as well...  in general, my vote goes for grid imports on every `__init__.py`

src/compas/topology/__init__.py
```diff
@@ -58,21 +58,48 @@
...
+    dijkstra_distances,
+    dijkstra_path
+)
+from .combinatorics import vertex_coloring, connected_components
+from .orientation import face_adjacency, unify_cycles
+from .connectivity import adjacency_from_edges
 
 if compas.IPY:
-    from .orientation_rhino import *  # noqa: F401 F403
+    from .orientation_rhino import face_adjacency_rhino, unify_cycles_rhino
```
gonzalocasas: not to change on this PR, but orientation_rhino should move to `compas_rhino` at some point

gonzalocasas (approved):
Other than one comment about a breaking change that I think we should revert, everything else looks good to me. Although, it's a very big change and very hard to be sure we're not breaking stuff.

My only other comment would be to settle on a defined style for the `__init__` long imports, for which I would suggest the grid import (basically what `isort` calls the `Pycharm` profile).

gonzalocasas:
@yck011522 if you work off of source code, would you have time to give this a try and see if nothing breaks on your end?

src/compas/datastructures/mesh/planarisation_numpy.py
tomvanmele: they are in `planarisation.py`. the numpy version was just a duplicate that was never updated to be based on numpy... so this doesn't break anything afaik...

src/compas/datastructures/network/planarity_.py
tomvanmele: yes they are in `planarity.py`. the file was just still on my system and therefore in the PR...

src/compas/datastructures/network/planarity.py
tomvanmele: sure i can put it back, but this `proxy` version of functions is just a left over from old code...

src/compas/robots/__init__.py
tomvanmele: by grid imports you mean in brackets and one per line?

src/compas/topology/__init__.py
tomvanmele: not sure if i agree. in my opinion it is like a `numpy` backend alternative but just with `rhino`...

src/compas/robots/__init__.py
gonzalocasas: Yep, exactly

src/compas/topology/__init__.py
gonzalocasas: But we should handle it with plugins, same as we do for booleans in which the operations are provided by compas_rhino when inside of it, otherwise they can be numpy-based

tomvanmele:
> Wow that was a hell lot of refactoring.... Just curious did you do all of these manually or used some sort of tool??

manually :)

src/compas/numerical/__init__.py
tomvanmele: no reason. was easier to keep things organised while moving stuff around. will combine...

src/compas/geometry/__init__.py
tomvanmele: maybe the latter

src/compas/topology/__init__.py
tomvanmele: okay so the pluggable is in the core lib in `topology` but the rhino plugin is in `compas_rhino`?

gonzalocasas: yep!

yck011522:
> @yck011522 if you work off of source code, would you have time to give this a try and see if nothing breaks on your end?

I would love to, I just couldn't do it this week. Static API sounds amazing.

PR #790, opened 2021-03-18
Stable API fixes

gonzalocasas:
Adding a unit tests that verifies if nothing has been removed from the public API.
Eventually more tests could be added to make sure that things have not changed signature, but for now, this should do.

The unit test fails like this if there's a removal compared the the recorded API:
image

### Checklist


tomvanmele (approved):

tomvanmele:
btw, i would argue that the import of functions that are already bound to the mesh class should be deprecated as well. and the ones that are not yet attached to the mesh should be added :)

gonzalocasas:
> btw, i would argue that the import of functions that are already bound to the mesh class should be deprecated as well. and the ones that are not yet attached to the mesh should be added :)

yes, absolutely agree. for now, i'd suggest we merge this (once all tests are green) and we do all deprecations in a separate one.

PR #796, opened 2021-03-23
fixes #795 Unetary support for vectors

### LINKED ISSUE #795: Unetary support for vectors
there seems no method of reversing / inverting a `compas.geometry.Vector` 
its pretty readible to write `Frame(pnt, -vX, -vY)`
this is mostly syntactically I suppose

```python
def invert(): # Transformation has this method
   self.scale(-1.)

def __neg__(self):
   return self.scaled(-1.)
```

jf---:
see issue #795

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+* Added a `invert` and `inverted` method `compas.geometry.Vector`.
+* Added unetary `__neg__` operator for `compas.geometry.Vector`
+
 ### Changed
 
 ### Removed
--- a/src/compas/geometry/primitives/vector.py
+++ b/src/compas/geometry/primitives/vector.py
@@ -255,6 +255,9 @@ class Vector(Primitive):
         """
         return Vector(self.x ** n, self.y ** n, self.z ** n)
 
+    def __neg__(self):
+        return self.scaled(-1.0)
+
     def __iadd__(self, other):
         """Add the components of the other vector to this vector.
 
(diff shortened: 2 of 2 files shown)


tomvanmele (approved):

tomvanmele:
LGTM
can you add an entry in the changelog?

src/compas/geometry/primitives/vector.py
```diff
@@ -676,6 +679,39 @@ def unitized(self):
...
+
+        Note
+        ____
+        a negation of a vector is similar to inverting a vector
+
+        Examples
+        --------
+        >>> u = Vector(1.0, 0.0, 0.0)
+        >>> v = u.copy()
+        >>> u.invert()
```
tomvanmele, line 693: perhaps add here
```python
>>> u == v
False
```
otherwise the test is also correct if basically nothing happens :)

tomvanmele:
what happens in the case that
```python
u = Vector(0, 0, 0)
v = -u
u == v
```

gonzalocasas:
cool! lgtm

I was surprised the github action that checks for changelog updates did not raise any errors. I took a closer look and it seems to be failing with an odd error but is totally unrelated to this PR:
image

gonzalocasas (approved):

romanarust:
yay!

src/compas/geometry/primitives/vector.py
jf---: great catch, added...

jf---:
> what happens in the case that
> 
> ```python
> u = Vector(0, 0, 0)
> v = -u
> u == v
> ```

Shoot infinite has a sign

```python
In [1]: import math

In [2]: math.inf
Out[2]: inf

In [3]: math.inf == -math.inf
Out[3]: False
```

Hmmm, no idea frankly...

> LGTM
> can you add an entry in the changelog?

Sure

CHANGELOG.md
```diff
@@ -6,10 +6,13 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 
-## Unreleased
+## Unreleased [1.2.1] 2021-xx-yy
```
beverlylytle: I don't think this should be added.

jf---:
> LGTM
> can you add an entry in the changelog?

CHANGELOG.md
jf---: Back to `## Unreleased`

PR #797, opened 2021-03-24
Upstream Configuration

beverlylytle:
`Configuration` believe the grass is greener over here in `compas`.  

### Checklist

### DIFF
--- a/src/compas/robots/__init__.py
+++ b/src/compas/robots/__init__.py
@@ -9,7 +9,7 @@ robots
 Model
 =====
 
-.. inheritance-diagram:: RobotModel Joint Link ToolModel
+.. inheritance-diagram:: RobotModel Joint Link ToolModel Configuration
     :parts: 1
 
 The root of the model is the :class:`RobotModel` class, which
@@ -24,6 +24,7 @@ elements connecting the links together.
     Joint
     Link
     ToolModel
+    Configuration
 
 
 Geometric description
@@ -115,6 +116,10 @@ the processing of these resources.
 """
 from __future__ import absolute_import
 
+from .configuration import (
+    Configuration,
+    FixedLengthList
+)
 from .model import (
     Axis,
     Box,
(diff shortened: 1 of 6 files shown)


src/compas/robots/configuration.py
```diff
@@ -0,0 +1,614 @@
...
+        joint_types = [_type_dict[name] for name in joint_names]
+
+        return Configuration(joint_values, joint_types, joint_names)
+
+
+if __name__ == "__main__":
+    import math  # noqa F401
+    from compas.geometry import allclose   # noqa F401
+    import doctest
+    doctest.testmod(globs=globals())
```
gonzalocasas: I think we can get rid of these blocks

src/compas/robots/__init__.py
```diff
@@ -115,6 +116,10 @@
 """
 from __future__ import absolute_import
 
+from .configuration import (
+    Configuration,
+    FixedLengthList
```
gonzalocasas: Should we expose this class or rather keep it an implementation detail?

src/compas/robots/model/robot.py
```diff
@@ -704,7 +766,7 @@ def compute_transformations(self, joint_state, link=None, parent_transformation=
 
         Parameters
         ----------
-        joint_state : dict
+        joint_state : dict or :class:`compas.robots.Configuration`
```
gonzalocasas: totally minor detail, but I'd swap the order of these to make `Configuration` the perceived preferred option

gonzalocasas (approved):
LGTM! We've been through the details of the impl on compas_fab repo, so I only checked the stuff outside the config class.

src/compas/robots/__init__.py
beverlylytle: Yeah, this is tricky... I use it in `JointTrajectoryPoint`.

beverlylytle: Should I make it available as a third level import rather than second?

gonzalocasas: yes, that sounds like a good alternative

PR #803, opened 2021-03-30
Update offset module

tomvanmele:


src/compas/datastructures/mesh/offset.py
```diff
@@ -85,4 +130,4 @@ def mesh_thicken(mesh, thickness=1.0):
 if __name__ == '__main__':
 
     import doctest
-    doctest.testmod(globs=globals())
+    doctest.testmod()
```
gonzalocasas, line 133: I think we can get rid of this whole `if __name__ ...` thing, right?

gonzalocasas (approved):
LGTM!

src/compas/datastructures/mesh/offset.py
tomvanmele: am testing it now to see if sphinx doctesting still works (and fixing the doctests while at it). will send separate PR for that...

PR #807, opened 2021-04-07
use custom action

Licini:


gonzalocasas:
cool! would it be possible to have these custom actions be more "modular", ie. so that we can add the "lint" step, or the "test" step, or the "deploy docs" step, etc, instead of all-or-nothing? Because this should be reusable by other compas extensions and not all actions will fit exactly to all extensions.

.github/workflows/build.yml
```diff
@@ -23,19 +23,5 @@ jobs:
...
-        run: |
-          invoke lint
-      - name: Test import
-        run: |
-          python -m compas
-      - name: Test with pytest
-        run: |
-          invoke test
+
+      - uses: actions/compas-actions.build@v1
```
beverlylytle:
```suggestion
      - uses: compas-dev/compas-actions.build@v1
```

Licini: Ha! Thanks, was confused by their example

Licini:
> cool! would it be possible to have these custom actions be more "modular", ie. so that we can add the "lint" step, or the "test" step, or the "deploy docs" step, etc, instead of all-or-nothing? Because this should be reusable by other compas extensions and not all actions will fit exactly to all extensions.

Yes definitely. at least for this one I will add option to whether run lint or test respectively. Two major limitations I learned today on making custom actions is that:
1. You cannot use `uses` field to take advantage of other custom actions like checkout and publish we currently have in the workflow, unless we make our own implementation of them.
2.  Each custom action has to be its own repo, so we will have to have separate repos like `compas-build.docs`, `compas-build.release` etc.

Still trying to experiment out a most flexible yet generalized way to set this up. Until then you guys might continue to receive fail build warnings : )

Licini:
Alright I think this PR is finally ready for review. I added three custom actions:
- [compas-actions.build](https://github.com/compas-dev/compas-actions.build/blob/main/action.yml)
- [compas-actions.docs](https://github.com/compas-dev/compas-actions.docs/blob/main/action.yml)
- [compas-actions.docversions](https://github.com/compas-dev/compas-actions.docversions/blob/main/action.yml)

Each of them have few input parameters (see in their action files) that allows certain degrees of general usage for other compas packages. Let me know what you guys think or if I missed anything. If everything looks ok I will publish v1 for each actions.

tomvanmele (approved):
LGTM (i think, bit difficult to say :)

PR #809, opened 2021-04-13
Fix to support all read modes

gonzalocasas:
No changelog entry because is a fix for a new feature that has not been released yet.

### Checklist

### DIFF
--- a/src/compas/_iotools.py
+++ b/src/compas/_iotools.py
@@ -42,8 +42,9 @@ def open_file(file_or_filename, mode='r'):
     if not hasattr(file, 'read'):
         # If it's a URL, open and read its response into an in-memory stream
         if hasattr(file, 'startswith') and file.startswith('http'):
-            if mode != 'r':
-                raise ValueError('URLs can only be opened in read mode')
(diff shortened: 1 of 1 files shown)


src/compas/_iotools.py
```diff
@@ -42,8 +42,9 @@ def open_file(file_or_filename, mode='r'):
     if not hasattr(file, 'read'):
         # If it's a URL, open and read its response into an in-memory stream
         if hasattr(file, 'startswith') and file.startswith('http'):
-            if mode != 'r':
-                raise ValueError('URLs can only be opened in read mode')
+            # support all read modes (r, r+, rb, etc)
+            if not mode.startswith('r'):
```
beverlylytle, line 46: mode `'r+'` starts with 'r' but also is writable.  should that be allowed?

gonzalocasas: I would say so. Especially for the cases in which `r+` is used, but never upgraded for write (one could claim then `r+` is being used incorrectly but it's not on us to judge that 😉)

beverlylytle (approved):
👍

PR #818, opened 2021-04-27
Grasshopper components!

gonzalocasas:
This PR adds all the scaffolding needed to build Grasshopper components for COMPAS directly from source code (instead of binary blobs). It also adds one first component, but more as a proof of concept than anything else, the component outputs compas version and active environment.

This is how it looks:
image

Currently the components are built using a `pyinvoke` task and I added them to the ironpython build in Github Actions but (as agreed with Tom) we should change the release step of `pyinvoke` to release based on version tags and build the gh components in that step (see here for reference: https://github.com/compas-dev/compas_fab/pull/294).

### Checklist


src/compas_ghpython/__init__.py
```diff
@@ -14,14 +14,40 @@
 """
 import os
 import compas
+import compas_rhino
 
+# TODO: I believe this should be removed, as it pulls all utilities funcs
```
tomvanmele, line 19: agreed. perhaps we can do 1 big cleanup for the release of 2.0

src/compas_ghpython/components/Compas_Info/code.py
```diff
@@ -0,0 +1,14 @@
...
+Displays information about the active COMPAS environment.
+"""
+import compas_bootstrapper
+from ghpythonlib.componentbase import executingcomponent as component
+
+import compas
+
+
+class CompasInfo(component):
+    def RunScript(self):
```
tomvanmele, line 11: this is actually very similar to how the python command plugins work for rhino (used for example in rv2) perhaps we should formalize that mechanism as well, and especially add a base plugin for rhino from which it is easier to do compas related infrastructure stuff. like a default compas menu for rhino...

tomvanmele (approved):

src/compas_ghpython/components/Compas_Info/code.py
gonzalocasas: Added it as topic for the next dev meeting!

src/compas_ghpython/__init__.py
gonzalocasas: 👍

PR #827, opened 2021-05-03
Check API doc completeness

tomvanmele:

### Checklist


tomvanmele:
@Licini this works locally but fails online. any ideas?

beverlylytle:
About the failing workflows, it looks like the folder `docs/api/generated` in the api test doesn't exist yet as the docs haven't been generated in the "build" workflow.  Maybe "compas-dev/compas-actions.docs" needs to be included in the build workflow, but that would slow things down quite a lot.

Licini:
> @Licini this works locally but fails online. any ideas?

@beverlylytle was right about it

Licini:
perhaps we can make this a separate test, seems this particular one dose not need to be tested in the matrix of platforms?

brgcode:
a separate API test/workflow would make sense indeed. @Licini could you add something along those lines?

tomvanmele:
in addition, we don't need to generate the complete docs for this. we just need the auto generated stubs. those can also be generated with 
```bash
sphinx-autogen docs/api/compas.**.rst -o temp/stubs
```

tomvanmele:
@Licini i think it is working now, but could you still make a separate workflow for this because it indeed only needs to be checked once. and not for the entire matrix...

gonzalocasas:
@brgcode ah, i think the ironpython-testrunner is evaluating fixtures even before they are needed, so, to fix the build fail on ipy you probably need to skip (`if compas.IPY: return`) inside the fixtures as well.

PR #829, opened 2021-05-03
Optimize rhino mesh to compas conversion

gonzalocasas:
Converting large Rhino meshes to COMPAS meshes is pretty slow. This optimization is rather simple but shaves about half the runtime cost of the conversion. Further optimization is possible if we use a custom `add_vertex` and `add_face` with less checking, as pointed out by @brgcode on Slack.

### Checklist


src/compas_rhino/geometry/mesh.py
```diff
@@ -100,16 +100,19 @@ def to_compas(self, cls=None):
...
         cls = cls or Mesh
-        faces = []
-        for face in self.faces:
-            if face[0] == face[-1]:
-                faces.append(face[:-1])
-            elif face[-2] == face[-1]:
-                faces.append(face[:-1])
+        mesh = cls()
+
+        for key, vertex in enumerate(self.geometry.Vertices):
```
tomvanmele: enumeration and key assignment are not necessary because 0-based continuous integers are the default behaviour

src/compas_rhino/geometry/mesh.py
```diff
@@ -100,16 +100,19 @@ def to_compas(self, cls=None):
...
+            mesh.add_vertex(key=key, attr_dict=dict(x=float(vertex.X), y=float(vertex.Y), z=float(vertex.Z)))
+
+        for fkey, face in enumerate(self.geometry.Faces):
```
tomvanmele: enumeration and key assignment are superfluous...

src/compas_rhino/geometry/mesh.py
gonzalocasas: true! fixed at https://github.com/compas-dev/compas/pull/829/commits/6f7c98532300e9e39037f667c3466127a9f1b192

tomvanmele (approved):

PR #839, opened 2021-05-24
Update data module and add schemas

tomvanmele:

### Checklist

### DIFF
--- a/schemas/line.json
+++ b/schemas/line.json
@@ -1,6 +1,6 @@
 {
     "$schema": "http://json-schema.org/draft-07/schema#",
-    "$id": "https://github.com/compas-dev/compas/schemas/frame.json",
+    "$id": "https://github.com/compas-dev/compas/schemas/line.json",
     "$compas": "1.6.2",
 
     "type": "object",
--- a/schemas/line.json
+++ b/schemas/line.json
@@ -1,0 +1,12 @@
+{
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "$id": "https://github.com/compas-dev/compas/schemas/frame.json",
+    "$compas": "1.6.2",
+
+    "type": "object",
+    "properties": {
+        "start": {"type": "array", "minItems": 3, "maxItems": 3, "items": {"type": "number"}},
+        "end": {"type": "array", "minItems": 3, "maxItems": 3, "items": {"type": "number"}}
+    },
+    "required": ["start", "end"]
+}
--- a/src/compas/geometry/primitives/ellipse.py
+++ b/src/compas/geometry/primitives/ellipse.py
@@ -64,6 +64,43 @@ class Ellipse(Primitive):
         self.minor = minor
 
     @property
+    def DATASCHEMA(self):
+        import schema
+        return schema.Schema({
+            "plane": schema.And(
+                lambda x: len(x[0]) == 3 and all(isinstance(i, float) for i in x[0]),
+                lambda x: len(x[1]) == 3 and all(isinstance(i, float) for i in x[1])
+            ),
+            "major": schema.And(float, lambda x: x > 0),
+            "minor": schema.And(float, lambda x: x > 0),
+        })
+
+    @property
+    def JSONSCHEMA(self):
+        from compas import versionstring
+        schema = {
+            "$schema": "http://json-schema.org/draft-07/schema#",
+            "$id": "https://github.com/compas-dev/compas/schemas/ellipse.json",
+            "$compas": versionstring,
+            "type": "object",
+            "properties": {
+                "plane": {
+                    "type": "array",
+                    "minItems": 2,
+                    "maxItems": 2,
+                    "items": [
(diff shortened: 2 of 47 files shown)


tomvanmele:
the changes here go a bit hand-in-hand with what is needed for the exchange meeting
https://docs.google.com/presentation/d/13wWEFO8oJKwYUQ1f2i0UaazT1gVJAYEowxHF0nsnOeQ/edit?usp=sharing

src/compas/geometry/primitives/ellipse.py
```diff
@@ -63,6 +63,43 @@ def __init__(self, plane, major, minor):
         self.major = major
         self.minor = minor
 
+    @property
+    def DATASCHEMA(self):
+        import schema
+        return schema.Schema({
+            "plane": schema.And(
+                lambda x: len(x[0]) == 3 and all(isinstance(i, float) for i in x[0]),
```
gonzalocasas: since this will be a veeeery common check, should we have a `compas.data.validations.is_vector3` helper function or similar?

gonzalocasas (approved):
It's looking good to me so far!

Licini:
Looks good to me too!

schemas/line.json
```diff
@@ -0,0 +1,12 @@
+{
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "$id": "https://github.com/compas-dev/compas/schemas/frame.json",
```
beverlylytle:
```suggestion
    "$id": "https://github.com/compas-dev/compas/schemas/line.json",
```

src/compas/geometry/primitives/quaternion.py
```diff
@@ -237,7 +260,7 @@ def __len__(self):
         return 4
 
     def __repr__(self):
-        return 'Quaternion({:.{prec}f}, {:.{prec}f}, {:.{prec}f}, {:.{prec}f})'.format(self.w, self.x, self.y, self.z, prec=3)
+        return "Quaternion({:.{prec}f}, {:.{prec}f}, {:.{prec}f}, {:.{prec}f})".format(self.w, self.x, self.y, self.z, prec=3)
```
gonzalocasas, line 263: haven't we mostly settled for single quote instead of the double ones that this and other files now use?

src/compas/geometry/primitives/quaternion.py
```diff
@@ -123,6 +123,40 @@ def __init__(self, w, x, y, z):
...
 
+    @property
+    def DATASCHEMA(self):
+        from schema import Schema
+        return Schema({"w": float, "x": float, "y": float, "z": float})
+
+    @property
+    def JSONSCHEMA(self):
+        from compas import versionstring
+        schema = {
```
gonzalocasas, line 134: not to change in this PR, but shouldn't these things be loaded directly from the files instead of inlined/repeated here?

tomvanmele: ideally yes but somehow haven't found a builtin way for making that work

src/compas/geometry/transformations/rotation.py
```diff
@@ -72,6 +72,9 @@ def __init__(self, matrix=None):
                 raise ValueError('This is not a proper rotation matrix.')
         super(Rotation, self).__init__(matrix=matrix)
 
+    # def __repr__(self):
+    #     return "Rotation({0})".format(self.matrix)
```
gonzalocasas, line 76: why is this commented out?

tests/compas/geometry/test_point.py
```diff
@@ -1,8 +1,20 @@
+import compas
 from compas.geometry import Point
 
 
+def test_data():
+    if compas.IPY:
+        return
```
gonzalocasas, line 7: it would be much much better if we don't skip ipy here... 😇

gonzalocasas (approved):
LGTM

src/compas/geometry/transformations/rotation.py
tomvanmele: it doesn't work in all cases because the init of rotation checks for a true/pure rotation matrix. so if you specify a base point for the rotation it fails...

tests/compas/geometry/test_point.py
tomvanmele: i know but `schema` is not available for ironpython...

src/compas/geometry/primitives/quaternion.py
tomvanmele: we could add `compas.schema(...)`?

tomvanmele: similar to `compas.get(...)`

src/compas/geometry/primitives/quaternion.py
tomvanmele: not sure. seems pretty random to me at the moment... but perhaps we can decide on something and make a source wide change in a separate PR?

PR #844, opened 2021-05-28
Load jsonschema from file

tomvanmele:

### Checklist


tomvanmele:
@Licini could you check if the modification i made to the setup file and manifest correctly include the schemas in the distribution?

tomvanmele:
@gonzalocasas this should address the duplication of schemas in code

src/compas/data/data.py
```diff
@@ -48,20 +49,46 @@ class Data(object):
...
+        self._jsonschema = None
+        self._jsonvalidator = None
 
     def __str__(self):
         """Generate a readable representation of the data of the object."""
         return json.dumps(self.data, sort_keys=True, indent=4)
 
     @property
-    def DATASCHEMA(self):
+    def dataschema(self):
```
gonzalocasas: I would prefer not to introduce a breaking change with this lower-casing, considering schemes are consts, it is fair to have them upper-cased.

CHANGELOG.md
```diff
@@ -490,15 +490,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-* Added abstract `DATASCHEMA` to `compas.base.Base`.
-* Added abstract `JSONSCHEMA` to `compas.base.Base`.
+* Added abstract `dataschema` to `compas.base.Base`.
```
gonzalocasas: Retroactively changing the changelog history is a dangerous thing.

tomvanmele:
ping pang pong

src/compas/data/validators.py
```diff
@@ -40,6 +50,21 @@ def validate_data(data, cls):
...
+
+    definitions_path = os.path.join(here, 'schemas', 'compas.json')
+    with open(definitions_path, 'r') as fp:
+        definitions = json.load(fp)
+
+    resolver = RefResolver.from_schema(definitions)
+    validator = Draft7Validator(schema, resolver=resolver)
+    validator.validate(data)
+
+    return json.loads(json.dumps(data, cls=DataEncoder), cls=DataDecoder)
```
beverlylytle, line 150: The whole point of the validation is so that this loads/dumps combo will raise some exception if the `data` isn't valid, right?  Maybe it would be nice to wrap this in a `try/except` statement which would basically return the error that `json` raises, but wrapped in something that says "Sorry, compas validation failed"

Licini:
@brgcode Regarding the manifest, it does package the schemas correctly into the build. Just tried `pip install .` and I can see them installed into the `site_packages` folder
image

src/compas/_os.py
```diff
@@ -241,6 +243,14 @@ def absjoin(*parts):
     return os.path.abspath(os.path.join(*parts))
 
 
+# def here(sibbling, name):
+#     return os.path.join(os.path.dirname(sibbling), name)
+
+
+def here(name):
```
gonzalocasas: I know we use the `HERE` constant always, but a `here` method that gets imported from "there" does not convey the same clear purpose. From the unix tradition, I would tend to call this `pwd` (even if the p stands for "print", but still, it's a well-known command), so a more verbose alternative would be `get_current_dir`. And that actually points out a bit of a semantic discrepancy with the idea of the method and the actual behavior. If it would be called `get_current_dir` but actually returns a sub-dir of the current, then it's clear that the name is wrong. So perhaps this should be called `get_relative_path` or `get_relative_dir` (or the variants, like `get_rel_dir`)?

src/compas/data/data.py
```diff
@@ -36,33 +37,98 @@ class Data(object):
...
         return json.dumps(self.data, sort_keys=True, indent=4)
 
+    def __getstate__(self):
+        """Return the object data for state serialization with older pickle protocols."""
+        return {'__dict__': self.__dict__.copy(), 'dtype': self.dtype, 'data': self.data}
+
+    def __setstate__(self, state):
+        """Assign a deserialized state to the object data to support older pickle protocols."""
+        self.__dict__.update(state['__dict__'])
+        self.data = state['data']
```
gonzalocasas: These two methods need to be unit tested.

src/compas/data/data.py
```diff
@@ -36,33 +37,98 @@ class Data(object):
...
         """:class:`schema.Schema` : The schema of the data of this object."""
         raise NotImplementedError
 
     @property
-    def JSONSCHEMA(self):
+    def jsonschema(self):
         """dict : The schema of the JSON representation of the data of this object."""
+        if not self._jsonschema:
+            schema_name = '{}.json'.format(self.__class__.__name__.lower())
+            schema_path = os.path.join(os.path.dirname(__file__), 'schemas', schema_name)
```
gonzalocasas: Is there a case in which `os.path.dirname(__file__)` will not be `data`?

src/compas/data/schemas/circle.json
```diff
@@ -10,8 +10,8 @@
             "minItems": 2,
             "maxItems": 2,
             "items": [
-                {"type": "array", "minItems": 3, "maxItems": 3, "items": {"type": "number"}},
-                {"type": "array", "minItems": 3, "maxItems": 3, "items": {"type": "number"}}
+                {"$ref": "compas.json#/definitions/point"},
```
gonzalocasas: I think it would be a good idea to add unit tests that validate the schemas themselves, especially now that they become non-trivial with references to other docs and so on, it will be very easy to break them without noticing.

src/compas/data/schemas/graph.json
```diff
@@ -1,6 +1,6 @@
 {
     "$schema": "http://json-schema.org/draft-07/schema#",
-    "$id": "https://github.com/compas-dev/compas/schemas/graph.json",
+    "$id": "https://github.com/compas-dev/compas/raw/main/compas/data/schemas/graph.json",
```
gonzalocasas: shouldn't it be just `graph.json` as the other ones?

src/compas/data/schemas/halfedge.json
```diff
@@ -0,0 +1,20 @@
+{
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "$id": "https://github.com/compas-dev/compas/raw/main/compas/data/schemas/halfedge.json",
```
gonzalocasas: also here, shouldn't it be just `halfedge.json`?

src/compas/data/schemas/halfface.json
```diff
@@ -1,6 +1,6 @@
 {
     "$schema": "http://json-schema.org/draft-07/schema#",
-    "$id": "https://github.com/compas-dev/compas/schemas/halfface.json",
+    "$id": "https://github.com/compas-dev/compas/raw/main/compas/data/schemas/halfface.json",
```
gonzalocasas: and one more

src/compas/data/validators.py
```diff
@@ -2,11 +2,21 @@
...
+
+from compas.data.encoders import DataEncoder
+from compas.data.encoders import DataDecoder
+
 
 def is_int3(items):
     return len(items) == 3 and all(isinstance(item, int) for item in items)
 
 
+def is_intx(items):
```
gonzalocasas: why not `are_ints` or `is_int_list`?

gonzalocasas (changes requested):

CHANGELOG.md
tomvanmele: side effect of find-replace :)

src/compas/data/data.py
tomvanmele: not so far, but perhaps i could replace by `os.path.dirname(compas.data.__file__)`?

src/compas/data/schemas/circle.json
tomvanmele: yes! will send a few more updates, including tests

src/compas/_os.py
tomvanmele: but perhaps i will just remove for now. did not really mean for this to be part of the PR. was just experimenting with making this "here" thing a bit more convenient...

src/compas/data/data.py
tomvanmele: i think the lowercase class name is probably more problematic. it for example already doesn't work for mesh :)

tomvanmele:
@gonzalocasas this is ready i think

tomvanmele:
ping pang pong

tomvanmele:
@gonzalocasas @beverlylytle

src/compas/data/data.py
```diff
@@ -45,24 +39,87 @@ class Data(object):
...
+        if not self._JSONSCHEMA:
+            schema_filename = '{}.json'.format(self.JSONSCHEMANAME.lower())
+            schema_path = os.path.join(os.path.dirname(__file__), 'schemas', schema_filename)
+            with open(schema_path, 'r') as fp:
+                self._JSONSCHEMA = json.load(fp)
+        return self._JSONSCHEMA
+
+    @property
+    def jsondefinitions(self):
+        """dict : Reusable schema deinitions."""
```
beverlylytle:
```suggestion
        """dict : Reusable schema definitions."""
```

src/compas/data/schemas/box.json
```diff
@@ -0,0 +1,29 @@
+{
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "$id": "box.json",
+    "$compas": "1.6.3",
```
beverlylytle: shouldn't all of these be changed to `"1.7.1"`?

src/compas/datastructures/mesh/core/halfedge.py
```diff
@@ -49,100 +42,44 @@ class HalfEdge(Datastructure):
...
-                "max_face": schema.And(int, lambda x: x >= -1)
-            }
+            "attributes": dict,
+            "dva": dict,
+            "dea": dict,
+            "dfa": dict,
+            "vertex": schema.And(
+                dict,
+                lambda x: all(isinstance(key, int) for key in x),
+                lambda x: all(key >= 0 for key in x),
```
beverlylytle: This looks like it gets reused often.  Maybe adding a `is_sequence_of_indices` method would be easier to read.

src/compas/geometry/pointclouds/pointcloud.py
```diff
@@ -16,18 +16,30 @@
...
     """Class for working with pointclouds."""
 
-    def __init__(self, points):
-        super(Pointcloud, self).__init__()
+    @property
+    def DATASCHEMA(self):
+        from schema import Schema
+        from compas.data import is_float3
+        return Schema({
+            'points': lambda x: [is_float3(i) for i in x]
```
beverlylytle: all?

src/compas/geometry/primitives/polygon.py
```diff
@@ -76,45 +67,34 @@ class Polygon(Primitive):
...
-        self._lines = []
-        self.points = points
-
     @property
     def DATASCHEMA(self):
         from schema import Schema
         from compas.data import is_float3
         return Schema({
-            "points": lambda x: [is_float3(i) for i in x]
+            'points': lambda x: [is_float3(i) for i in x]
```
beverlylytle: all?

src/compas/geometry/primitives/polyline.py
```diff
@@ -64,36 +55,25 @@ class Polyline(Primitive):
...
-        self._lines = []
-        self.points = points
-
     @property
     def DATASCHEMA(self):
         from schema import Schema
         from compas.data import is_float3
         return Schema({
-            "points": lambda x: [is_float3(i) for i in x]
+            'points': lambda x: [is_float3(i) for i in x]
```
beverlylytle: all?

src/compas/data/schemas/box.json
tomvanmele: yes indeed

tomvanmele:
@Licini @beverlylytle @gonzalocasas sorry to keep pushing for this...

src/compas/geometry/pointclouds/pointcloud.py
```diff
@@ -158,6 +158,25 @@ def __setitem__(self, key, value):
...
+        other : :class:`compas.geometry.Pointcloud` or list
+            The pointcloud to compare.
+
+        Returns
+        -------
+        bool
+            True if the pointclouds are equal.
+            False otherwise.
+        """
+        return all(a == b for a, b in zip(self, other))
```
beverlylytle: This doesn't check that the pointclouds have the same number of points.

tomvanmele: true. will update

beverlylytle: Also, does the order of the points in a pointcloud really matter?

tomvanmele: no it doesn't, but couldn't think of a different comparison. we could compare max distance between nearest neighbours?

beverlylytle: Yeah, I think lexicographically sorting the points and then comparing them pairwise would make sense here.

tomvanmele: could do something like this
```python
import random
from compas.geometry import Pointcloud
a = Pointcloud.from_bounds(10, 10, 10, 10)
points = a.points[:]
random.shuffle(points)
b = Pointcloud(points)
a_points = sorted(a.points, key=lambda point: (point.x, point.y, point.z))
b_points = sorted(b.points, key=lambda point: (point.x, point.y, point.z))
```
and then compare?

beverlylytle: yeah, that looks good.

tomvanmele:
@gonzalocasas @beverlylytle sorry i know i am being pushy, but is this now good to go?

beverlylytle (approved):
LGTM!

Issue #848, opened 2021-06-07
bestfit_circle_numpy returns wrong result


yck011522:
**Describe the bug**
I was trying the function `bestfit_circle_numpy` but I found it produces erratic results.
I created a function to generate some test points and test the function but I cannot figure out the pattern of how it went wrong.

**To Reproduce**
```
import math
import random

import compas
from compas.geometry import Point, add_vectors, bestfit_circle_numpy, circle_evaluate, distance_point_point

print("compas.__version__ = %s" % compas.__version__)


def test_bestfit_circle_numpy(input_radius, real_center, number_of_points, equal_sample_space):
    # Generate Points
    if equal_sample_space:
        parameters = [t / number_of_points * 2 * math.pi for t in range(number_of_points)]
    else:
        random.seed(42)
        parameters = [random.random() * 2 * math.pi for _ in range(number_of_points)]

    input_points = [circle_evaluate(t, input_radius) for t in parameters]
    input_points = [add_vectors(pt, real_center) for pt in input_points]
    print("input_radius  = %s" % input_radius)
    print("input_points  = %s" % input_points)

    center, normal, radius = bestfit_circle_numpy(input_points)

    print("center = %s" % center)
    print("normal = %s" % normal)
    print("radius = %s" % radius)

    distances = [distance_point_point(center, point) for point in input_points]
    print("input_points distance to center: %s" % distances)
    print("--- --- ---")


# All samples are centered around zero and
test_bestfit_circle_numpy(100, [10, 10, 10], 3, True)
test_bestfit_circle_numpy(100, [10, 10, 10], 4, True)
test_bestfit_circle_numpy(100, [10, 10, 10], 9, True)

test_bestfit_circle_numpy(100, [10, 10, 10], 3, False)
test_bestfit_circle_numpy(100, [10, 10, 10], 4, False)
test_bestfit_circle_numpy(100, [10, 10, 10], 9, False)
```
**My result**
```
compas.__version__ = 1.5.0
input_radius  = 100
input_points  = [[110.0, 10.0, 10], [-39.99999999999998, 96.60254037844388, 10], [-40.00000000000004, -76.60254037844385, 10]]
center = [10. 10. 10.]
normal = [0. 0. 1.]
radius = 64.39505508593787
input_points distance to center: [100.0, 99.99999999999999, 100.00000000000001]
--- --- ---
input_radius  = 100
[...]

PR #851, opened 2021-06-08
Avoid stable points

beverlylytle:
This is a continuation of the fix for #848 See the comment in code for an explanation.

### Checklist

### DIFF
--- a/src/compas/geometry/bestfit/bestfit_numpy.py
+++ b/src/compas/geometry/bestfit/bestfit_numpy.py
@@ -7,10 +7,9 @@ from numpy import sqrt
 from numpy import mean
 from numpy import zeros
 from numpy.linalg import lstsq
-from numpy.linalg import norm
 from scipy.optimize import leastsq
 
-from compas.geometry import world_to_local_coordinates_numpy, allclose
+from compas.geometry import world_to_local_coordinates_numpy
 from compas.geometry import local_to_world_coordinates_numpy
 from compas.numerical import pca_numpy
 
--- a/src/compas/geometry/bestfit/bestfit_numpy.py
+++ b/src/compas/geometry/bestfit/bestfit_numpy.py
@@ -124,7 +124,7 @@ def bestfit_circle_numpy(points):
 
     # The addition of epsilon is to promote convergence within leastsq,
     # which seems to strongly dislike (0, 0) as an initial guess.
-    epsilon = .000001
+    epsilon = .00001
     xm = mean(x) + epsilon
     ym = mean(y) + epsilon
     c0 = xm, ym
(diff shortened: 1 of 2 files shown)


beverlylytle:
I blame Fastly for the builds failing.

src/compas/geometry/bestfit/bestfit_numpy.py
```diff
@@ -122,8 +122,11 @@ def f(c):
         Ri = dist(*c)
         return Ri - Ri.mean()
 
-    xm = mean(x)
-    ym = mean(y)
+    # The addition of epsilon is to promote convergence within leastsq,
+    # which seems to strongly dislike (0, 0) as an initial guess.
+    epsilon = .00001
+    xm = mean(x) + epsilon
+    ym = mean(y) + epsilon
```
tomvanmele: this makes more sense to me than the previous version, but what if `mean(x) == -epsilon` and `mean(y) == -epsilon` :)

tomvanmele:
the tests fail because the location from where the bunny is downloaded in the tests seems often unavailable lately. i have changed this test in the PR about json schemas, but that one is still pending. perhaps it should be move to a different PR...

src/compas/geometry/bestfit/bestfit_numpy.py
beverlylytle: yeah... so `mean(x)` empirically is on the order of e-15. i'm not sure that i could prove that is always the case, but the definition of `x` sure makes it seem like `mean(x)` will always be quite tiny. in my experiments, an epsilon of `0.0000001` would reveal the instability, while `0.000001` didn't.  so i went with `.00001`. i could just go with `xm = ym = epsilon` or even `xm = ym =  1`.  1 takes a smidge longer to converge, but it would work.

beverlylytle:
The site probably crashes because of all the tests trying to pull the bunny.

src/compas/geometry/bestfit/bestfit_numpy.py
```diff
@@ -122,8 +122,10 @@ def f(c):
         Ri = dist(*c)
         return Ri - Ri.mean()
 
-    xm = mean(x)
-    ym = mean(y)
+    # The mean of x and y are very nearly 0 (1.0e-15), which reveals a numerical
+    # instability of the problem.  So, we choose our initial guess
+    # to be an epsilon bigger than that.
+    xm = ym = 0.00001
```
tomvanmele, line 127: i guess that makes a lot of sense indeed because of the transformation to the local coordinate system of the points

tomvanmele (approved):

src/compas/geometry/bestfit/bestfit_numpy.py
beverlylytle: I've just been playing with this value a bit.  1e-8 and smaller gives crazy results, except 0.  0 and anything greater than 1e-7 seems to be fine.  But I will leave it as it is.

PR #852, opened 2021-06-08
Install function not available from compas_rhino

tetov:

- Fix code example


tomvanmele (commented):
should we not rather pull up the install function?

tetov:
@brgcode:
> should we not rather pull up the install function?

if we do
```
from .install import install
```
in `__init__.py` then that would be a name conflict with module install. The module is called when installing from the command line.
```
from .install import install as install_function
```
?

tomvanmele (approved):

PR #856, opened 2021-06-09
fix closest_point_on_segment_xy

### LINKED ISSUE #837: closest_point_on_segment_xy returns a 3D point (Z corrdinate !=0)
**Describe the bug**

If the `closest_point_on_segment_xy` is one of the endpoints, it is returned with z coordinate different than 0.
This is because the [logic returns the endpoint as-is](https://github.com/compas-dev/compas/blob/961b15e7560d2b450e6caaa77ad5676af4103beb/src/compas/geometry/_core/distance.py#L766), without zeroing the Z coordinate.

**To Reproduce**

```py
from compas.geometry import Point, Line, closest_point_on_segment_xy


def test_closest_point_segment_xy():
    point = Point(0, 0, -10)
    line = Line([1, 1, -15], [10, 10, 20])
    ponl= closest_point_on_segment_xy(point, line)
    assert ponl == Point(1, 1, 0)
```

it returns

```
=============================== FAILURES ================================
_________________ test_closest_point_segment_xy ________________________

    def test_closest_point_segment_xy():
        point = Point(0, 0, -10)
        line = Line([1, 1, -15], [10, 10, 20])
        ponl = closest_point_on_segment_xy(point, line)
>       assert ponl == Point(1, 1, 0)
E       assert Point(1.000, 1.000, -15.000) == Point(1.000, 1.000, 0.000)
E         Use -v to get the full diff
```

sanzoghenzo:
fixes #837

### Checklist

### DIFF
--- a/tests/compas/geometry/test_distance.py
+++ b/tests/compas/geometry/test_distance.py
@@ -1,10 +1,14 @@
+import pytest
+
 from compas.geometry import Point
 from compas.geometry import Line
 from compas.geometry import closest_point_on_segment_xy
 
 
-def test_closest_point_segment_xy():
-    point = Point(0, 0, -10)
+@pytest.mark.parametrize(
+    "point,point_on_line", [[[0, 0, -10], [1, 1, 0]], [[5, 4, 80], [4.5, 4.5, 0]]]
+)
+def test_closest_point_segment_xy(point, point_on_line):
     line = Line([1, 1, -15], [10, 10, 20])
-    ponl = closest_point_on_segment_xy(point, line)
-    assert ponl == Point(1, 1, 0)
+    ponl = closest_point_on_segment_xy(Point(*point), line)
+    assert ponl == Point(*point_on_line)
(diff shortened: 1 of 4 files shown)


tests/compas/geometry/test_distance.py
```diff
@@ -0,0 +1,10 @@
+from compas.geometry import Point
+from compas.geometry import Line
+from compas.geometry import closest_point_on_segment_xy
+
+
+def test_closest_point_segment_xy():
```
tomvanmele: would you mind including non-boundary cases as well?

tomvanmele (commented):
cool, thanks for contributing!
just a small request about non-boundary cases....

tests/compas/geometry/test_distance.py
sanzoghenzo: There you have it, I parametrized the test and added a non boundary case.

tomvanmele (approved):

PR #866, opened 2021-06-21
fix dr_numpy for np array inputs

Licini:

### Checklist

When inputs like `qpre` `fpre` `lpre` are given as numpy array, direct evaluation of these variables will throw error below:
```
Traceback (most recent call last):
  File "d:/Github/compas_cloud/examples/dr_numpy_test.py", line 99, in <module>
    kmax=100, callback=callback)
  File "d:\github\compas\src\compas\numerical\dr\dr_numpy.py", line 119, in dr_numpy
    qpre = qpre or [0.0 for _ in range(num_e)]
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

The fix is to check explicitly if these variables are `None`, if so then initiate them.

### DIFF
--- a/src/compas/numerical/dr/dr_numpy.py
+++ b/src/compas/numerical/dr/dr_numpy.py
@@ -116,18 +116,19 @@ def dr_numpy(vertices, edges, fixed, loads, qpre,
     # --------------------------------------------------------------------------
     # input processing
     # --------------------------------------------------------------------------
-    if qpre is None:
-        qpre = zeros((num_e,), dtype=float)
-    if fpre is None:
-        fpre = zeros((num_e,), dtype=float)
-    if lpre is None:
-        lpre = zeros((num_e,), dtype=float)
-    if linit is None:
-        linit = zeros((num_e,), dtype=float)
-    if E is None:
-        E = zeros((num_e,), dtype=float)
-    if radius is None:
-        radius = zeros((num_e,), dtype=float)
+
+    def init_array(array, length):
+        if array is None or (isinstance(array, list) and len(array) == 0):
+            return zeros((length,), dtype=float)
+        else:
+            return array
+
+    qpre = init_array(qpre, num_e)
+    fpre = init_array(fpre, num_e)
+    lpre = init_array(lpre, num_e)
+    linit = init_array(linit, num_e)
+    E = init_array(E, num_e)
+    radius = init_array(radius, num_e)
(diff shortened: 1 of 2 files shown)


src/compas/numerical/dr/dr_numpy.py
```diff
@@ -116,12 +116,18 @@ def dr_numpy(vertices, edges, fixed, loads, qpre,
     # --------------------------------------------------------------------------
     # input processing
     # --------------------------------------------------------------------------
-    qpre = qpre or [0.0 for _ in range(num_e)]
-    fpre = fpre or [0.0 for _ in range(num_e)]
-    lpre = lpre or [0.0 for _ in range(num_e)]
-    linit = linit or [0.0 for _ in range(num_e)]
-    E = E or [0.0 for _ in range(num_e)]
-    radius = radius or [0.0 for _ in range(num_e)]
+    if qpre is None:
```
tomvanmele: this will not be triggered on an empty list. so i would do
```python
if qpre is None or not qpre:
```

tomvanmele: and for the others as well of course

src/compas/numerical/dr/dr_numpy.py
```diff
@@ -116,12 +116,19 @@ def dr_numpy(vertices, edges, fixed, loads, qpre,
...
-    E = E or [0.0 for _ in range(num_e)]
-    radius = radius or [0.0 for _ in range(num_e)]
+
+    def init_array(array, length):
+        if array is None or (isinstance(array, list) and len(array) == 0):
+            return zeros((length,), dtype=float)
+        else:
+            return array
+
+    qpre = init_array(qpre, num_e)
```
Licini, line 125: @brgcode `not qpre` behind `or` will still trigger  evaluation on np array inputs thus throws the error. How about this?

tomvanmele: you don't have to check for `list`, `len(array)` works both for `numpy` arrays and python lists

tomvanmele: the `else` is also unnecessary :)

Licini: @brgcode right.... makes sense. updated!

tomvanmele (approved):

PR #871, opened 2021-06-29
Remove default __str__ implementation of data objects

tomvanmele:

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Fixed a bug in `compas.numerical.dr_numpy` when using numpy array as inputs.
 * Allowed for varying repository file structures in `compas.robots.GithubPackageMeshLoader`.
 * Remove default implementation of `__str__` for data objects.
-* Remove implementation of `__str__` for transformation objects.
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,15 +11,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 ### Changed
+
 * `compas.robots.Axis` is now normalized upon initialization.
 * Fixed a bug in `compas.numerical.dr_numpy` when using numpy array as inputs.
 * Allowed for varying repository file structures in `compas.robots.GithubPackageMeshLoader`.
+* Remove default implementation of `__str__` for data objects.
+* Remove implementation of `__str__` for transformation objects.
 
 ### Fixed
+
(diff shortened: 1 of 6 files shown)


beverlylytle:
A consequence of this is that subclasses of `Data` that don't have `__str__` or `__repr__` defined will be printed with Python's somewhat useless and unaesthetic default.  `Mesh` and `Network` are examples of this.  Is that desired?

gonzalocasas (approved):
LGTM!

gonzalocasas:
> A consequence of this is that subclasses of `Data` that don't have `__str__` or `__repr__` defined will be printed with Python's somewhat useless and unaesthetic default. `Mesh` and `Network` are examples of this. Is that desired?

Yeah, that's true but currently all data objects print their data output which is rather annoying as a default, that's why we had to add `repr` in most examples of the workshops we did recently.  And big data structures should anyway default to something like `<Mesh at 0x0000127memorylocaltiona32662>`

tomvanmele:
> A consequence of this is that subclasses of `Data` that don't have `__str__` or `__repr__` defined will be printed with Python's somewhat useless and unaesthetic default.  `Mesh` and `Network` are examples of this.  Is that desired?

no, indeed. but i think we perhaps need to resolve this by adding a meaningful repr to the data structures and even a print function that indicates with its name that it will be printing the data and not a representation of the object...

beverlylytle:
> all data objects print their data output which is rather annoying as a default

Totally agree with this.

>  And big data structures should anyway default to something like <Mesh at 0x0000127memorylocaltiona32662>

They should?  Wouldn't it be nicer if it contained higher level information like the mesh's name and number of vertices?

gonzalocasas:
> They should? Wouldn't it be nicer if it contained higher level information like the mesh's name and number of vertices?

You're right, `should` is not the correct word. They `could` and that would be enough. And agree with you, if they show any additional summary info, that'd be nice. The `RobotModel` does exactly what you say, it prints out a one-liner (eg `Robot name=abb_irb6640, Links=11, Joints=10 (6 configurable)`)

tomvanmele:
> You're right, `should` is not the correct word. They `could` and that would be enough. And agree with you, if they show any additional summary info, that'd be nice. The `RobotModel` does exactly what you say, it prints out a one-liner (eg `Robot name=abb_irb6640, Links=11, Joints=10 (6 configurable)`)

with the mesh there is a "summary" method ```print(mesh.summary())```. it outputs something like this
```
Mesh
==========
- vertices: x
- faces: y
- edges: z
```
we could add an overview of the default vertex/edge/face attributes and the items stored in `attributes`?
and make sure that `__str__` returns this...

gonzalocasas:
the content of `summary()` sound good for the `Mesh.__str__`. I would personally prefer it to be a one-liner in `__str__` (ie. `<Mesh of vertices: x, faces: y, edges: z>`).

beverlylytle (approved):
Looks good!

PR #873, opened 2021-07-02
Default attributes for data structures

tomvanmele:

### Checklist


src/compas/datastructures/network/core/graph.py
```diff
@@ -65,15 +65,22 @@ def DATASCHEMA(self):
...
+                 default_edge_attributes=None):
+        super(Graph, self).__init__(name=name)
         self._max_node = -1
-        self.attributes = {'name': 'Graph'}
+        self.attributes = {}
         self.node = {}
         self.edge = {}
         self.adjacency = {}
-        self.default_node_attributes = {}
+        self.default_node_attributes = {'x': 0.0, 'y': 0.0, 'z': 0.0}
```
beverlylytle, line 78: I'm surprised to see this.  I thought `Graph`s were purely topological.

tomvanmele: they are. before this was added in the network init. however, even graphs have to be visualised so purely from a practical point of view we might as well add the default coordinates here. can change this back if it doesn't feel right ...

tomvanmele: in general, i think the separation between graph and network, and between halfedge and mesh, etc. is a bit artificial and is only there to make the individual class definitions a bit shorter. especially in case of the mesh it doesn't make much sense because halfedge is not meaningful by itself...

beverlylytle: If a graph were visualized with all the nodes at the origin, it's not much of a visualization...

beverlylytle (approved):
Looks good!

src/compas/datastructures/network/core/graph.py
tomvanmele: indeed it is not. the defaults are just there such that more meaningful coordinates can be set through the attribute functions. if no geometry is explicitly assigned, the graph could be considered as purely topological and visualisations are meaningless without a chosen layout. we could wrap the layout mechanisms of `networkx` for this...

gonzalocasas: I would vote for not having coords in the graph and keeping it as purely topological as possible. Then, as mentioned, visualization needs layouting, but we can either use some of the `networkx` ones, or our own, I wrote a few simple ones for our workshops, which surely need improvement but might be a starting point.

gonzalocasas (commented):
Besides the comment about moving coordinates out of `Graph`, it lgtm. If possible, throw in a couple of unit tests to verify basic default attr behavior.

src/compas/datastructures/network/core/graph.py
tomvanmele: sure will move it to network

tomvanmele:
closed in favoour of #892

PR #877, opened 2021-07-08
Support for named objects in OBJ reading/writing

tomvanmele:

### Checklist

### DIFF
--- a/src/compas/files/obj.py
+++ b/src/compas/files/obj.py
@@ -508,7 +508,7 @@ class OBJWriter(object):
             else:
                 self._write_vertices(mesh)
                 self._write_faces(mesh)
-            self._v += mesh.number_of_vertices()
+                self._v += mesh.number_of_vertices()
 
     def _write_vertices(self, mesh):
         for key in mesh.vertices():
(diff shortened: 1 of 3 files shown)


src/compas/files/obj.py
```diff
@@ -371,69 +448,90 @@ def parse(self):
         self.polylines = [[index_index[index] for index in line] for line in self.reader.lines if len(line) > 2]
         self.faces = [[index_index[index] for index in face] for face in self.reader.faces]
         self.groups = self.reader.groups
+        self.objects = {}
```
Licini, line 451: Just curious why is here not using `defaultdict`  like the reader class?

src/compas/files/obj.py
```diff
@@ -371,69 +448,90 @@ def parse(self):
...
+    def _write_faces(self, mesh):
+        key_index = mesh.key_index()
+        for fkey in mesh.faces():
+            vertices = mesh.face_vertices(fkey)
+            vertices = [key_index[key] + self._v for key in vertices]
+            vertices_str = ' '.join([str(index) for index in vertices])
+            self.file.write('f {0}\n'.format(vertices_str))
+
+    def _write_vertices_and_faces(self, mesh):
+        index = 0 + self._v
```
Licini: should be 1 + ?

tomvanmele: no because `self._v` now starts at 1 instead of 0

src/compas/files/obj.py
tomvanmele: there is no need because i am just parsing the corresponding dict of the reader. so there is no possibility of uninitialised items in the dict

Licini: I see!

src/compas/files/obj.py
Licini: cool would better to change to simply `index = self._v`  then?

Licini (approved):

src/compas/files/obj.py
tomvanmele: yes indeed

tomvanmele: hmm actually there might be a small bug related to that

ffurrer:
Great, thanks! This worked perfectly for us :).

PR #881, opened 2021-07-28
New small utility functions

romanarust:

### Checklist


src/compas/utilities/itertools.py
```diff
@@ -185,10 +189,79 @@ def linspace(start, stop, num=50):
...
+    """
+    if len(lst) % num:
+        raise ValueError("The length of the array must be a factor of n: %d %% %d == 0" % (len(lst), num))
+    return [lst[i:i + num] for i in range(0, len(lst), num)]
+
+
+def reshape(lst, shape):
+    """Gives a new shape to an array without changing its data.
+
+    This function mimicks the functionality of ``numpy.reshape`` [1]_, but in a simpler form.
```
beverlylytle:
```suggestion
    This function mimics the functionality of ``numpy.reshape`` [1]_, but in a simpler form.
```

src/compas/utilities/itertools.py
```diff
@@ -185,10 +189,79 @@ def linspace(start, stop, num=50):
...
+    Examples
+    --------
+    >>> a = [[1, 2, 3], [4, 5, 6]]
+    >>> list(flatten(a))
+    [1, 2, 3, 4, 5, 6]
+    """
     return chain.from_iterable(list_of_lists)
 
 
+def unflatten(lst, num):
```
beverlylytle: This function seems to be a limited version of `reshape`, that is, `unflatten(l, num) == reshape(l, (len(l)//num), num))`. I would suggest for the sake of not cluttering the API that this should be removed.

src/compas/utilities/itertools.py
```diff
@@ -185,10 +189,79 @@ def linspace(start, stop, num=50):
 
 
 def flatten(list_of_lists):
-    """Flatten one level of nesting"""
+    """Flatten one level of nesting.
```
beverlylytle: Since numpy's `flatten` doesn't flatten just one level, I think this name might be confusing.  So I would suggest either renaming this function or replacing it with something that behaves as `numpy.flatten` does.  Eg
```
def flatten(l):
    if not hasattr(l[0], '__len__'):
        return l
    return flatten(list(chain(*l)))

File: src/compas/utilities/itertools.py [hunk 4]
```diff
@@ -185,10 +189,79 @@ def linspace(start, stop, num=50):
...
+def reshape(lst, shape):
+    """Gives a new shape to an array without changing its data.
+
+    This function mimicks the functionality of ``numpy.reshape`` [1]_, but in a simpler form.
+
+    Parameters
+    ----------
+    lst : list
+        A list of items.
+    shape : int or tuple of ints
```
beverlylytle, line 216: I don't think it makes sense for `shape` to be an `int`.  As it is now, that would result in a `TypeError`, but I don't think it really makes sense to call `reshape([1,2,3], 3)`.  It would make sense if the input were something more complicated than a flat list.  You could address this by using the version of `flatten` suggested above and then recursively calling a helper function which does what `reshape` currently does, but maybe that's more than you need:
```
def reshape(l, shape):
    def helper(l, shape):
        if len(shape) == 1:
            if len(l) % shape[0] != 0:
                raise ValueError()
            return [l[i:i + shape[0]] for i in range(0, len(l), shape[0])]
        n = reduce(mul, shape[1:])
        return [helper(lst[i * n:(i + 1) * n], shape[1:]) for i in range(len(lst) // n)]

    shape = (shape,) if instance(shape, int) else shape
    l = flatten(l)
    if len(l) != reduce(lambda x, y: x * y, shape):
        raise ValueError("ValueError: cannot reshape array of size %d into shape %s" % (len(lst), shape))
    return helper(l, shape)

**Review by beverlylytle — COMMENTED** [review]
These look like useful functions when `numpy` is not available.  I strongly encourage adding tests though.

File: src/compas/utilities/itertools.py (hunk 2 above)
Comment (romanarust) [reply src/compas/utilities/itertools.py]: 👍

File: src/compas/utilities/itertools.py (hunk 4 above)
Comment (romanarust) [reply src/compas/utilities/itertools.py:216]: thank you for suggesting even better utility functions ;)

**Comment by romanarust** [comment]
ok, i've implemented changes and added tests.

File: CHANGELOG.md [hunk 5]
```diff
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Added `compas.data.is_sequence_of_uint`.
 * Added general plotter for geometry objects and data structures based on the artist registration mechanism.
 * Added support for multimesh files to OBJ reader/writer.
+* Added `reshape` and `unflatten` in `compas.utilities`
```
```suggestion
* Added `reshape` and generalized `flatten` in `compas.utilities`.
```

beverlylytle (approved):
Looks good!

beverlylytle:
@romanarust is there any reason not to merge this?

src/compas/utilities/itertools.py
tomvanmele: i have the feeling this will break a lot of existing code. for example, in case flatten is used on a list of lists of points, the new version will basically return a list of floats, and that is not the case with the previous version.
```python
for point in flatten(list_of_polylines):
    # point is a point in the old version
    # point is a coordinate in the new version
```

src/compas/utilities/itertools.py
```diff
@@ -185,8 +188,67 @@ def linspace(start, stop, num=50):
...
+
+    Examples
+    --------
+    >>> a = [[1, 2, 3], [4, 5, 6]]
+    >>> flatten(a)
+    [1, 2, 3, 4, 5, 6]
+    """
+    if not hasattr(list_of_lists[0], '__len__'):
+        return list_of_lists
+    return flatten(list(chain(*list_of_lists)))
```
tomvanmele: we need to discuss this because i think this is a breaking change...

tomvanmele (changes requested):
just to make sure this is not yet merged by accident

beverlylytle:
We discussed this today in the compas dev meeting.  `flatten` as it currently is written is used quite a lot in compas code and in particular changing it to what I suggested would flatten lists of lists of points to just a list of floats rather than the expected list of points, breaking many things.  It was suggested that this new `flatten` be made an internal function, say `_flatten` or so, for use within `reshape` and that before the release of version 2.0 current usage of `flatten` would be switched to `itertools.chain` and the new `_flatten` would be made external.  Does this make sense?

romanarust:
Ping here again! remove the changes to the "flatten" function, think it is ready now

tomvanmele:
@romanarust black is complaining

tomvanmele (approved):

PR #884, opened 2021-08-11
extend and shorten polyline methods

mattiskoh:
Added two methods for extending and shortening polylines.

### Checklist


src/compas/geometry/primitives/polyline.py
```diff
@@ -394,3 +394,73 @@ def divide_polyline_by_length(self, length, strict=True, tol=1e-06):
...
 
         return division_pts
+
+    def extend(self, length, both_ends=False):
+        """Extends a polyline from its start or end by a given length
+
+        Parameters:
+        -----------
+        length: list of float [start_offset, end_offset]
+
```
tomvanmele, line 421: only the parameters before the blank line will be processed. please remove...

src/compas/geometry/primitives/polyline.py
```diff
@@ -394,3 +394,73 @@ def divide_polyline_by_length(self, length, strict=True, tol=1e-06):
             division_pts.append(self.points[-1])
 
         return division_pts
+
+    def extend(self, length, both_ends=False):
+        """Extends a polyline from its start or end by a given length
```
tomvanmele, line 416: i think this description is a bit misleading. perhaps just "Extends a polyline by a given length."?

src/compas/geometry/primitives/polyline.py
```diff
@@ -394,3 +394,73 @@ def divide_polyline_by_length(self, length, strict=True, tol=1e-06):
...
+        both_ends: bool
+            If True, both ends of Polyline will be shortened
+
+        Returns
+        -------
+        Polyline
+        """
+
+        if not isinstance(length, list):
+            length = [length, length]
```
tomvanmele, line 457: this is a bit restrictive, because it excludes the use of tuples. the following would be more pythonic
```python
try:
    start, end = length
except TypeError:
    start = end = length
```

src/compas/geometry/primitives/polyline.py
```diff
@@ -394,3 +394,73 @@ def divide_polyline_by_length(self, length, strict=True, tol=1e-06):
...
+
+        both_ends: bool
+            If True, both ends of Polyline will be extended
+
+        Returns
+        -------
+        Polyline
+        """
+        if not isinstance(length, list):
+            length = [length, length]
```
tomvanmele, line 430: this is a bit restrictive, because it excludes the use of tuples. the following would be more pythonic
```python
try:
    start, end = length
except TypeError:
    start = end = length
```

mattiskoh:
bump

tomvanmele (commented):
i seem to never have submitted this

src/compas/geometry/primitives/polyline.py
```diff
@@ -412,4 +412,74 @@ def divide_polyline_by_length(self, length, strict=True, tol=1e-06):
 
         return division_pts
 
+    def extend(self, length, both_ends=False):
+        """Extends a polyline from its start or end by a given length
+
+        Parameters:
+        -----------
+        length: list of float [start_offset, end_offset]
```
gonzalocasas, line 420: We now write these a bit different, in this case, I think it would be:
```suggestion
        length: list[float, float]
```
@tomvanmele is that correct?

src/compas/geometry/primitives/polyline.py
```diff
@@ -412,4 +412,74 @@ def divide_polyline_by_length(self, length, strict=True, tol=1e-06):
...
+
+        both_ends: bool
+            If True, both ends of Polyline will be extended
+
+        Returns
+        -------
+        Polyline
```
gonzalocasas, line 427:
```suggestion
        :class:`~compas.geometry.Polyline`
```

src/compas/geometry/primitives/polyline.py
```diff
@@ -412,4 +412,74 @@ def divide_polyline_by_length(self, length, strict=True, tol=1e-06):
...
+        """
+        if not isinstance(length, list):
+            length = [length, length]
```
gonzalocasas, line 430: Not sure if this is such a great idea, I would stick to more strict input: it's either a list or we let it fail.

src/compas/geometry/primitives/polyline.py
```diff
@@ -412,4 +412,74 @@ def divide_polyline_by_length(self, length, strict=True, tol=1e-06):
...
+
```
gonzalocasas, line 431: I would add here argument validation, because you check for specific indices later on, so perhaps, something like:
```suggestion
        length = list(length)

        if len(length) < 2:
            raise ValueError('length')
```

src/compas/geometry/primitives/polyline.py
```diff
@@ -412,4 +412,74 @@ def divide_polyline_by_length(self, length, strict=True, tol=1e-06):
...
+            self.points.insert(0, new_start_pt)
+
+        return self
+
+    def shorten(self, length, both_ends=False):
+        """Shorten a polyline from start or end by a given length
+
+        Parameters:
+        -----------
+        length: float or list of float [start_offset, end_offset]
```
gonzalocasas, line 446: same as above, I would rather be strict and stick to a single input type, if not, the syntax of `or` is with `|` so: `list[float, float] | float`

gonzalocasas (commented):
Added a few couple more small comments but it's looking good

src/compas/geometry/primitives/polyline.py
tomvanmele: not really.
you can't specify a fixed length sequence as a list.
you could do `tuple[float, float]`, but we have been using `[float, float]` in case any sequence of 2 floats is acceptable...

tomvanmele:
i would still like to see this behaviour somewhat clarified. for example, without looking at the code, i would expect the last segment of the polyline to become longer by a given amount. however, this doesn't seem to be what is happening. if i understand correctly, an additional point is added. the resulting polyline thus has more points and segments than the original one. furthermore, it is also not clear when "both_ends" is specified, but only 1 length, whether this length will be added in total, or at each end...

PR #885, opened 2021-08-11
added Plane offset method

mattiskoh:
Added Plane.offset(), offset a plane in the normal direction 

### Checklist

### DIFF
--- a/src/compas/geometry/primitives/plane.py
+++ b/src/compas/geometry/primitives/plane.py
@@ -298,3 +298,16 @@ class Plane(Primitive):
         """
         self.point.transform(T)
         self.normal.transform(T)
+
+    def offset(self, distance):
+        """Returns a new offset plane by a given distance.
+        Plane normal is used as positive direction.
+
+        Parameters:
+        -----------
+        plane: compas.geometry.Plane
+
+        distance: float
+        """
+
+        return Plane(self.point+self.normal.scaled(distance), self.normal)
--- a/src/compas/geometry/primitives/plane.py
+++ b/src/compas/geometry/primitives/plane.py
@@ -309,6 +309,11 @@ class Plane(Primitive):
             The plane to be offset.
         distance: :obj:`float`
             The offset distance.
+
+        Returns
+        --------
+        :class:`Plane`
+            The offset plane.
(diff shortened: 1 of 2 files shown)


src/compas/geometry/primitives/plane.py
```diff
@@ -298,3 +298,16 @@ def transform(self, T):
...
         self.point.transform(T)
         self.normal.transform(T)
+
+    def offset(self, distance):
+        """Returns a new offset plane by a given distance.
+        Plane normal is used as positive direction.
+
+        Parameters:
+        -----------
+        plane: compas.geometry.Plane
```
tomvanmele: to properly link to the class definitions in the docs, this should be 
```
plane: :class:`Plane`
```

src/compas/geometry/primitives/plane.py
```diff
@@ -298,3 +298,16 @@ def transform(self, T):
...
+
```
tomvanmele, line 312: there can't be blank lines in the parameter list. the parameters after the blank line will not be recognized...

src/compas/geometry/primitives/plane.py
```diff
@@ -298,3 +298,16 @@ def transform(self, T):
...
+        distance: float
```
tomvanmele: ideally this would be 
```
distance: :obj:`float`
```

tomvanmele (changes requested):

src/compas/geometry/primitives/plane.py
```diff
@@ -298,3 +298,17 @@ def transform(self, T):
...
+        """Returns a new offset plane by a given distance.
+        Plane normal is used as positive direction.
+
+        Parameters:
+        -----------
+        plane: :class:`compas.geometry.Plane`
+            The plane to be offset.
+        distance: :obj:`float`
+            The offset distance.
+        """
```
tomvanmele, line 317: would be good to add a return value here:
```python
Returns
--------
:class:`Plane`
    The offset plane.

tomvanmele (commented):
sorry last request

tomvanmele (approved):

PR #887, opened 2021-08-25
Basic Assembly data structure

tomvanmele:

### Checklist


src/compas/datastructures/assembly/assembly.py
```diff
@@ -0,0 +1,44 @@
...
+    def JSONSCHEMANAME(self):
+        return 'assembly'
+
+    def __init__(self, *args, **kwargs):
+        super(Assembly, None).__init__(*args, **kwargs)
+        self.attributes = {}
+        self.network = Network()
+
+    def __str__(self):
+        tpl = "<Assembly with {} elements and {} interactions>"
```
gonzalocasas: Naming -in this line- is perhaps the most defining contribution of this PR right now.

We've internally discussed assembly-related naming at great lengths:
- `Element`: we've recently settled on calling it `Part` as it matches terminology in many fields that already deal with assembly and assemblability.
- `Interactions`: this seems to be skewed towards equilibrium analysis. A more neutral terminology would be `Connections`, which also seems to be used in some of literature. (See in particular, [State of the Art on Computational Design of Assemblies with Rigid Parts](https://kiki007.github.io/publication/assemblysurvey/paper.pdf) by Ziqi et al, and to some degree also the part-graph formalization in [Computational Interlocking Furniture Assembly](https://dl.acm.org/doi/pdf/10.1145/2766892) which seems to use `connections` as the abstract term for what eventually turns into a concrete `joint`.

As a reference of the direction we're taking for assemblies in robotic (and XR) fabrication, here's a high level block diagram, the white boxes are most likely going to be classes:
image

Another aspect of assemblies that has been discussed as a very important need is the ability to support `sub-assemblies`, which does not need to be included in this PR for sure, but it's relevant to keep in mind.

tomvanmele: part + connections works for me. sub-assemblies should not be a problem i think since the data structures can be nested to whatever level needed...

will add a few more things in the next few days, but want to keep this base class simple and then extend with more specialised versions later on.

gonzalocasas: 👍

src/compas/datastructures/assembly/assembly.py
```diff
@@ -0,0 +1,69 @@
...
+        """The parts of the assembly."""
+        return self.network.nodes()
+
+    def connections(self):
+        """The connections between the parts."""
+        return self.network.edges()
+
+    def add_part(self, part, id=None, **kwargs):
+        """Add a part to the assembly."""
+        self.network.add_node(key=id, part=part, **kwargs)
```
beverlylytle: I don't follow... is the `part` a network node (as is returned by `self.parts`)? or something else?  And if `id` isn't specified, shouldn't this return the identifier that is generated?

gonzalocasas: yep, `Part` is the new name for what we used to call `Element` in other assembly structures, so they are stored as network nodes.

And you're right about the need to return the generated identifier.

gonzalocasas: Ah, maybe I misunderstood your question. Are you asking why `.parts()` will return identifiers instead of the `part` attribute stored in the node?

tomvanmele: it all doesn't really make sense to me yet either :) will have to try this on a few concrete examples first to get it right...

beverlylytle: Exactly.

tomvanmele: can a part be anything, or is it also an assembly (or a "part" which is a subclass of "assembly", since perhaps there should be only 1 "assembly" in an assembly) with at least one node?

gonzalocasas: I would generally argue that the `Assembly` stuff should lean towards being normative/opinionated, unlike `Network` or `Mesh` that are more general.
So, a `Part` should not be anything, there should be a `Part` base class.
And I have the feeling that it is `Assembly` that needs to sub-class `Part` and not the other way around.

Also, we def need a way to lookup parts quickly, so, a `.part(id)` method or even an dict-like accessor to get to parts quick would be important.

docs/tutorial/assembly.rst
```diff
@@ -0,0 +1,37 @@
+**********
+Assemblies
+**********
+
+.. rst-class:: lead
+
+The :class:`Assembly` data structure provides functionality for modelling and managing
```
beverlylytle, line 7: 🇺🇸 
```suggestion
The :class:`Assembly` data structure provides functionality for modeling and managing
```

src/compas/datastructures/assembly/assembly.py
```diff
@@ -2,68 +2,145 @@
...
+        ----------
+        part: :class:`compas.datastructures.Node`
+            The part or assembly of parts to add.
+        key: int or str, optional
+            The identifier of the part in the assembly.
+            Note that the key is unique only in the context of the current assembly.
+            Nested assemblies may have the same ``key`` value for one of their parts.
+            Default is ``None`` in which case the key will be automatically assigned integer value.
+        kwargs: dict
+            Additional named parameters colledted in a dict.
```
beverlylytle:
```suggestion
            Additional named parameters collected in a dict.
```

src/compas/datastructures/assembly/assembly.py
```diff
@@ -2,68 +2,145 @@
...
+
+        """
+        for node in self.graph.nodes():
+            part = self.graph.node_attribute(node, 'part')
+            if part.guid == guid:
+                return part
+        # this is effectively a depth-first search through all the parts of the assembly
+        for node in self.graph.nodes():
+            part = self.graph.node_attribute(node, 'part')
+            part = part.find(guid)
```
beverlylytle: This could result in an infinite loop, but I don't know if we want to check for that or expect the user to not do that.

beverlylytle: Also, `Part`/`Node`s don't have a `find` method, so I think this would raise some exceptions.

src/compas/datastructures/assembly/assembly.py
```diff
@@ -0,0 +1,146 @@
...
+            A globally unique identifier.
+            This identifier is automatically assigned when parts are created.
+
+        Returns
+        -------
+        :class:`compas.datastructures.Node` or None
+            The identified part, if any.
+
+        """
+        for node in self.graph.nodes():
```
gonzalocasas: `find` by guid should be a O(1) operation, ie. we should maintain a map of parts keyed by guid and just do a lookup. Sub-assemblies will need some extra love like looping over them (which triggers the question of how to get them without having to loop through all parts), and calling the O(1) `.find()`  on them.

Another consideration here is naming. Other libraries and languages use `find` for queries that return 0..n elements, and `get` for lookups that must return 0..1 only. Unless this would create inconsistencies with other parts of compas, I would vote for going a similar route.

src/compas/datastructures/assembly/part.py
```diff
@@ -0,0 +1,168 @@
...
+            "key": int,
+            "frame": Frame,
+            "transformations": list,
+            "geometry": Shape,
+            "features": list
+        })
+
+    @property
+    def JSONSCHEMANAME(self):
+        return 'assembly'
```
gonzalocasas: I guess this one should be 

```suggestion
        return 'part'
```

src/compas/datastructures/assembly/part.py
```diff
@@ -0,0 +1,168 @@
...
+            'mesh': self.mesh.data,
+            'graph': self.graph.data
+        }
+        return data
+
+    @data.setter
+    def data(self, data):
+        self.attributes.update(data['attributes'] or {})
+        self.mesh.data = data['mesh']
+        self.graph.data = data['graph']
```
gonzalocasas: these two methods look like left overs of a previous implementation, right? (mesh and graph are no longer here)

src/compas/datastructures/assembly/assembly.py
```diff
@@ -0,0 +1,142 @@
...
+        tuple of str or int
+            The tuple of node identifiers that identifies the connection.
+
+        Raises
+        ------
+        :class:`AssemblyError`
+            If ``a`` and/or ``b`` are not in the assembly.
+        """
+        if a.key is None or b.key is None:
+            raise AssemblyError('Both parts have to be added to the asembly before a connection can be created.')
```
gonzalocasas:
```suggestion
            raise AssemblyError('Both parts have to be added to the assembly before a connection can be created.')
```

src/compas/datastructures/assembly/assembly.py
```diff
@@ -0,0 +1,142 @@
...
+        if not self.graph.has_node(a.key) or not self.graph.has_node(b.key):
+            raise AssemblyError('Both parts have to be added to the asembly before a connection can be created.')
```
gonzalocasas:
```suggestion
            raise AssemblyError('Both parts have to be added to the assembly before a connection can be created.')
```

src/compas/datastructures/assembly/assembly.py
```diff
@@ -0,0 +1,142 @@
...
+
+        Returns
+        -------
+        int or str
+            The identifier of the part in the current assembly graph.
+
+        """
+        key = self.graph.add_node(key=key, part=part, **kwargs)
+        part._key = key
+        self.parts[part.guid] = part
```
gonzalocasas: here I feel we need to add a second dict for parts by key as well.. and then a `find/get_by_key` method to go with it

src/compas/datastructures/assembly/part.py
```diff
@@ -0,0 +1,168 @@
...
+        """Add a feature to the geometry and the operation through which they should be combined.
+
+        Parameters
+        ----------
+        shape : :class:`compas.geometry.Shape`
+            The shape of the feature.
+        operation : {'union', 'difference', 'intersection'}
+            The boolean operation through which the feature should be integrated in the geometry.
+        """
+        if operation not in ('union', 'difference', 'intersection'):
```
gonzalocasas: Maybe change to `if operation not in Part.operations.keys()`?

src/compas/datastructures/assembly/part.py
```diff
@@ -0,0 +1,168 @@
...
+        The shape will be available in ``Part.shape``.
+        """
+        geometry = self.geometry
+        for shape, operation in self.features:
+            geometry = Part.operations[operation](geometry, shape)
+        self._shape = geometry
+
+    def apply_transformations(self):
+        """Apply all transformations to the part shape."""
+        T = reduce(multiply_matrices, self._transformations)
```
gonzalocasas: not sure if correct, but, shouldn't we start with the `Transformation.from_frame(self.frame)` as the first transformation in the queue so that non-worldXY frames are taken into account?

src/compas/datastructures/assembly/part.py
```diff
@@ -0,0 +1,168 @@
...
+            raise FeatureError
+        self._features.append(shape, operation)
+
+    def apply_features(self):
+        """Apply all features to the base geometry to construct the final shape of the part.
+        The shape will be available in ``Part.shape``.
+        """
+        geometry = self.geometry
+        for shape, operation in self.features:
+            geometry = Part.operations[operation](geometry, shape)
```
gonzalocasas: isn't this missing a `geometry.to_vertices_and_faces(), shape.to_vertices_and_faces()`?

src/compas/datastructures/assembly/part.py
```diff
@@ -0,0 +1,168 @@
...
+        return 'assembly'
+
+    def __init__(self, name, **kwargs):
+        super(Part, self).__init__()
+        self._key = None
+        self._frame = None
+        self._transformations = deque()
+        self._geometry = None
+        self._features = []
+        self._shape = None
```
gonzalocasas: The terms `geometry` and `shape` are confusingly close, but I can't think of a good alternative right now. `base_geometry` and `final_geometry` come to mind, but they are not really correct. Also, `base_geometry` might be the only geometry in a lot of cases, so it's a bit annoying to have to prefix it all the time

src/compas/datastructures/assembly/part.py
tomvanmele: indeed ... updated!

src/compas/datastructures/assembly/assembly.py
tomvanmele: will do

src/compas/datastructures/assembly/part.py
tomvanmele: done

src/compas/datastructures/assembly/part.py
tomvanmele: indeed

src/compas/datastructures/assembly/part.py
tomvanmele: added a bit of a hack to solve this

src/compas/datastructures/assembly/part.py
tomvanmele: is somewhat in line with the terminology in OCC

src/compas/datastructures/assembly/assembly.py
```diff
@@ -0,0 +1,142 @@
...
+            "attributes": dict,
+            "graph": Graph,
+        })
+
+    @property
+    def JSONSCHEMANAME(self):
+        return 'assembly'
+
+    def __init__(self, name=None, **kwargs):
+        super(Assembly, None).__init__()
```
gonzalocasas:
```suggestion
        super(Assembly, self).__init__()
```

src/compas/datastructures/assembly/assembly.py
```diff
@@ -0,0 +1,142 @@
...
+        return schema.Schema({
+            "attributes": dict,
+            "graph": Graph,
+        })
+
+    @property
+    def JSONSCHEMANAME(self):
+        return 'assembly'
+
+    def __init__(self, name=None, **kwargs):
```
gonzalocasas, line 33: Isn't this missing assigning `name` to something?

src/compas/datastructures/assembly/part.py
```diff
@@ -0,0 +1,187 @@
...
+            "shape": Shape,
+            "features": list,
+            "transformations": list,
+        })
+
+    @property
+    def JSONSCHEMANAME(self):
+        return 'part'
+
+    def __init__(self, name, frame=None, shape=None, features=None, **kwargs):
```
gonzalocasas: `name` should be optional (the code in line 80 implies it should be at least)

src/compas/datastructures/assembly/assembly.py
```diff
@@ -0,0 +1,142 @@
...
+        :class:`AssemblyError`
+            If ``a`` and/or ``b`` are not in the assembly.
+        """
+        if a.key is None or b.key is None:
+            raise AssemblyError('Both parts have to be added to the assembly before a connection can be created.')
+        if not self.graph.has_node(a.key) or not self.graph.has_node(b.key):
+            raise AssemblyError('Both parts have to be added to the assembly before a connection can be created.')
+        return self.graph.add_edge(a.key, b.key, **kwargs)
+
+    def parts(self):
```
gonzalocasas, line 124: this method is not callable because the `self.parts` attribute is shadowing it

src/compas/datastructures/assembly/part.py
```diff
@@ -0,0 +1,187 @@
...
+    @property
+    def geometry(self):
+        # this is a temp solution
+        A = Mesh.from_shape(self.shape)
+        for shape, operation in self.features:
+            A.quads_to_triangles()
+            B = Mesh.from_shape(shape)
+            B.quads_to_triangles()
+            A = Part.operations[operation](A.to_vertices_and_faces(), B.to_vertices_and_faces())
+        geometry = Shape(*A)
```
gonzalocasas, line 145: this fails if there are no features assigned to the part

src/compas/datastructures/assembly/part.py
```diff
@@ -0,0 +1,187 @@
...
+        if not self._frame:
+            self._frame = Frame.worldXY()
+        return self._frame
+
+    @frame.setter
+    def frame(self, frame):
+        self._frame = frame
+
+    @property
+    def geometry(self):
```
gonzalocasas, line 136: this method would definitely need memoization of some kind otherwise, it will be recalculated every time the property is accessed even when no changes were made to the shape or features.

src/compas/datastructures/assembly/part.py
```diff
@@ -0,0 +1,187 @@
...
+        })
+
+    @property
+    def JSONSCHEMANAME(self):
+        return 'part'
+
+    def __init__(self, name, frame=None, shape=None, features=None, **kwargs):
+        super(Part, self).__init__()
+        self._frame = None
+        self.attributes = {'name': name or 'Part'}
```
gonzalocasas, line 80: should we update `self.attributes` with `kwargs`? (same for `Assembly`) or are assembly/part attributes not supposed to be assigned like that?

src/compas/datastructures/assembly/part.py
```diff
@@ -0,0 +1,187 @@
...
+    def geometry(self):
+        # this is a temp solution
+        A = Mesh.from_shape(self.shape)
+        for shape, operation in self.features:
+            A.quads_to_triangles()
+            B = Mesh.from_shape(shape)
+            B.quads_to_triangles()
+            A = Part.operations[operation](A.to_vertices_and_faces(), B.to_vertices_and_faces())
+        geometry = Shape(*A)
+        T = Transformation.from_frame_to_frame(Frame.worldXY(), self.frame)
```
gonzalocasas, line 146: Not sure about what's expected for the queue of transformations. Should they be applied automatically when retrieving the geometry?

src/compas/datastructures/assembly/part.py
```diff
@@ -0,0 +1,187 @@
...
+            "transformations": [T.data for T in self.transformations],
+        }
+        return data
+
+    @data.setter
+    def data(self, data):
+        self.attributes.update(data['attributes'] or {})
+        self.key = data['key']
+        self.frame.data = data['frame']
+        self.shape.data = data['shape']
```
gonzalocasas, line 121: this doesn't deserialize properly in my tests... If I change it to rely on the recursive serialization thou, it works:

```suggestion
            "frame": self.frame,
            "shape": self.shape,
            "features": [(shape.data, operation) for shape, operation in self.features],
            "transformations": [T.data for T in self.transformations],
        }
        return data

    @data.setter
    def data(self, data):
        self.attributes.update(data['attributes'] or {})
        self.key = data['key']
        self.frame = data['frame']
        self.shape= data['shape']
```

src/compas/datastructures/assembly/assembly.py
```diff
@@ -0,0 +1,142 @@
...
+        """The parts of the assembly.
+
+        Yields
+        ------
+        :class:`compas.datastructures.Part`
+            The individual parts of the assembly.
+        """
+        for node in self.graph.nodes():
```
gonzalocasas, line 132: this iteration is not stable across the cpython-ironpython divide. The iter order of cpython is not a match to the one in ironpython. Once option would be to explicitly do `sorted(self.graph.nodes())` to make sure they match.

src/compas/datastructures/assembly/part.py
tomvanmele: the explicit data conversion is necessary through. so something else must be wrong. i assume you used the json serializers of the data framework. this will indeed work, but the data dict should be a pure Python representation of the object even without serialisation... without the use of `.data` it is basically a collection of COMPAS objects.

src/compas/datastructures/assembly/assembly.py
tomvanmele: we could use an ordered dict explicitly to be compatible with ipy, but we would have to make the change on the graph directly ...

src/compas/datastructures/assembly/assembly.py
tomvanmele: if we use the guid to make a map, all parts should be findable from the top level since the guid is "guaranteed" to be unique. would not be too complicated to manage if we only allow parts to be added using a method of the assembly and not the (sub)part.

src/compas/datastructures/assembly/part.py
tomvanmele: the idea is that "shapes" are well-defined regular objects (like the shapes in `compas.geometry`). "geometry" is more general and can basically be anything, but cannot be expressed directly as a (named) shape. so `(base) shape + features => (general) geometry`

tomvanmele:
will go through the entire thing later today and give it a bit of a psuh...

src/compas/datastructures/assembly/assembly.py
gonzalocasas: Changing graph sounds to radical if only for this reason...

In the "fab domain", the concept of an assembly sequencer is important... perhaps one option is to implement the interface of such a class at the compas core level, add a basic implementation that does linear sequencing based on key sorting, and then this method can use that.

gonzalocasas:
@brgcode I will propose a few updates to this PR in another PR based on this one, because -if possible- I'd like to get this merged this coming week to use it at a workshop on the 13th.

tomvanmele:
we will start our internal hackathon today so many more small updates to come :)

gonzalocasas (approved):
LGTM! Let's go for it 🚀

gonzalocasas:
Yay!

PR #888, opened 2021-08-27
Rhino interop through conversion and coersion

tomvanmele:

### Checklist


src/compas/geometry/curves/nurbs.py
```diff
@@ -0,0 +1,471 @@
+from math import sqrt
+
+from typing import Dict, List
```
gonzalocasas: this fails in ironpython

src/compas/geometry/curves/nurbs.py
```diff
@@ -0,0 +1,471 @@
...
+            'multiplicities': is_sequence_of_int,
+            'degree': int,
+            'is_periodic': bool
+        })
+
+    @property
+    def JSONSCHEMANAME(self):
+        raise NotImplementedError
+
+    def __init__(self, name=None) -> None:
```
gonzalocasas: same here, typing hints wont work on ipy

src/compas/geometry/surfaces/__init__.py
```diff
@@ -0,0 +1,8 @@
+from ._surface import Surface  # noqa F401
+
+try:
+    import compas_occ  # noqa F401
```
gonzalocasas: OCC is LGPL, right? I am not entirely sure this is still ok to reference explicitly without violating licenses. I guess it is, because there's a fallback to non-occ implementation... 🤷‍♀️

src/compas/geometry/surfaces/nurbs.py
```diff
@@ -0,0 +1,398 @@
...
+    # ==============================================================================
+
+    def to_step(self, filepath: str, schema: str = "AP203") -> None:
+        """Write the surface geometry to a STP file."""
+        raise NotImplementedError
+
+    def to_mesh(self, nu: int = 100, nv: Optional[int] = None) -> 'compas.datastructures.Mesh':
+        """Convert the surface to a quad mesh."""
+        from itertools import product
+        from functools import lru_cache
```
gonzalocasas: why are these imports inlined in the function?

src/compas_rhino/interop/__init__.py
```diff
@@ -0,0 +1,175 @@
+"""
+********************************************************************************
+interop
```
gonzalocasas: `interop` doesn't sound right, the whole of compas is interoperable with rhino, this package is more specific, is really about conversions, I would be explicit in its name as well, e.g. `compas_rhino.conversions` or `compas_rhino.convert` .

src/compas_rhino/interop/exceptions.py
```diff
@@ -0,0 +1,2 @@
+class COMPASRhinoInteropError(Exception):
```
gonzalocasas: How about `ConversionError`? the class is already in the `compas_rhino` package, it's redundant to add that to the name as well.

src/compas_rhino/interop/__init__.py
```diff
@@ -0,0 +1,175 @@
...
+
+
+Primitives
+==========
+
+.. autosummary::
+    :toctree: generated/
+    :nosignatures:
+
+    compas_point_to_rhino_point
```
gonzalocasas: the return of the non object-oriented API? ;)

I think this is confusing in a few ways:
 - for devs, the link between what objects exists and which one have implemented conversions is hard to follow, might lead to incomplete API
 - the names are too verbose, except a few exceptions (plane-to-frame, and some curves to different types of curves), most of these are from compas_X to rhino_X. I don't think the exceptions really justify the verbosity of all methods
 - there should be one way of doing things, and now there's at least 2 ways of each conversion. And the more worring part is that they don't share the code: many objects have `to_compas` which does the `rhino_bla_to_compas_bla` side of the conversion, and some conversions to rhino geometry exist, for instance, in `compas_ghpython`, doing the other side. If there are multiple ways of doing the same, they should definitely invoke the same code.

gonzalocasas (changes requested):
I added a bunch of comments. Generally, I think this is heading in the right direction, but it still needs a few adjustments before merge. The API for conversions is particularly something I think deserves more attention.

Another one is that there's a large amount of `NotImplemented` methods. I'm not sure if the plan is to add their implementation before merge or not, but they are a bit too much to add to a release, so, I would prefer to merge this only once at least a majority of them contain implementations.

src/compas/geometry/surfaces/__init__.py
tomvanmele: https://opensource.stackexchange.com/questions/5175/including-untouched-lgpl-library-in-a-mit-licenced-project

src/compas_rhino/interop/__init__.py
tomvanmele: okay, i will move this to the geometry objects. might create a breaking change here and there though...

tomvanmele:
> Another one is that there's a large amount of `NotImplemented` methods. I'm not sure if the plan is to add their implementation before merge or not, but they are a bit too much to add to a release, so, I would prefer to merge this only once at least a majority of them contain implementations.

this relates a bit to where i got stuck with this PR. i don't think a pure Python version of the NURBS functionality is particularly useful: inside Rhino we use Rhino, and outside there is no reason to not use OCC.

however, there should be sufficient functionality provided by the version included in `compas.geometry` to facilitate data exchange between Rhino (from inside Rhino) to CPython (outside Rhino). so perhaps it would be better to just provide a base class with all the data methods in `compas.geometry` and then add further functionality in `compas_rhino.geometry` and `compas_occ.geometry`?

src/compas/geometry/curves/nurbs.pyi
```diff
@@ -0,0 +1,164 @@
...
+
+    def locus(self, resolution: int = 100) -> List[Point]: ...
+
+    def point_at(self, u: float) -> Point: ...
+
+    def tangent_at(self, u: float) -> Vector: ...
+
+    def curvature_at(self, u: float) -> Vector: ...
+
+    def frame_at(self, u: float) -> Vector: ...
```
beverlylytle:
```suggestion
    def frame_at(self, u: float) -> Frame: ...
```

src/compas/geometry/surfaces/nurbs.pyi
```diff
@@ -0,0 +1,157 @@
...
+
+    # ==============================================================================
+    # Conversions
+    # ==============================================================================
+
+    def to_step(self, filepath: str, schema: str = "AP203") -> None: ...
+
+    def to_mesh(self, nu: int = 100, nv: Optional[int] = None) -> 'compas.datastructures.Mesh': ...
+
+    def to_triangles(self, nu: int = 100, nv: Optional[int] = None) -> List[Tuple[float, float, float]]: ...
```
beverlylytle:
```suggestion
    def to_triangles(self, nu: int = 100, nv: Optional[int] = None) -> List[Point]: ...
```

src/compas_rhino/interop/primitives.py
```diff
@@ -0,0 +1,277 @@
...
+def rhino_polyline_to_compas_polyline(polyline):
+    """Convert a Rhino polyline to a COMPAS polyline.
+
+    Parameters
+    ----------
+    polyline : :class:`Rhino.Geometry.Polyline`
+
+    Returns
+    -------
+    :class:`compas.geometry.Ellipse`
```
beverlylytle:
```suggestion
    :class:`compas.geometry.Polyline`
```

src/compas_rhino/interop/primitives.py
```diff
@@ -0,0 +1,277 @@
...
+    """
+    return Polyline([rhino_point_to_compas_point(point) for point in polyline])
+
+
+def compas_polyline_to_rhino_polyline(polyline):
+    """Convert a COMPAS polyline to a Rhino polyline.
+
+    Parameters
+    ----------
+    polyline : :class:`compas.geometry.Ellipse`
```
beverlylytle:
```suggestion
    polyline : :class:`compas.geometry.Polyline`
```

src/compas_rhino/interop/primitives.py
```diff
@@ -0,0 +1,277 @@
...
+def rhino_polygon_to_compas_polygon(polygon):
+    """Convert a Rhino polygon to a COMPAS polygon.
+
+    Parameters
+    ----------
+    polygon : :class:`Rhino.Geometry.Polygon`
+
+    Returns
+    -------
+    :class:`compas.geometry.Ellipse`
```
beverlylytle:
```suggestion
    :class:`compas.geometry.Polygon`
```

src/compas_rhino/interop/primitives.py
```diff
@@ -0,0 +1,277 @@
...
+    """
+    return Polygon([rhino_point_to_compas_point(point) for point in polygon])
+
+
+def compas_polygon_to_rhino_polygon(polygon):
+    """Convert a COMPAS polygon to a Rhino polygon.
+
+    Parameters
+    ----------
+    polygon : :class:`compas.geometry.Ellipse`
```
beverlylytle:
```suggestion
    polygon : :class:`compas.geometry.Polygon`
```

src/compas_rhino/interop/primitives.py
```diff
@@ -0,0 +1,277 @@
...
+
+    Returns
+    -------
+    :class:`Rhino.Geometry.Ellipse`
```
beverlylytle:
```suggestion
    :class:`Rhino.Geometry.Polygon`
```

beverlylytle (commented):
just some doc suggestions

src/compas_rhino/interop/__init__.py
tomvanmele: before i remove the function-based conversion though, how do we feel about the following?

1. the RhinoGeometry wrappers as they currently exist are a bit of a mixed bag: they are not strictly geometry wrappers in the sense that they can be instantiated from doc objects, selections (also doc objects), Rhino geometry, and COMPAS geometry
2. although they can be used to convert COMPAS geometry to Rhino Geometry, the process is not immediately obvious (`rhino_circle = RhinoCircle.from_geometry(compas_circle).geometry`)
3. although there is a `from_object` method, there is no `to_object` method (use cases would be very confusing since one would have to add the geometry object to the doc objects table, even though the object could have been created from a doc object in the first place)
4. to be compatible with the data framework, we would have to derive them from `compas.data.Data`, and make the data dicts compatible with the data dicts of the corresponding COMPAS objects

there is however the advantage of being able to add additional methods such as `closest_point` which allows the Rhino objects to be used as constraints in optimisation processes, without having to check for the type of object at every iteration.

there is also the possibility to add some of the Rhino user data to the data dict representation such that this can be passed to CPython via RPC. however, there is currently no corresponding data attribute on the geometry objects on the COMPAS side

tomvanmele: @gonzalocasas @beverlylytle

src/compas/geometry/curves/bezier.py
```diff
@@ -0,0 +1,192 @@
+from compas.geometry import Point
+from compas.geometry import Vector
+from ._curve import Curve
+
+from math import factorial
+
+
+def binomial_coefficient(n, k):
```
gonzalocasas: It looks like this method and the next one (`bernstein`) are not really used nor exposed as part of the API, right? Should we add them to the API?

src/compas_rhino/conversions/__init__.py
```diff
@@ -0,0 +1,182 @@
+"""
+********************************************************************************
+interop
```
gonzalocasas: `conversions` ?

src/compas_rhino/conversions/__init__.py
```diff
@@ -0,0 +1,182 @@
...
+********************************************************************************
+
+.. currentmodule:: compas_rhino.interop
```
gonzalocasas:
```suggestion
.. currentmodule:: compas_rhino.conversions
```

docs/api/compas_rhino.interop.rst
```diff
@@ -0,0 +1,2 @@
+
```
gonzalocasas: `conversions` in the automodule and also filename, right?

src/compas_rhino/__init__.py
```diff
@@ -12,6 +12,7 @@
     compas_rhino.conduits
     compas_rhino.forms
     compas_rhino.geometry
+    compas_rhino.interop
```
gonzalocasas:
```suggestion
    compas_rhino.conversions
    compas_rhino.forms
    compas_rhino.geometry
```

src/compas_rhino/conversions/curves.py
```diff
@@ -0,0 +1,200 @@
...
+
+from .primitives import line_to_rhino
+from .primitives import circle_to_rhino
+from .primitives import ellipse_to_rhino
+from .primitives import point_to_compas
+from .primitives import circle_to_compas
+from .primitives import ellipse_to_compas
+from .primitives import polyline_to_compas
+
+from Rhino.Geometry import NurbsCurve as RhinoNurbsCurve
```
gonzalocasas, line 17: i know this is not something you do very often but in this case, wouldn't it be clearer to do `import Rhino.Geometry as rg` and then refer to `rg.NurbsCurve` in code?

src/compas_rhino/conversions/exceptions.py
```diff
@@ -0,0 +1,2 @@
+class ConversionError(Exception):
+    pass
```
gonzalocasas, line 7: github doesn't let me comment on the empty file right below, but +1 for mesh conversions! ;)

src/compas_rhino/geometry/box.py
```diff
@@ -0,0 +1,103 @@
+from __future__ import print_function
+from __future__ import absolute_import
+from __future__ import division
+
+import Rhino
+import compas_rhino
+from compas.geometry import Box
+from ..conversions import box_to_rhino
+from ..conversions import box_to_compas
+from ..conversions import ConversionError
```
gonzalocasas: I'd be tempted to suggest explicit imports here instead of one-level-up relative import.

src/compas_rhino/geometry/mesh.py
```diff
@@ -3,54 +3,37 @@
...
-        The colors of the vertices.
-        Setting this to ``None`` unsets the vertex colors.
-    border (read-only) : list
-        The GUIDs of the border curves.
     """
 
-    def __init__(self):
-        super(RhinoMesh, self).__init__()
-
-    @property
```
gonzalocasas, line 35: isn't all these commenting out quite a big breaking change?

src/compas_rhino/geometry/plane.py
```diff
@@ -92,9 +50,19 @@ def from_selection(cls):
...
         """Convert to a COMPAS geometry object.
 
+        Returns
+        -------
+        :class:`compas.geometry.Plane`
+            A COMPAS plane.
+        """
+        return plane_to_compas(self.geometry)
+
+    def to_compas_frame(self):
```
gonzalocasas, line 55: Should these be `staticmethod`s instead? Because they don't really construct the object in which they are located, but they are very useful and clearly related.

A related comment: in `compas_fab` we added methods like `coerce_frame` , which take as argument either a compas `Frame` or a rhino `Plane` and return a compas `Frame`. Do we have a similar thing now in this PR that can return the compas object of X regardless of whether the input is already of that type?

src/compas_rhino/geometry/sphere.py
```diff
@@ -0,0 +1,113 @@
...
+                    raise ConversionError('Object brep cannot be converted to a sphere.')
+                face = geometry.Faces.Item[0]
+                if not face.IsSphere():
+                    raise ConversionError('Object brep cannot be converted to a sphere.')
+                result, geometry = face.TryGetSphere()
+                if not result:
+                    raise ConversionError('Object brep cannot be converted to a sphere.')
+            else:
+                raise ConversionError('Rhino object cannot be converted to a sphere: {}'.format(obj))
+        wrapper.geometry = geometry
```
gonzalocasas: these blocks (present in several of the new methods/classes feel like they could become the `coerce###` stuff I just mentioned.... right?

src/compas_rhino/geometry/surface.py
```diff
@@ -4,354 +4,190 @@
...
+        geometry : :class:`Rhino.Geometry.Surface`
+            The geometry object defining a surface.
+
+        Returns
+        -------
+        :class:`RhinoSurface`
+            The Rhino surface wrapper.
+        """
+        if not isinstance(geometry, Rhino.Geometry.Surface):
+            raise NotImplementedError
```
gonzalocasas: `NotImplementedError` or just a `TypeError`? because I don't think we'll ever implement a `RhinoSurface` conversion from anything that is not a surface..

src/compas/geometry/curves/nurbs.py
```diff
@@ -0,0 +1,462 @@
...
+            'degree': int,
+            'is_periodic': bool
+        })
+
+    @property
+    def JSONSCHEMANAME(self):
+        raise NotImplementedError
+
+    def __init__(self, name=None):
+        super().__init__(name=name)
```
beverlylytle: Does this not need to be ipy friendly? Or is the intention that compas nurbs curves and surfaces can only pass in and out of rhino in the form of `data`?

src/compas/geometry/curves/nurbs.py
```diff
@@ -0,0 +1,462 @@
...
+        -------
+        Frame
+            The corresponding local frame.
+
+        """
+        raise NotImplementedError
+
+    def closest_point(self, point, distance=None):
+        """Compute the closest point on the curve to a given point."""
+        raise NotImplementedError
```
gonzalocasas: The idea is that these method's implementation are coming from OCC or Rhino SDK, right? If so, two questions:
1) how about throwing a plugin-specific type of exception, e.g the existing `compas.plugins.PluginNotInstalledError` or maybe a new, more precise one, like `compas.plugins.NoPluggableImplFound`
2) when the surfaces are *not* based on the OCC implementation, how exactly are these things bound to the Rhino SDK ones? I think we discussed this, but I'm lost about where the link is.... :P

gonzalocasas (commented):
Here we go! The PR is growing quite a bit, so, I am pretty sure I've missed or misunderstood parts of it...!

src/compas/geometry/curves/nurbs.py
```diff
@@ -0,0 +1,462 @@
...
+        u : float
+            The value of the curve parameter. Must be between 0 and 1.
+
+        Returns
+        -------
+        Vector
+            The corresponding tangent vector.
+
+        """
+        pass
```
beverlylytle: why `pass` and not `NotImplementedError`?

src/compas/geometry/curves/nurbs.py
```diff
@@ -0,0 +1,462 @@
...
+            curve should be computed. Defaults to 100.
+
+        Returns
+        -------
+        list
+            Points along the curve.
+        """
+        return self.xyz(resolution)
+
+    def point_at(self, u):
```
beverlylytle: Is it convention that Bezier curves are parameterized by `t` and nurbs by `u`?  I personally want them both to be `t`, but maybe that's not the standard thing to do.

src/compas/geometry/surfaces/nurbs.py
```diff
@@ -0,0 +1,381 @@
...
+            'is_u_periodic': bool,
+            'is_v_periodic': bool
+        })
+
+    @property
+    def JSONSCHEMANAME(self):
+        raise NotImplementedError
+
+    def __init__(self, name=None):
+        super().__init__(name=name)
```
beverlylytle: Again, ipy-friendliness? along with all the numpy/lru_cache stuff that follows...

src/compas_rhino/geometry/curve.py
```diff
@@ -3,296 +3,104 @@
...
-        return self.is_polyline() and self.is_closed()
+        if not isinstance(geometry, Rhino.Geometry.Curve):
+            if isinstance(geometry, Line):
+                geometry = line_to_rhino_curve(geometry)
+            elif isinstance(geometry, Circle):
+                geometry = circle_to_rhino_curve(geometry)
+            elif isinstance(geometry, Ellipse):
+                geometry = ellipse_to_rhino_curve(geometry)
+            else:
+                raise ConversionError('The geometry cannot be convertex to a curve.')
```
beverlylytle:
```suggestion
                raise ConversionError('The geometry cannot be converted to a curve.')
```

src/compas/geometry/__init__.py
```diff
@@ -1250,6 +1284,13 @@
 
     'Collection',
     'PointCollection',
+
+    'Curve',
+    'BezierCurve',
+    'NurbsCurve',
```
beverlylytle: Does this and the surface need to be bumped down into the conditional?

beverlylytle (commented):
Maybe this nurbs/rhino stuff was discussed before, but I don't remember...

docs/api/compas_rhino.interop.rst
tomvanmele: yes

src/compas_rhino/geometry/plane.py
tomvanmele: > Should these be `staticmethod`s instead? Because they don't really construct the object in which they are located, but they are very useful and clearly related.

how will this work with `self.geometry`?

tomvanmele: > A related comment: in `compas_fab` we added methods like `coerce_frame` , which take as argument either a compas `Frame` or a rhino `Plane` and return a compas `Frame`. Do we have a similar thing now in this PR that can return the compas object of X regardless of whether the input is already of that type?

this is indeed the case for, for example, `RhinoPlane.from_geometry` which takes both a Rhino plane and a COMPAS frame

src/compas_rhino/geometry/sphere.py
tomvanmele: yes

src/compas_rhino/geometry/surface.py
tomvanmele: +1 for `TypeError`

src/compas/geometry/curves/nurbs.py
tomvanmele: i think the idea would be to intercept the object instantiation like with the artists and have both occ and Rhino provide a plugin for that, but not sure

src/compas/geometry/curves/nurbs.py
tomvanmele: i was thinking of the latter, but we could also go for a different approach using pluggable `__new__` with plugins from Rhino and OCC as mentioned above...

src/compas/geometry/curves/nurbs.py
tomvanmele: should be the latter indeed

src/compas/geometry/curves/nurbs.py
tomvanmele: `t` sounds good

src/compas/geometry/curves/nurbs.py
gonzalocasas: got it! yep, that's what I thought we discussed, but was a bit lost about where we were on that

tomvanmele: this is now solved for the OCC implementation. not yet for Rhino

src/compas/geometry/curves/bezier.py
tomvanmele: they are used inside the bezier curve implementation. not sure if there would be much point to exposing them beyond that...

src/compas_rhino/geometry/mesh.py
tomvanmele: it is breaking, yes, but these are very obscure and undocumented parts of the framework, and i doubt it will affect anyone's code other than my own :)

what do you suggest?

src/compas/geometry/__init__.py
tomvanmele: not anymore

tomvanmele:
both the nurbs curve an surface from OCC are now accessible through the plugin mechanism.
it is not clear to me what exactly we expect from the versions in `compas_rhino.geometry`. to be discussed...

src/compas_rhino/geometry/_geometry.py
```diff
@@ -8,21 +8,26 @@
 from compas.utilities import abstractclassmethod
 
 
-__all__ = ['BaseRhinoGeometry']
-
-
-class BaseRhinoGeometry(object):
-    """Base class for Rhino geometry objects.
+class RhinoGeometry(object):
```
gonzalocasas, line 9: would it be required to alias `BaseRhinoGeometry = RhinoGeometry` for backwards compat?

src/compas_rhino/geometry/curve.py
```diff
@@ -339,95 +147,280 @@ def closest_points(self, points, maxdist=0.0):
...
+    #     """
+    #     self.control_points_on()
+    #     rc, grip = Rhino.Input.RhinoGet.GetGrip("Select control point.")
+    #     if rc != Rhino.Commands.Result.Success:
+    #         return
+    #     if grip.OwnerId != self.guid:
+    #         return
+    #     grip.Select(True, True)
+    #     sc.doc.Views.Redraw()
+    #     return grip
```
gonzalocasas, line 412: delete commented out code?

src/compas_rhino/geometry/mesh.py
gonzalocasas: ok, then let's go ahead

gonzalocasas (approved):
LGTM

PR #903, opened 2021-09-17
Improved API for shape booleans

tomvanmele:

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-* Added `Plane.offset`
 * Added `draw_vertexlabels`, `draw_edgelabels`, `draw_facelabels`, `draw_vertexnormals`, and `draw_facenormals` to `compas_blender.artists.MeshArtist`.
 * Added optional `triangulated` flag to `to_vertices_and_faces` of all shapes.
 * Added `compas.geometry.Geometry` base class.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,14 +6,20 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
-* Added `Plane.offset`
+
 ### Added
 
+* Added `Plane.offset`
 * Added `draw_vertexlabels`, `draw_edgelabels`, `draw_facelabels`, `draw_vertexnormals`, and `draw_facenormals` to `compas_blender.artists.MeshArtist`.
+* Added optional `triangulated` flag to `to_vertices_and_faces` of all shapes.
+* Added `compas.geometry.Geometry` base class.
+* Added `__add__`, `__sub__`, `__and__` to `compas.geometry.Shape` for boolean operations using binary operators.
 
 ### Changed
 
 * Fixed bug in `compas_blender.draw_texts`.
+* Changed default resolution for shape discretisation to 16 for both u and v where relevant.
+* Changed base class of `compas.geometry.Primitive` and `compas.geometry.Shape` to `compas.geometry.Geometry`.
 
 ### Removed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Added optional `triangulated` flag to `to_vertices_and_faces` of all shapes.
 * Added `compas.geometry.Geometry` base class.
 * Added `__add__`, `__sub__`, `__and__` to `compas.geometry.Shape` for boolean operations using binary operators.
+* Added `is_closed` to `compas.geometry.Polyhedron`.
 
 ### Changed
 
--- a/src/compas/geometry/geometry.py
+++ b/src/compas/geometry/geometry.py
@@ -1,0 +1,45 @@
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+from compas.data import Data
+
+
+class Geometry(Data):
+    """Base class for all geometric objects."""
+
+    def __ne__(self, other):
+        # this is not obvious to ironpython
+        return not self.__eq__(other)
+
+    def transform(self, transformation):
+        """Transform the geometry.
+
+        Parameters
+        ----------
+        transformation : :class:`Transformation`
+            The transformation used to transform the geometry.
+
+        Returns
+        -------
+        None
+
+        """
+        raise NotImplementedError
+
+    def transformed(self, transformation):
+        """Returns a transformed copy of this geometry.
+
+        Parameters
+        ----------
+        transformation : :class:`Transformation`
+            The transformation used to transform the geometry.
+
+        Returns
+        -------
+        :class:`Geometry`
+            The transformed geometry.
+        """
+        primitive = self.copy()
+        primitive.transform(transformation)
+        return primitive
--- a/src/compas/geometry/shapes/_shape.py
+++ b/src/compas/geometry/shapes/_shape.py
@@ -68,6 +68,25 @@ class Shape(Geometry):
         return Polyhedron(V, F)
 
     def __and__(self, other):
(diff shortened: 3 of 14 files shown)


tomvanmele:
I have added an optional triangulation parameter to `to_vertices_and_faces`, increased the default resolution to 16 for u and v discretisation, and added a few boolean magic methods.
```python
from compas.geometry import Sphere

A = Box.from_width_height_depth(2, 2, 2)
B = Sphere([1, 1, 1], 1)

C = A + B
D = A - B
E = A & B
``` 

`C`, `D` and `E` are polyhedrons (yes, i know). Tried adding a general solid class but it doesn't make things better. I guess since we allow for skew polygons, we might as well allow for skew polyhedrons. I added `is_closed` to the polyhedron class such that for random inputs of vertices and faces this can be checked. on the upside, since all shapes are triangulated before applying booleans, the resulting polyhedrons are actually perfectly valid :)

If more control over the resolution is desired, this also works.

```python
from compas.geometry import Sphere
from compas.geometry import boolean_union_mesh_mesh
from compas.geometry import boolean_difference_mesh_mesh
from compas.geometry import boolean_intersection_mesh_mesh

box = Box.from_width_height_depth(2, 2, 2)
sphere = Sphere([1, 1, 1], 1)

A = box.to_vertices_and_faces(triangulated=True)
B = sphere.to_vertices_and_faces(u=32, v=32, triangulated=True)

C = boolean_union_mesh_mesh(A, B)
D = boolean_difference_mesh_mesh(A, B)
E = boolean_intersection_mesh_mesh(A, B)
```
The `_mesh_mesh` suffix doesn't feel correct anymore though. Perhaps is should be `_shape_shape`.

While working on this i also realised that our shapes are actually solids and that the boolean operations only work on proper solids. Perhaps we could consider differentiating between primitives (point/vector, line, polyline, curve, plane, frame, quaternion), shapes (circle, ellipse, polygon), and solids (box, sphere, ...)

Anyway, some other issues remain unresolved

* The polyhedron still allows for creating random (skewed) polygonal surfaces that are not closed, which would not be valid input for the booolean operations.
* We don't have an object (other than polyhedron :) ) that allows for creating general polygonal surfaces.

I guess adding a mesh object to the geometry package will be very confusing, but i still think we need something like this...

src/compas/geometry/shapes/_shape.py
```diff
@@ -3,12 +3,77 @@
...
+    def __and__(self, other):
+        from compas.geometry import boolean_intersection_mesh_mesh
+        from compas.geometry import Polyhedron
+        A = self.to_vertices_and_faces(triangulated=True)
+        B = other.to_vertices_and_faces(triangulated=True)
+        V, F = boolean_intersection_mesh_mesh(A, B)
+        return Polyhedron(V, F)
+
+    def __or__(self, other):
+        return self.__add__(other)
```
beverlylytle: Does not compute. 🤖 Why isn't this the union?

tomvanmele:
```python
A + B  # union
A - B  # difference
A & B  # intersection
A | B  # alias for union
```
but maybe the alias should be removed?

tomvanmele: or the other way round?

beverlylytle: oh! sorry, i read `__and__` rather than `__add__`.  this is fine.

CHANGELOG.md
```diff
@@ -9,14 +9,21 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+* Added `Plane.offset`
```
gonzalocasas: looks like a merging issue because this same line is already present a few lines later.

src/compas/geometry/geometry.py
```diff
@@ -0,0 +1,45 @@
...
+from __future__ import print_function
+
+from compas.data import Data
+
+
+class Geometry(Data):
+    """Base class for all geometric objects."""
+
+    def __ne__(self, other):
+        # this is not obvious to ironpython
```
gonzalocasas, line 12: 😆  +1 for moar of these in code

src/compas/geometry/geometry.py
```diff
@@ -0,0 +1,45 @@
...
+        ----------
+        transformation : :class:`Transformation`
+            The transformation used to transform the geometry.
+
+        Returns
+        -------
+        None
+
+        """
+        raise NotImplementedError
```
gonzalocasas, line 28: this is not obvious for me :P  why is `transform` raising, but `transformed` is implemented? (i realize this is not a new change, and just been moved around, but I still don't quite get why it is like this)

src/compas/geometry/shapes/_shape.py
```diff
@@ -3,12 +3,77 @@
 from __future__ import division
 
 import abc
-from compas.geometry import Primitive
+from ..geometry import Geometry
 
 
-class Shape(Primitive):
+class Shape(Geometry):
```
gonzalocasas, line 9: technically, this is a breaking change, but I guess it's not very problematic. However, it would be fair to add a bigger warning/notice in the changelog indicating that if someone depends on this fact, they need to update.

src/compas/geometry/shapes/_shape.py
```diff
@@ -3,12 +3,77 @@
...
+        >>> C = A - B
+        """
+        from compas.geometry import boolean_difference_mesh_mesh
+        from compas.geometry import Polyhedron
+        A = self.to_vertices_and_faces(triangulated=True)
+        B = other.to_vertices_and_faces(triangulated=True)
+        V, F = boolean_difference_mesh_mesh(A, B)
+        return Polyhedron(V, F)
+
+    def __and__(self, other):
```
gonzalocasas, line 70: pls add a docstring to this one and `__or__` as well

src/compas/geometry/shapes/capsule.py
```diff
@@ -263,6 +263,16 @@ def to_vertices_and_faces(self, u=10, v=10):
             nn = len(vertices) - 3 - (j + 1) % u
             faces.append([np, nn, nc])
 
+        if triangulated:
+            triangles = []
+            for face in faces:
+                if len(face) == 4:
```
gonzalocasas, line 269: do we need to check for invalid cases, like `len(face) > 4` and raise?

src/compas/geometry/shapes/polyhedron.py
```diff
@@ -263,16 +267,49 @@ def from_planes(cls, planes):
...
+
+        Returns
+        -------
+        bool
+            True if the polyhedron is closed.
+            False otherwise.
+        """
+        from compas.datastructures import Mesh
+        mesh = Mesh.from_vertices_and_faces(self.vertices, self.faces)
+        return mesh.is_closed()
```
gonzalocasas, line 312: not very critical, but in order to avoid the circular dependency here, perhaps the `is_closed` should be entirely copied in this class (considering it's only a couple of lines) or moved as a shared utility function somewhere else.

gonzalocasas (approved):
LGTM!  I added a bunch of small-ish notes, but it's looking great!

src/compas/geometry/geometry.py
tomvanmele: is an old one by @beverlylytle if i remember correctly :)

src/compas/geometry/geometry.py
tomvanmele: the implementation of transform is different for most objects, but transformed can just call transform (whatever it ends up being)

src/compas/geometry/shapes/_shape.py
tomvanmele: there is this in the log
```
* Changed base class of `compas.geometry.Primitive` and `compas.geometry.Shape` to `compas.geometry.Geometry`.
```

src/compas/geometry/shapes/capsule.py
tomvanmele: technically it is not possible to have an invalid case because we generate the faces and there are only tris and quads

src/compas/geometry/shapes/polyhedron.py
tomvanmele: it is true that it is only a few lines of code, but those depend on a lot of the data structure infrastructure so it is not so simple to copy it out

PR #904, opened 2021-09-19
Universal CAD artists

tomvanmele:

### Checklist


tomvanmele:
Based on @gonzalocasas recipe for making nurbs functionality pluggable...

Added `compas.artists.Artist` and pluggable factory `compas.artists.new_artist`, and plugins `compas_rhino.artists.new_artist_rhino` and `compas_blender.artists.new_artist_blender`.

This enables artist instantiation without CAD-specific imports.

```python
from compas.artists import Artist
from compas.geometry import Polyhedron

shape = Polyhedron.from_platonicsolid(12)
mesh = Mesh.from_shape(shape)

artist = Artist(mesh)
artist.draw()

print(type(artist))  # compas_rhino.artists.meshartist.MeshArtist (if executed in Rhino)
                     # compas_blender.artists.meshartist.MeshArtist (if executed in Blender)
```

src/compas_blender/artists/__init__.py
```diff
@@ -39,6 +49,27 @@
...
+def new_artist_blender(cls, *args, **kwargs):
+    data = args[0]
+    dtype = type(data)
+    if dtype not in Artist.ITEM_ARTIST:
+        raise DataArtistNotRegistered('No Blender artist is registered for this data type: {}'.format(dtype))
+    cls = Artist.ITEM_ARTIST[dtype]
+    for name, value in inspect.getmembers(cls):
+        if inspect.isfunction(value):
+            if hasattr(value, '__isabstractmethod__'):
+                raise Exception('Abstract method not implemented: {}'.format(value))
```
gonzalocasas, line 146: This looks cool as a safe-guard/fail-fast mechanism, but I would consider moving it into the plugin discovery stage, otherwise we will incur this inspection cost on every instantiation of an artist (and also, this is generally a very useful feature to have generalized).   Since moving it to the plugins discovery would need some bigger changes, I think it's ok to just add a `TODO: ` comment here and we go back to it at a later point, especially to not delay this PR much.

gonzalocasas: A quicker alternative would be to move this bit of code into a function like `ensure_abstract_methods_impl`, which does the check but keeps track of classes already checked, so, we can liberally call it anywhere, and it will not be expensive.

src/compas_blender/artists/frameartist.py
```diff
@@ -40,34 +39,13 @@ class FrameArtist(BaseArtist):
...
     def __init__(self,
                  frame: Frame,
                  collection: Optional[bpy.types.Collection] = None,
-                 scale: float = 1.0):
-        super(FrameArtist, self).__init__()
-        self.collection = collection
-        self.frame = frame
+                 scale: float = 1.0,
+                 **kwargs: Any):
+        super().__init__(primitive=frame, collection=collection or frame.name, **kwargs)
```
gonzalocasas, line 51: Just wondering: the `super()` syntax is being intentionally switched to python3 here and everywhere on the `compas_blender` package, right?

src/compas_blender/artists/meshartist.py
```diff
@@ -253,17 +114,15 @@ def draw(self) -> None:
 
         """
         self.clear()
-        if self.show_vertices:
-            self.draw_vertices()
-        if self.show_faces:
-            self.draw_faces()
-        if self.show_edges:
-            self.draw_edges()
+        self.draw_vertices()
```
gonzalocasas: not sure here, why is the check for `show_*` not needed anymore?

src/compas_blender/artists/meshartist.py
```diff
@@ -284,18 +143,18 @@ def draw_vertices(self,
...
             points.append({
                 'pos': self.mesh.vertex_coordinates(vertex),
                 'name': f"{self.mesh.name}.vertex.{vertex}",
-                'color': vertex_color[vertex],
+                'color': self.vertex_color.get(vertex, self.default_vertexcolor),
                 'radius': 0.01
             })
         objects = compas_blender.draw_points(points, self.vertexcollection)
-        self.object_vertex = zip(objects, vertices)
+        self.objects += objects
```
gonzalocasas: I don't see where `self.objects` is cleared, does this mean that now calling `.draw_vertices()` twice in a row will end-up drawing duplicated vertices or did I miss the clear somewhere?

src/compas_rhino/__init__.py
```diff
@@ -88,7 +93,7 @@ def _get_ironpython_lib_path_mac(version):
     lib_paths = {
         '5.0': ['/', 'Applications', 'Rhinoceros.app', 'Contents'],
         '6.0': ['/', 'Applications', 'Rhinoceros.app', 'Contents', 'Frameworks', 'RhCore.framework', 'Versions', 'A'],
-        '7.0': ['/', 'Applications', 'Rhino 7.app', 'Contents', 'Frameworks', 'RhCore.framework', 'Versions', 'A']
+        '7.0': ['/', 'Applications', 'Rhinoceros.app', 'Contents', 'Frameworks', 'RhCore.framework', 'Versions', 'A']
```
gonzalocasas, line 96: is this correct? I guess doesn't belong to this technically, but regardless of that.. I believe this was working in the past, right? did rhino install defaults change?

src/compas/artists/__init__.py
```diff
@@ -0,0 +1,56 @@
...
+from __future__ import print_function
+from __future__ import absolute_import
+from __future__ import division
+
+from .exceptions import DataArtistNotRegistered
+from .artist import Artist
+from .meshartist import MeshArtist
+from .networkartist import NetworkArtist
+from .primitiveartist import PrimitiveArtist
+from .robotmodelartist import BaseRobotModelArtist
```
gonzalocasas: how about renaming this to `RobotModelArtist` to match all others? (and maybe keep the alias `BaseRobotModelArtist = RobotModelArtist` for backwards compat)

beverlylytle:
Just so you know, I'm currently paused on a PR where I'm adding some Blender artists, waiting for you to finish this. So far I've got point, vector, and line as well as a version of a primitive artist.

gonzalocasas (changes requested):
Generally looking very good, but there's an important missing part here: these also needs to be done for `compas_ghpython` before merging, and that might require some extra logic to figure out how to do the detection on every call to `new_artist`, because a one-time resolution of the plugin will not be enough (user can switch back and forth between GH and Rhino).  Since the plugin system doesn't really have a way to determine which document is active in Rhino (GH's or Rhino's), we might need to have one unified `new_artist` for both that knows this.. or figure out a way to add arbitrary constraints/filters that are tested every single time, not just on discovery.

tomvanmele:
> Just so you know, I'm currently paused on a PR where I'm adding some Blender artists, waiting for you to finish this. So far I've got point, vector, and line as well as a version of a primitive artist.

okay i just added box and sphere.
shall i keep adding shapes and you take care of the primitives?

src/compas_blender/artists/frameartist.py
tomvanmele: yes

beverlylytle:
> shall i keep adding shapes and you take care of the primitives?

sounds good

src/compas_rhino/__init__.py
tomvanmele: we don't really use this anymore, but in any case it is only correct if you have multiple versions of Rhino installed and named one of them `Rhino 7`.

i think we should add an additional option to specify the path of the Rhino installation folder...

src/compas_blender/artists/artist.py
```diff
@@ -0,0 +1,47 @@
...
+            self._collection = compas_blender.create_collection(value)
+        else:
+            raise Exception('Collection must be of type `str` or `bpy.types.Collection`.')
+
+    def clear(self):
+        """Delete all objects created by the artist."""
+        if not self.objects:
+            return
+        compas_blender.delete_objects(self.objects)
+        self.objects = []
```
beverlylytle: @gonzalocasas  this is where blender objects are cleared.

src/compas_blender/artists/meshartist.py
beverlylytle: but yes, i think that `clear` will clear out anything that is in `self.objects` and that if `draw_vertices` is called multiple times duplicate objects will be created. but this is behavior is consistent both before and after this pr.

beverlylytle: ah, perhaps i misunderstood.  are you suggesting that `self.clear()` is called at the start of any `draw*` method?

tomvanmele:
i have a few pending updates to push that might address the clearing concerns...

src/compas/artists/__init__.py
```diff
@@ -0,0 +1,58 @@
...
+from .exceptions import DataArtistNotRegistered
+from .artist import Artist
+from .meshartist import MeshArtist
+from .networkartist import NetworkArtist
+from .primitiveartist import PrimitiveArtist
+from .robotmodelartist import RobotModelArtist
+from .shapeartist import ShapeArtist
+from .volmeshartist import VolMeshArtist
+
+BaseRobotModelArtist = RobotModelArtist
```
gonzalocasas, line 47: Maybe add an internal deprecation note. Not sure if this specific suggestion is the way, but we should come up with a way to tag all these things that we intend to get rid of before 2.0 so that we can do a search-all-files at the right time.

```suggestion
# NOTE: deprecated, will be removed in the next major release
BaseRobotModelArtist = RobotModelArtist
```

src/compas/artists/artist.py
```diff
@@ -0,0 +1,33 @@
...
+from __future__ import division
+from __future__ import print_function
+
+from abc import abstractmethod
+from compas.plugins import pluggable
+
+
+@pluggable(category='factories')
+def new_artist(cls, *args, **kwargs):
+    raise NotImplementedError
```
gonzalocasas, line 21: same comment as in #888: maybe these things should not throw `NotImplementedError` but something that clearly signals (both to the user of compas and also to the developer/user reading code to understand) that this is an interface and the plugin is actually implemented (most likely) but just not here. So, something like `compas.plugins.NoPluggableImplFound` or similar.

src/compas_blender/artists/__init__.py
```diff
@@ -8,40 +8,139 @@
...
+from .robotmodelartist import RobotModelArtist
+from .sphereartist import SphereArtist
+from .torusartist import TorusArtist
+
+
+@plugin(category='factories', pluggable_name='new_artist', tryfirst=True, requires=['bpy'])
+def new_artist_blender(cls, *args, **kwargs):
+    # "lazy registration" seems necessary to avoid item-artist pairs to be overwritten unintentionally
+
+    BlenderArtist.register(Box, BoxArtist)
```
gonzalocasas: Doesn't this registration trigger every time a new artist is created? (Probably side-effect free, but still unnecessary)

gonzalocasas (changes requested):
I added some comments here, but things were not working well on GH, so, I sent a PR to merge into this before we continue reviewing: https://github.com/compas-dev/compas/pull/909

gonzalocasas:
It's hard to really make sure there are no breaking changes in this one because of the size and the fact that all changes happen in non-testable parts of the framework. Maybe after another round of fixes, we could split up and try to manually test groups of features in the team (would also help to list things were we are missing feature parity between the different CADs).

src/compas/artists/artist.py
```diff
@@ -0,0 +1,33 @@
...
+    def register(item_type, artist_type):
+        Artist.ITEM_ARTIST[item_type] = artist_type
+
+    @abstractmethod
+    def draw(self):
+        raise NotImplementedError
+
+    @staticmethod
+    def draw_collection(collection):
+        raise NotImplementedError
```
beverlylytle, line 51: I'm confused why this is on the base `Artist` and not on the `BlenderArtist`.

CHANGELOG.md
```diff
@@ -15,13 +15,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Added `__add__`, `__sub__`, `__and__` to `compas.geometry.Shape` for boolean operations using binary operators.
 * Added `is_closed` to `compas.geometry.Polyhedron`.
 * Added `Plane.offset`.
+* Added `compas.artists.Artist`.
+* Added pluggable `compas.artists.new_artist`.
+* Added plugin `compas_rhino.artists.new_artist_rhino`.
+* Added plugin `compas_blender.artists.new_artist_blender`.
+* Added 'compas.artist.DataArtistNotRegistered'.
```
```suggestion
* Added `compas.artist.DataArtistNotRegistered`.
```

src/compas/artists/artist.py
tomvanmele: ha perhaps the name is confusing. was mean to be a static method to draw many objects at the same time. is not related to the "collection" of blender. but since there is no implementation for it anywhere, perhaps it can just be removed...

src/compas_rhino/utilities/drawing.py
```diff
@@ -88,7 +88,7 @@ def wrap_drawfunc(f):
     def wrapper(*args, **kwargs):
         layer = kwargs.get('layer', None)
         clear = kwargs.get('clear', False)
-        redraw = kwargs.get('redraw', True)
+        redraw = kwargs.get('redraw', False)
```
beverlylytle, line 91: I don't like this... if a script is run on mac with the `EditPythonScript` command, then the canvas freezes.  so then, just in `EditPythonScript` one always has to remember to bookend the script with `import compas_rhino .... compas_rhino.redraw()`

tomvanmele: i can change it back, but it will not make much difference, because then all of the individual draw functions will set it to false, which is how it was before. just figured it would be less repetitive this way...

beverlylytle: mmm... then can there be a `redraw` method on `RhinoArtist`?  it's slightly less annoying to type `artist.redraw()`

tomvanmele: i we redraw every draw function, stuff pops up in Rhino 1 object at a time. for example in the case of meshes, this can be really annoying

tomvanmele: > mmm... then can there be a `redraw` method on `RhinoArtist`? it's slightly less annoying to type `artist.redraw()`

as a static method?
i could add clear like this as well...

beverlylytle: yeah, i agree it's bad to always have redraw enabled

beverlylytle: > as a static method?

yes

>i could add clear like this as well...

what would `clear` do?  clear all layers?

tomvanmele: clear would be like `compas_rhino.clear()`. it clears everything...

tomvanmele:
```python
    @staticmethod
    def clear():
        compas_rhino.clear()

    @staticmethod
    def redraw():
        compas_rhino.redraw()
```

tomvanmele (commented):
haha, apparently, because i am the original author of the PR i can't approve the changes, since "pull request authors cannot approve their own pull requests" :)

tomvanmele:
i guess @gonzalocasas this means that you will have to approve of yourself :)

beverlylytle (approved):
Here I come to save the day!

tomvanmele:
merging still blocked since @gonzalocasas requested changes at some point...

gonzalocasas (approved):

gonzalocasas:
how picky this thing gets! ;)
I approved the changes, but I could add the missing shape artists to this before we merge.. or shall I do it in a new PR?

tomvanmele:
since @beverlylytle will also still add the primitives for blender in a separate PR i would vote for merging!

PR #909, opened 2021-10-07
Artist factory plugins

gonzalocasas:
Changed the plugin system to allow defining runtime selection based on arbitrary conditions, used that to select between rhino and ghpython artists, and fixed a few missing bits about Ghpython artists.

The ghpython artists still don't all work as expected, in particular, the "canonical" mesh artist example from all our slides fails because ghpython's meshartist is missing some abstract method implementations (`clear*` stuff). Robot and Frame artists work. I didn't test much more thou. I didn't test blender either.

### DIFF
--- a/docs/devguide.rst
+++ b/docs/devguide.rst
@@ -393,8 +393,8 @@ There are a few additional options that plugins can use:
 * ``requires``: List of requirements. COMPAS will filter out plugins if their
   requirements list is not satisfied at runtime. This allows to have multiple implementations
   of the same operation and have them selected based on different criteria.
-  The requirement can either be a packages name string (e.g. ``requires=['scipy']``) or
-  a ``callable``, in which any arbitrary check can be implemented
+  The requirement can either be a package name string (e.g. ``requires=['scipy']``) or
+  a ``callable`` with a boolean return value, in which any arbitrary check can be implemented
   (e.g. ``requires=[lambda: is_rhino_active()]``).
 * ``tryfirst`` and ``trylast``: Plugins cannot control the exact priority they will have
   but they can indicate whether to try to prioritize them or demote them as fallback using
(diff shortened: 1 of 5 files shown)


src/compas/plugins.py
```diff
@@ -321,9 +321,11 @@ def plugin(method=None, category=None, requires=None, tryfirst=False, trylast=Fa
         The method to decorate as ``plugin``.
     category : str, optional
         An optional string to group or categorize plugins.
-    requires : list of str, optional
-        Optionally defines a list of packages that should be importable
-        for this plugin to be used.
+    requires : list, optional
+        Optionally defines a list of requirements that should be fulfilled
+        for this plugin to be used. The requirement can either be a packages
```
beverlylytle: package or package's

docs/devguide.rst
```diff
@@ -390,10 +390,12 @@ Advanced options
 
 There are a few additional options that plugins can use:
 
-* ``requires``: List of required python modules. COMPAS will filter out plugins if their
+* ``requires``: List of requirements. COMPAS will filter out plugins if their
   requirements list is not satisfied at runtime. This allows to have multiple implementations
-  of the same operation and have them selected based on which packages are installed.
-  on the system. Eg. `requires=['scipy']`.
+  of the same operation and have them selected based on different criteria.
+  The requirement can either be a packages name string (e.g. ``requires=['scipy']``) or
```
beverlylytle: package or package's

docs/devguide.rst
```diff
@@ -390,10 +390,12 @@ Advanced options
...
+  a ``callable``, in which any arbitrary check can be implemented
```
beverlylytle: maybe a couple of words about what is expected of this callable would be useful here... it's usage is basically assuming that the callable is boolean-valued based on whether an import was successful.

beverlylytle (commented):
This looks very tidy.

tomvanmele (approved):

PR #917, opened 2021-10-26
Artist subclass

beverlylytle:
Some classes subclass classes with artists.  For example, `Origin` is essentially a `Frame`, and should be drawn as such.

### Checklist

### DIFF
--- a/src/compas_blender/artists/__init__.py
+++ b/src/compas_blender/artists/__init__.py
@@ -136,7 +136,7 @@ def new_artist_blender(cls, *args, **kwargs):
         dtype = type(data)
         for type_ in BlenderArtist.ITEM_ARTIST:
             if issubclass(dtype, type_):
-                cls = BlenderArtist.ITEM_ARTIST[dtype]
+                cls = BlenderArtist.ITEM_ARTIST[type_]
                 break
         else:
             raise DataArtistNotRegistered('No GH artist is registered for this data type: {}'.format(dtype))
--- a/src/compas_blender/artists/__init__.py
+++ b/src/compas_blender/artists/__init__.py
@@ -139,7 +139,7 @@ def new_artist_blender(cls, *args, **kwargs):
                 cls = BlenderArtist.ITEM_ARTIST[type_]
                 break
         else:
-            raise DataArtistNotRegistered('No GH artist is registered for this data type: {}'.format(dtype))
+            raise DataArtistNotRegistered('No Blender artist is registered for this data type: {}'.format(dtype))
 
     # TODO: move this to the plugin module and/or to a dedicated function
 
--- a/src/compas_blender/artists/__init__.py
+++ b/src/compas_blender/artists/__init__.py
@@ -134,9 +134,12 @@ def new_artist_blender(cls, *args, **kwargs):
         cls = kwargs['artist_type']
     else:
         dtype = type(data)
-        if dtype not in BlenderArtist.ITEM_ARTIST:
-            raise DataArtistNotRegistered('No Blender artist is registered for this data type: {}'.format(dtype))
-        cls = BlenderArtist.ITEM_ARTIST[dtype]
+        for type_ in BlenderArtist.ITEM_ARTIST:
+            if issubclass(dtype, type_):
+                cls = BlenderArtist.ITEM_ARTIST[dtype]
+                break
+        else:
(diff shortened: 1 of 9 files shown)


src/compas_blender/artists/__init__.py
```diff
@@ -134,9 +134,12 @@ def new_artist_blender(cls, *args, **kwargs):
         cls = kwargs['artist_type']
     else:
         dtype = type(data)
-        if dtype not in BlenderArtist.ITEM_ARTIST:
-            raise DataArtistNotRegistered('No Blender artist is registered for this data type: {}'.format(dtype))
-        cls = BlenderArtist.ITEM_ARTIST[dtype]
+        for type_ in BlenderArtist.ITEM_ARTIST:
+            if issubclass(dtype, type_):
+                cls = BlenderArtist.ITEM_ARTIST[type_]
+                break
```
tomvanmele: i think this can be problematic because it doesn't respect class hierarchy.

for example, consider `FormDiagram` which is a sub class of `Mesh` and is registered with a `FormDiagramArtist` that provides more specific visualisation functionality than `MeshArtist`. there is nothing in this loop that gives precedence to `FormDiagramArtist` over `MeshArtist`.

not sure if it would solve the problem entirely, but perhaps checking for direct registration before running the loop could help.

```python
if dtype in BlenderArtist.ITEM_ARTIST:
    cls = BlenderArtist.ITEM_ARTIST[dtype]
else:
    for type_ in BlenderArtist.ITEM_ARTIST:
        if issubclass(dtype, type_):
            cls = BlenderArtist.ITEM_ARTIST[type_]
            break
```

src/compas_blender/artists/__init__.py
```diff
@@ -134,9 +134,12 @@ def new_artist_blender(cls, *args, **kwargs):
...
+        else:
+            raise DataArtistNotRegistered('No GH artist is registered for this data type: {}'.format(dtype))
```
gonzalocasas:
```suggestion
            raise DataArtistNotRegistered('No Blender artist is registered for this data type: {}'.format(dtype))
```

src/compas_rhino/artists/__init__.py
```diff
@@ -200,9 +200,12 @@ def new_artist_rhino(cls, *args, **kwargs):
...
         dtype = type(data)
-        if dtype not in RhinoArtist.ITEM_ARTIST:
-            raise DataArtistNotRegistered('No Rhino artist is registered for this data type: {}'.format(dtype))
-        cls = RhinoArtist.ITEM_ARTIST[dtype]
+        for type_ in RhinoArtist.ITEM_ARTIST:
+            if issubclass(dtype, type_):
+                cls = RhinoArtist.ITEM_ARTIST[type_]
+                break
+        else:
+            raise DataArtistNotRegistered('No GH artist is registered for this data type: {}'.format(dtype))
```
gonzalocasas:
```suggestion
            raise DataArtistNotRegistered('No Rhino artist is registered for this data type: {}'.format(dtype))
```

gonzalocasas (approved):
LGTM, but I would definitely like to have one or two unit tests to check for this sub-class matching. And if you're feeling inspired, refactoring the repetitive code that we already have in place for the `new_artist` stuff into a more general location would be ideal, but if not, we can also leave it for another PR.

tomvanmele (changes requested):
sorry, but we need to at least first have a discussion about the class hierarchy thing. my concern was not really hypothetical. the formdiagram example is a real case...

src/compas_blender/artists/__init__.py
beverlylytle: I take your point, but where does `FormDiagram` exist?  I guess there should be tests or examples somewhere about registering external artists.

tomvanmele: `FormDiagram` exists in `compas_tna` but it applies to any artist type that builds upon a base artist. there are many such cases in `compas_assembly`, `compas_ags`, `compas_fofin`, ...

perhaps we can set up an artificial test where we make a custom mesh class which is registered to a custom mesh artist and check if the correct type of artist is indeed constructed?

gonzalocasas:
@brgcode we now are handling correctly the type hierarchy with the latest change by @beverlylytle ... 
Re-review?

tomvanmele (approved):
very nice!

PR #918, opened 2021-10-29
Fix robot model artist for Blender

gonzalocasas:

### Checklist

### DIFF
--- a/src/compas_blender/artists/artist.py
+++ b/src/compas_blender/artists/artist.py
@@ -26,7 +26,6 @@ class BlenderArtist(Artist):
         self.collection = collection
         super().__init__(**kwargs)
 
-
     @property
     def collection(self) -> bpy.types.Collection:
         return self._collection
--- a/src/compas_blender/artists/artist.py
+++ b/src/compas_blender/artists/artist.py
@@ -21,11 +21,11 @@ class BlenderArtist(Artist):
     def __init__(self,
                  collection: Optional[Union[str, bpy.types.Collection]] = None,
                  **kwargs: Any):
-
-        super().__init__(**kwargs)
-
+        # Initialize collection before even calling super because other classes depend on that
         self._collection = None
         self.collection = collection
+        super().__init__(**kwargs)
+
 
     @property
     def collection(self) -> bpy.types.Collection:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 * Changed new artist registration to check if subclass.
+* Fixed `RobotModelArtist` for blender: missing abstract method impl and handle init order.
 
 ### Removed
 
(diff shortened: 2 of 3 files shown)


gonzalocasas:
This is a pretty annoying bug, so, after review ok and merge, I would suggest to create a `1.9.2` release asap.

beverlylytle (approved):
review, ok and ...

src/compas_blender/artists/artist.py
```diff
@@ -21,11 +21,11 @@ class BlenderArtist(Artist):
...
                  collection: Optional[Union[str, bpy.types.Collection]] = None,
                  **kwargs: Any):
-
-        super().__init__(**kwargs)
-
+        # Initialize collection before even calling super because other classes depend on that
         self._collection = None
         self.collection = collection
+        super().__init__(**kwargs)
+
```
beverlylytle, line 28:
```suggestion
```

gonzalocasas: is the suggestion to remove the `super()` or to remove an empty line?

gonzalocasas: ah... got it ;)

gonzalocasas: (committed the fix from my local files instead of gh web ui, the suggestion should have be marked as outdated, but it isn't... whatever)

gonzalocasas:
@beverlylytle added one more fix: `clear()` should do the hide_set magic, instead of deleting the objects

beverlylytle:
I'm not sure that that's true.  In  the base `BlenderArtist` we have 
```
    def clear(self) -> None:
        """Delete all objects created by the artist."""
        compas_blender.delete_objects(self.collection.objects)

gonzalocasas:
> I'm not sure that that's true. In the base `BlenderArtist` we have
> 
> ```
>     def clear(self) -> None:
>         """Delete all objects created by the artist."""
>         compas_blender.delete_objects(self.collection.objects)
> ```

yes, and that indeed will delete the objects, but it's not what the robot model artist expects.. I dont recall why, but I remember this was changed long ago for hiding instead of deleting (it was an external contribution).

beverlylytle:
I recall that the instantiation of the artist resulted in the drawing of all the collision and visual geometries, without calling `draw*`.  So that was changed... the `create` method creates all the geometries and then hides them.  Then `draw*` would unhide them.  but it feels weird that `delete` would have different behaviors for different artists.   maybe there should be `hide_*` methods.

tomvanmele:
fwiw, i agree with @beverlylytle that the behaviour should be consistent. however, it is not clear to me what the use case for `hide/show` would be. is this to not have to redraw some of the meshes in between different calls to a script?

beverlylytle:
for a robot model, it would be nice to toggle whether you are looking at the collision geometries or the visual geometries, without all the computational effort of reading and parsing the URDF holding all the geometric data.  that would be the use case of `hid/show`.

gonzalocasas:
alright, changed the artist to really delete when calling `clear` and added a bit of code to ensure the thing recreates geometry when needed.

tomvanmele:
> for a robot model, it would be nice to toggle whether you are looking at the collision geometries or the visual geometries, without all the computational effort of reading and parsing the URDF holding all the geometric data.  that would be the use case of `hid/show`.

fair enough, but i guess this toggling would only be possible in an interactive scenario or in gh, no?

tomvanmele (approved):

PR #920, opened 2021-11-02
Integrate OCC nurbs through pluggables

tomvanmele:

### Checklist

### DIFF
--- a/src/compas/geometry/curves/nurbs.py
+++ b/src/compas/geometry/curves/nurbs.py
@@ -1,0 +1,495 @@
+from __future__ import print_function
+from __future__ import absolute_import
+from __future__ import division
+
+from math import sqrt
+
+from compas.plugins import pluggable
+from compas.geometry import Point
+from compas.geometry import Frame
+from compas.utilities import linspace
+
+from .curve import Curve
+
+
+@pluggable(category='factories')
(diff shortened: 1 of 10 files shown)


tomvanmele:
have separated this from the Rhino interop PR since they are independent contributions...

src/compas/geometry/curves/nurbs.py
```diff
@@ -0,0 +1,495 @@
...
+from compas.geometry import Point
+from compas.geometry import Frame
+from compas.utilities import linspace
+
+from .curve import Curve
+
+
+@pluggable(category='factories')
+def new_nurbscurve(cls, *args, **kwargs):
+    raise NotImplementedError
```
gonzalocasas, line 17: I will keep bugging about this: a specific exception that clearly states this is an extension-point that is filled in by code somewhere else would be very good. A lot of our users read the source code, and this leads them to think the feature is literally not implemented, but that is not really this case for it.

gonzalocasas (approved):
LGTM

src/compas/geometry/curves/nurbs.py
tomvanmele: will do this in separate pr for all pluggables

PR #923, opened 2021-11-06
Fix bug in continuous loops and strips

tomvanmele:

### Checklist


tests/compas/datastructures/test_halfedge.py
```diff
@@ -240,3 +251,30 @@ def test_del_edge_attribute_in_view(mesh, edge_key):
     del attrs["foo"]
     with pytest.raises(KeyError):
         attrs["foo"]
+
+
+# ==============================================================================
+# Tests - Loops & Strip
+# ==============================================================================
+
+def test_loops_and_strips(sphere):
```
beverlylytle: It would be good to have tests for each of the terminal `if` statements. For example, the `meridian` below is a closed loop.  A test for the cases that would use  the `vu_loop` I think would be a good idea.

tomvanmele: you mean to check the loop for a mesh that is not closed?

beverlylytle: i mean there are many `if` statements that `return` and the given test doesn't check all these 'leaves'.

src/compas/datastructures/halfedge/halfedge.py
```diff
@@ -2092,76 +2092,75 @@ def edge_loop(self, edge):
...
-        edges = []
-        current, previous = edge
-        edges.append((previous, current))
+        u, v = edge
+        uv_loop = self.halfedge_loop((u, v))
+        if uv_loop[0][0] == uv_loop[-1][1]:
+            return uv_loop
+        vu_loop = self.halfedge_loop((v, u))
+        vu_loop[:] = [(u, v) for v, u in vu_loop[::-1]]
+        return vu_loop + uv_loop[1:]
```
beverlylytle, line 2101: for example, i think the test exits with the first `if` and the code afterward isn't covered.

beverlylytle (approved):
lgtm!

PR #928, opened 2021-11-09
Primitive Blender Artists

beverlylytle:
Adds artists for primitives.

Also removes functionality from `RBGColour` and `Float`, but doesn't remove them entirely as that would break the api.

### Checklist


src/compas/utilities/colors.py
```diff
@@ -2,6 +2,9 @@
 from __future__ import absolute_import
 from __future__ import division
 
+from typing import Tuple
+from typing import Union
```
tomvanmele: with this, none of the color stuff is importable in Rhino...

src/compas_blender/artists/boxartist.py
```diff
@@ -27,7 +29,7 @@ def __init__(self,
 
         super().__init__(shape=box, collection=collection or box.name, **kwargs)
 
-    def draw(self, color=None):
+    def draw(self, color: RGBColor = None) -> List[bpy.types.Object]:
```
tomvanmele: i think this would need to be `color: Optional[RGBColor] = None`

src/compas_blender/artists/capsuleartist.py
```diff
@@ -28,12 +30,12 @@ def __init__(self,
 
         super().__init__(shape=capsule, collection=collection or capsule.name, **kwargs)
 
-    def draw(self, color=None, u=None, v=None):
+    def draw(self, color: RGBColor = None, u: int = None, v: int = None) -> List[bpy.types.Object]:
```
tomvanmele: i guess these would all need to be `Optional[...]`

src/compas_blender/artists/capsuleartist.py
```diff
@@ -28,12 +30,12 @@ def __init__(self,
...
         """Draw the capsule associated with the artist.
 
         Parameters
         ----------
-        color : tuple of float, optional
+        color : tuple of float or tuple of int, optional
```
tomvanmele, line 38: unfortunately, in the case of blender it has to be `tuple of float`, between 0 and 1. rhino on the other hand requires the values to be `tuple of int` between 0 and 255.

src/compas_blender/artists/circleartist.py
```diff
@@ -0,0 +1,65 @@
...
+        The name of the collection the object belongs to.
+    """
+
+    def __init__(self,
+                 circle: Circle,
+                 collection: Optional[Union[str, bpy.types.Collection]] = None,
+                 **kwargs: Any):
+        super().__init__(primitive=circle, collection=collection or circle.name, **kwargs)
+
+    def draw(self,  color: RGBColor = None, show_point=False, show_normal=False) -> List[bpy.types.Object]:
```
tomvanmele: here also, `Optional[RGBColor]`

src/compas_blender/artists/circleartist.py
```diff
@@ -0,0 +1,65 @@
...
+        """Draw the circle.
+
+        Parameters
+        ----------
+        color : tuple of float or tuple of int, optional
```
tomvanmele, line 38: tuple of float

src/compas_blender/artists/circleartist.py
```diff
@@ -0,0 +1,65 @@
...
+            Default is ``False``.
+
+        Returns
+        -------
+        list
+            The objects created in Blender.
+        """
+        color = color or self.color
+        point = list(self.primitive.plane.point)
+        normal = list(self.primitive.plane.normal)
```
tomvanmele: not sure, but i don't think the conversion to list i necessary in blender/cpython...

src/compas_blender/artists/frameartist.py
```diff
@@ -89,13 +91,13 @@ def draw_axes(self) -> List[bpy.types.Object]:
...
         list of :class:`bpy.types.Object`
         """
-        origin = list(self.frame.point)
-        X = list(self.frame.point + self.frame.xaxis.scaled(self.scale))
-        Y = list(self.frame.point + self.frame.yaxis.scaled(self.scale))
-        Z = list(self.frame.point + self.frame.zaxis.scaled(self.scale))
+        origin = list(self.primitive.point)
+        X = list(self.primitive.point + self.primitive.xaxis.scaled(self.scale))
+        Y = list(self.primitive.point + self.primitive.yaxis.scaled(self.scale))
+        Z = list(self.primitive.point + self.primitive.zaxis.scaled(self.scale))
```
tomvanmele: same here about the conversion

src/compas_blender/artists/frameartist.py
```diff
@@ -89,13 +91,13 @@ def draw_axes(self) -> List[bpy.types.Object]:
...
         lines = [
-            {'start': origin, 'end': X, 'color': self.color_xaxis, 'name': f"{self.frame.name}.xaxis"},
-            {'start': origin, 'end': Y, 'color': self.color_yaxis, 'name': f"{self.frame.name}.yaxis"},
-            {'start': origin, 'end': Z, 'color': self.color_zaxis, 'name': f"{self.frame.name}.zaxis"},
+            {'start': origin, 'end': X, 'color': self.color_xaxis, 'name': f"{self.primitive.name}.xaxis"},
+            {'start': origin, 'end': Y, 'color': self.color_yaxis, 'name': f"{self.primitive.name}.yaxis"},
+            {'start': origin, 'end': Z, 'color': self.color_zaxis, 'name': f"{self.primitive.name}.zaxis"},
```
tomvanmele, line 101: somewhat of a "note to self"... perhaps it would make sense to start using named tuples for this. they have a much lower memory footprint and fit the purpose of these mini dicts better...

src/compas_blender/artists/meshartist.py
```diff
@@ -166,16 +166,29 @@ def draw(self,
              vertices: Optional[List[int]] = None,
              edges: Optional[List[Tuple[int, int]]] = None,
              faces: Optional[List[int]] = None,
-             vertexcolor: Optional[Union[str, Color, List[Color], Dict[int, Color]]] = None,
-             edgecolor: Optional[Union[str, Color, List[Color], Dict[int, Color]]] = None,
-             facecolor: Optional[Union[str, Color, List[Color], Dict[int, Color]]] = None
+             vertexcolor: Optional[Union[str, RGBColor, List[RGBColor], Dict[int, RGBColor]]] = None,
```
tomvanmele, line 169: perhaps also here we should stop allowing for colors to be specified in hex format

src/compas_blender/artists/meshartist.py
```diff
@@ -75,9 +75,9 @@ def __init__(self,
                  vertices: Optional[List[int]] = None,
                  edges: Optional[List[int]] = None,
                  faces: Optional[List[int]] = None,
-                 vertexcolor: Optional[Color] = None,
-                 edgecolor: Optional[Color] = None,
-                 facecolor: Optional[Color] = None,
+                 vertexcolor: Optional[RGBColor] = None,
+                 edgecolor: Optional[RGBColor] = None,
+                 facecolor: Optional[RGBColor] = None,
```
tomvanmele, line 80: perhaps we should provide a default value here

src/compas_blender/artists/pointartist.py
```diff
@@ -0,0 +1,69 @@
...
+import bpy
+
+import compas_blender
+from compas.artists import PrimitiveArtist
+from compas.geometry import Point
+from compas.utilities import RGBColor
+from compas_blender.artists import BlenderArtist
+
+
+__all__ = ["PointArtist"]
```
tomvanmele: i started to remove all of this since we don't do any dynamic importing anymore...

src/compas_blender/artists/polygonartist.py
```diff
@@ -0,0 +1,66 @@
...
+                 polygon: Polygon,
+                 collection: Optional[Union[str, bpy.types.Collection]] = None,
+                 **kwargs: Any):
+        super().__init__(primitive=polygon, collection=collection or polygon.name, **kwargs)
+
+    def draw(self,
+             color: RGBColor = None,
+             show_points=False,
+             show_edges=False,
+             show_face=True) -> List[bpy.types.Object]:
```
tomvanmele: `show_points: bool = False`

src/compas_blender/artists/polygonartist.py
```diff
@@ -0,0 +1,66 @@
...
+        show_edges : bool, optional
+            Default is ``False``.
+        show_face : bool, optional
+            Default is ``True``.
+
+        Returns
+        -------
+        list of bpy.types.Object
+        """
+        _points = map(list, self.primitive.points)
```
tomvanmele: not sure if this is necessary

src/compas_blender/artists/robotmodelartist.py
```diff
@@ -37,17 +38,17 @@ def transform(self, native_mesh: bpy.types.Object, transformation: Transformatio
...
-                        color: Union[Tuple[int, int, int, int], Tuple[float, float, float, float]] = None
+                        color: Union[RGBColor, Tuple[int, int, int, int], Tuple[float, float, float, float]] = None
                         ) -> bpy.types.Object:
         # Imported colors take priority over a the parameter color
         if 'mesh_color.diffuse' in geometry.attributes:
             color = geometry.attributes['mesh_color.diffuse']
 
         # If we have a color, we'll discard alpha because draw_mesh is hard coded for a=1
-        if color:
+        if color and len(color) == 4:
```
tomvanmele: you could just take the slice and not even check

tomvanmele (changes requested):
generally looks good. just a few recurring things (use of `Optional` and the difference between colors in Rhino and Blender), and perhaps the unnecessary conversions from objects to lists...

src/compas/utilities/colors.py
beverlylytle: Yeah, I will move it to `compas_blender.utilities`.

src/compas_blender/artists/boxartist.py
beverlylytle: Will change.

src/compas_blender/artists/capsuleartist.py
beverlylytle: no, in blender it looks like you can provide either float or int and it is interpreted correctly.

src/compas_blender/artists/circleartist.py
beverlylytle: will check this

src/compas_blender/artists/meshartist.py
beverlylytle: Aren't those set by the `self.default_vertexcolor` in the base `MestArtist`?

src/compas_blender/artists/pointartist.py
beverlylytle: will do

src/compas_blender/artists/capsuleartist.py
tomvanmele: yes, i  know, but the meaning is not the same. red in blender is still `(1, 0, 0)`. increasing components beyond 1 will increase the brightness of the pixels. if you do `(255, 0, 0)`, there seems to be a light on in your mesh...

src/compas_blender/artists/meshartist.py
tomvanmele: yes true. perhaps we should update the network artist to align with this one then...

src/compas_blender/artists/capsuleartist.py
beverlylytle: and we want to disallow this?

tomvanmele: no, i guess not. perhaps it is more confusing the other way around. if you are used to the 0-1 color range and try to set colors like that in Rhino they are almost black...

tomvanmele (approved):

PR #929, opened 2021-11-09
Also try to remove broken symlinks during uninstall

tomvanmele:

### Checklist


tomvanmele:
If a user install a bunch of packages from an environment, and then removes the environment without uninstalling the packages this can create problems when trying to install again later on.

The uninstall process now also removes all broken symlinks...

Licini (approved):
Looks good!

tomvanmele:
the following will now also happen during the install procedure:

1. all installed packages that are not available in the current env will be removed (albeit quietly)
2. all installed packages that are available will be reinstalled

tomvanmele:
not sure if this is what we mean to happen though.
however, it seems to me that packages from a different env should not be allowed to float around in a new install...

src/compas_rhino/install.py
```diff
@@ -62,10 +64,20 @@ def install(version=None, packages=None):
     symlinks_to_uninstall = []
     exit_code = 0
 
+    for name in os.listdir(scripts_path):
+        if name.startswith('compas') and not name.endswith('.py'):
```
beverlylytle: Would it make sense for this to also include all `installable_rhino_packages`?  In the case that `packages` is provided by the user, then something like `roslibpy` (which is a default installable package for `compas_fab`) would be missed by this check and also would not be added to `packages` by `_filter_installable_packages` to be removed and added again.

CHANGELOG.md
```diff
@@ -20,6 +20,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+* Changed `compas_rhino.uninstall` to also remove broken symlinks if no specific packages are provided for installation.
+* Changed `compas_rhino.install` to also remove broken symlinks if no specific packages are provided for installation.
```
beverlylytle: It looks like broken symlinks are removed whether or not packages are specified on install.

src/compas_rhino/uninstall.py
```diff
@@ -45,13 +45,25 @@ def uninstall(version=None, packages=None):
...
+    ipylib_path = compas_rhino._get_ironpython_lib_path(version)
+
+    # If no specific packages are provided for uninstall
+    # everything should be removed,
+    # and not just the default packages
+    # or the packages that the bootstrapper is aware of.
+    if not packages:
+        packages = []
+        for name in os.listdir(scripts_path):
+            if name.startswith('compas') and not name.endswith('.py'):
```
beverlylytle: again, `roslibpy` or other default compas-installed packages?

CHANGELOG.md
```diff
@@ -20,6 +20,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+* Changed `compas_rhino.uninstall` to also remove broken symlinks if no specific packages are provided for installation.
```
beverlylytle: 'for un-installation'?

src/compas_rhino/uninstall.py
tomvanmele: is this also the case here?

beverlylytle: yes, i think so.

src/compas_rhino/install.py
```diff
@@ -47,25 +47,55 @@ def install(version=None, packages=None):
...
     results = []
     symlinks_to_install = []
     symlinks_to_uninstall = []
     exit_code = 0
 
+    # check all installable packages
+    # add the packages that can't be imported from the current env to the list of symlinks to uninstall
+    # and remove the package name from the list of installable packages
+    # make a copy of the list to avoid problems with removing items
+    # note: perhaps this should already happen in the filter function...
```
beverlylytle, line 72: This makes sense to me.

PR #930, opened 2021-11-10
Add Rhino plugins for NURBS curve pluggables

tomvanmele:

### Checklist


tomvanmele:
In Rhino you can now do

```python
from compas.geometry import NurbsCurve

points = [Point(0, 0, 0), Point(3, 6, 0), Point(6, -3, 3), Point(10, 0, 0)]

nurbs = NurbsCurve.from_points(points)

print(nurbs)
print(nurbs.data)
```

or 

```python
import compas_rhino
from compas_rhino.conversions import RhinoCurve
from compas.geometry import NurbsCurve

guid = compas_rhino.select_curve()

curve = RhinoCurve.from_guid(guid)
nurbs = NurbsCurve.from_rhino(curve.geometry)

print(nurbs)
print(nurbs.data)
```

src/compas_rhino/geometry/curves/__init__.py
```diff
@@ -0,0 +1,28 @@
+from .nurbs import RhinoNurbsCurve
+
+try:
+    from compas.geometry import NurbsCurve
```
gonzalocasas, line 4: Why is this in a try.. except block? 
What is the condition that would make it not importable?

src/compas_rhino/geometry/curves/__init__.py
```diff
@@ -0,0 +1,28 @@
...
+except ImportError:
+    pass
+else:
+    from compas.plugins import plugin
+
+    @plugin(category='factories', requires=['Rhino'])
+    def new_nurbscurve(cls, *args, **kwargs):
+        return super(NurbsCurve, RhinoNurbsCurve).__new__(RhinoNurbsCurve)
```
gonzalocasas, line 12: Is this one really correct with a `super` call here?

src/compas_rhino/geometry/__init__.py
```diff
@@ -5,65 +5,8 @@
...
-    'RhinoLine',
-    'RhinoMesh',
-    'RhinoPlane',
-    'RhinoPoint',
-    'RhinoPolyline',
-    'RhinoSphere',
-    'RhinoSurface',
-    'RhinoVector',
-]
+__all__ = []
```
gonzalocasas: Don't we need to alias all these moved classes to avoid breaking compat?

src/compas_rhino/geometry/curves/nurbs.py
```diff
@@ -0,0 +1,527 @@
...
+from compas.geometry import Point
+# from compas.geometry import Vector
+# from compas.geometry import Transformation
+# from compas.geometry import Frame
+# from compas.geometry import Circle
+# from compas.geometry import Box
+from compas.utilities import linspace
+
+try:
+    from compas.geometry import NurbsCurve
```
gonzalocasas, line 6: Again, unclear to me why this can fail, i think it deserves a comment, because it's not evident.

src/compas_rhino/geometry/curves/nurbs.py
```diff
@@ -0,0 +1,527 @@
...
+
+        """
+        curve = cls()
+        curve.rhino_curve = Rhino.Geometry.NurbsCurve.CreateHSpline([point_to_rhino(point) for point in points])
+        return curve
+
+    @classmethod
+    def from_step(cls, filepath):
+        """Load a NURBS curve from an STP file."""
+        pass
```
gonzalocasas: Maybe raise instead of pass?

src/compas_rhino/geometry/curves/nurbs.py
```diff
@@ -0,0 +1,527 @@
...
+
+    @classmethod
+    def from_arc(cls, arc, degree, pointcount=None):
+        pass
```
gonzalocasas: Same

src/compas_rhino/geometry/curves/nurbs.py
```diff
@@ -0,0 +1,527 @@
...
+        curve.rhino_curve = Rhino.Geometry.NurbsCurve.CreateFromLine(line_to_rhino(line))
+        return curve
+
+    # ==============================================================================
+    # Conversions
+    # ==============================================================================
+
+    def to_step(self, filepath, schema="AP203"):
+        """Write the curve geometry to a STP file."""
+        pass
```
gonzalocasas: These 3 methods as well, i guess they need to raise an exception instead of passing

gonzalocasas (approved):
Yay!! I added a couple of notes here and there but basically lgtm!

src/compas_rhino/geometry/curves/__init__.py
tomvanmele: this can go away indeed. was necessary at some point in between two versions...

src/compas_rhino/geometry/curves/__init__.py
tomvanmele: i think so, but i will check

gonzalocasas:
This is looking awesome! 🤩

PR #932, opened 2021-11-12
bug fix in mesh_slice_plane

nizartaha:
if intersection_segment_plane returns the mesh vertex (u or v of the edge ). it should NOT append the mesh vertex to intersections.
otherwise, mesh_split_face in split method will raise ValueError: The split vertices are neighbors.

### Checklist


src/compas/datastructures/mesh/slice.py
```diff
@@ -146,11 +146,7 @@ def intersect(self):
             x = intersection_segment_plane((a, b), self.plane)
             if not x:
                 continue
-            if len(x) == sum([1 for i, j in zip(x, a) if i == j]):
-                intersections.append(u)
-            elif len(x) == sum([1 for i, j in zip(x, b) if i == j]):
-                intersections.append(v)
-            else:
+            if not len(x) == sum([1 for i, j in zip(x, a) if i == j]) and not len(x) == sum([1 for i, j in zip(x, b) if i == j]):
```
beverlylytle: You want to check whether there is at least one index `n` where `x[n]` is not equal to `a[n]` and similarly for `x` and `b`, right?  I find the following to be an easier-to-read version of that, and also, it should run much faster.

```suggestion
            if any([i != j for i, j in zip(x, a)]) and any([i != j for i, j in zip(x, b)]):
```

but this is only a local comment.  I don't know enough about slice to approve this...

nizartaha: Thank you @beverlylytle :) i tested it , it works. its a bit slower though ( 04.549066 sec, instead of 04.491115) , but definitely shorter and easier-to-read

nizartaha: @brgcode What do you think of this? thank you

tomvanmele (approved):
lgtm.
however, can you remind me about why it is needed?

tomvanmele:
ha sorry it is in the description :)

tomvanmele:
can you still add something to the changelog?

tomvanmele:
if i understand correctly this deals with the case that the intersection result is the same as one of the end points of the segment?

tomvanmele:
would also be nice to add a test for that :)
i know it is annoying, but it is good practice!

nizartaha:
Sure, i will add something to changelog.

correct, it is the `intersection_segment_plane` return result.

hmm test. yes :) it will require to write separate code for it and the specific PLY mesh, where it raised the error. where do i upload the mesh?

tomvanmele:
i would say that tests can be based on a simple shape like a box, combined with different types of specific slicing planes for which you know what is supposed to happen. for example

* plane aligned with one of the sides of the box
* plane through a corner only
* plane with creating a triangle interface
* plane creating a quad interface
* ...

nizartaha:
i made again some changes after i wrote a test file. I tested it with `python 3.9.9` and `compas 1.13.3`

```
from compas.geometry import Plane, Frame, Box, Rotation
from compas.utilities import i_to_rgb

from compas_view2.app import App
from compas_view2.shapes import Text

# Box mesh

box = Box(Frame.worldXY(), 3, 3, 3)
mesh_from_box = Mesh.from_shape(box)

# slice planes

planes = []

# plane creating a quad interface
plane = Plane([0, 1, 0], [0, -1, 0])
planes.append(plane)

# plane creating a triangle interface
plane = Plane(box.diagonal[0], [1, 0, 0])
T = Rotation.from_axis_and_angle([0, 0, 1], 0.785398)
plane.transform(T)
planes.append(plane)

# plane through two corners and two edges
plane = Plane(box.diagonal[1], [1, 0, 0])
T = Rotation.from_axis_and_angle([0, 0, 1], 0.785398*3.5, box.diagonal[1])
plane.transform(T)
planes.append(plane)

# plane through a corner only
plane = Plane(box.diagonal[0], [1, 0, 0])
T = Rotation.from_axis_and_angle([0, 0, 1], 0.785398*3.5, box.diagonal[0])
plane.transform(T)
T = Rotation.from_axis_and_angle([0, 1, 0], 1, box.diagonal[0])
plane.transform(T)
planes.append(plane)

# plane through four corners
f = Frame.from_points(box.vertices[0], box.vertices[2], box.vertices[6])
planes.append(Plane.from_frame(f))

# # plane aligned with one of the sides of the box (will returen None)
# plane = Plane(box.diagonal[1], [0, 0, 1])
# planes.append(plane)

# Slices

slices = []

for i, plane in enumerate(planes):
    copy = mesh_from_box.copy()
    mesh, B = copy.slice_plane(plane)
    slices.append(B)

slices.append(mesh)

# viz

viewer = App(viewmode='wireframe')

for v in mesh_from_box.vertices():
    t = Text(f"{v}", mesh_from_box.vertex_attributes(v,"xyz"), height=40)
    viewer.add(t)

for i, plane in enumerate(planes):
    viewer.add(plane, color=i_to_rgb(i/len(planes), normalize=True))

for i, mesh in enumerate(slices):
    viewer.add(mesh, facecolor=i_to_rgb(i/len(slices), normalize=True))

viewer.run()
```

tomvanmele:
you have some lint that is making the tests fail...
https://github.com/compas-dev/compas/runs/4729176459?check_suite_focus=true#step:4:329

nizartaha:
it should be ok now :)

PR #934, opened 2021-11-20
Clean Rhino install

tomvanmele:

### Checklist

### DIFF
--- a/src/compas_rhino/__init__.py
+++ b/src/compas_rhino/__init__.py
@@ -78,7 +78,7 @@ def _get_ironpython_lib_path(version):
         # However, since it is only used to clean up potential old installations
         # it does not seem crucial anymore.
         ironpython_lib_path = None
-        print("The lib folder for IronPython does not exist in this location: {}".format(ironpython_lib_path))
+        # print("The lib folder for IronPython does not exist in this location: {}".format(ironpython_lib_path))
 
     return ironpython_lib_path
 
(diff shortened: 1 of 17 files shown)


src/compas_rhino/__init__.py
```diff
@@ -36,6 +36,8 @@
 PURGE_ON_DELETE = True
 
 INSTALLABLE_PACKAGES = ['compas', 'compas_rhino', 'compas_ghpython']
+SUPPORTED_VERSIONS = ['5.0', '6.0', '7.0', '8.0']
+DEFAULT_VERSION = '6.0'
```
gonzalocasas: shouldn't we default to 7.0 already?

gonzalocasas (approved):
I think this is good to go!

My only concern is this idea that only importable modules can be installed, because it could be that a package is only runnable under a specific runtime context which is not the installation one. I guess this never happens (right now at least) on the top-level packages, eg. `compas_blender` does not import `bpy` on the top-level package; but it's not completely unthinkable this might break at some point.

src/compas_rhino/__init__.py
tomvanmele: yes, would be more than happy to, but thought it might be considered a breaking change...

gonzalocasas: good point... I did not consider it breaking change because it's not a runtime behavior... of course, it's debatable, but at least, if you have `compas` already installed, and you conda update it, nothing would break due to this...

PR #941, opened 2021-12-03
bug in compas.geometry.normal_polygon 

juney-lee:
The cross products from each edge should be halved before summing. The length of this normal_polygon should be equal to area_polygon.


src/compas/geometry/_core/normals.py
```diff
@@ -61,9 +61,9 @@ def normal_polygon(polygon, unitized=True):
         n = cross_vectors(oa, ob)
         oa = ob
 
-        nx += n[0]
-        ny += n[1]
-        nz += n[2]
+        nx += n[0] / 2
+        ny += n[1] / 2
+        nz += n[2] / 2
```
tomvanmele: i would do `0.5 * n[...]`

https://stackoverflow.com/questions/15745819/why-is-division-more-expensive-than-multiplication

tomvanmele (commented):
you will have to add something to the changelog, because this is not a typo but a bug.
would also be good to add a test, because if there would have been one this could not have happened...

juney-lee:
```
from compas.geometry import area_polygon
from compas.geometry import length_vector

polygon = [
    (0, 0, 0),
    (10, 0, 0),
    (10, 10, 0),
    (0, 10, 0)
]

normal = normal_polygon(polygon, unitized=False)
area = area_polygon(polygon)

print(length_vector(normal))
print(area_polygon(polygon))
```

juney-lee:
This was probably never noticed because, in almost all situations, unitized polygon normals are needed (default is True)... the only time (that I can think of) where one would need a non-unitized polygon normal (so area with a direction) is for 3GS applications. I guess I still should have caught it a while back...

CHANGELOG.md
```diff
@@ -11,6 +11,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+* Fixed bug in `normal_polygon` in `compas.geometry._core`.
```
gonzalocasas: Since the changelog is user-facing, the references should be to the public API, not the internal implementation structure:
```suggestion
* Fixed bug in `normal_polygon` in `compas.geometry`.
```

gonzalocasas (commented):
I would say this definitely needs a test that exemplifies the bug (and the fixed behavior)

juney-lee:
@brgcode @gonzalocasas added a test...

tests/compas/geometry/test_core.py
```diff
@@ -202,3 +203,14 @@ def test_area_triangle(triangle, R):
...
+
+# ==============================================================================
+# normals
+# ==============================================================================\
+
+def test_normal_polygon():
+    polygon = [(0, 0, 0), (10, 0, 0), (10, 10, 0), (0, 10, 0)]
+    normal = normal_polygon(polygon, unitized=False)
+    area = area_polygon(polygon)
+    assert length_vector(normal) == area
```
tomvanmele: this could fail due to floating point precision. please use `assert close(area, length_vector(normal))`.
a sanity check would also be good `assert close(area, 100.0)`

tomvanmele (approved):
LGTM

PR #942, opened 2021-12-03
Update Check PR

Licini:
Somehow current PR-check workflow still let PRs with unchanged Changelog pass. This should fix it. See example at: https://github.com/compas-dev/compas/runs/4406776339?check_suite_focus=true


gonzalocasas:
Strange that this is the case, but in any case, LGTM!

gonzalocasas (approved):

PR #944, opened 2021-12-09
inform version in after_rhino_install

Licini:
In many occasions like RV2, IGS etc it's essential for `after_rhino_install` function to be aware of the chosen version in `compas_rhino.install()`. This commit is to pass it in as parameter, however this means all existing plugins would have to update their `after_rhino_install` to include version parameter as well. I couldn't come up with a non-breaking way of doing it yet. Maybe we can discuss here a bit what's the best way of addressing this.

### DIFF
--- a/src/compas_rhino/install.py
+++ b/src/compas_rhino/install.py
@@ -195,7 +195,7 @@ def install(version=None, packages=None, clean=False):
 
     if exit_code == 0 and len(installed_packages):
         print('\nRunning post-installation steps...\n')
-        if not _run_post_execution_steps(after_rhino_install(installed_packages)):
+        if not _run_post_execution_steps(after_rhino_install(installed_packages, version)):
             exit_code = -1
 
     print('\nInstall completed.')
@@ -265,7 +265,7 @@ def installable_rhino_packages():
 
 
 @compas.plugins.pluggable(category='install', selector='collect_all')
-def after_rhino_install(installed_packages):
+def after_rhino_install(installed_packages, version):
     """Allows extensions to execute actions after install to Rhino is done.
 
     Extensions providing Rhino or Grasshopper features
--- a/src/compas_rhino/install.py
+++ b/src/compas_rhino/install.py
@@ -15,11 +15,11 @@ import compas.plugins
 __all__ = [
     'install',
     'installable_rhino_packages',
-    'after_rhino_install',
-    'INSTALLED_VERSION'
+    'after_rhino_install'
 ]
 
 INSTALLED_VERSION = None
+
 
 def install(version=None, packages=None, clean=False):
     """Install COMPAS for Rhino.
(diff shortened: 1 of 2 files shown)


gonzalocasas:
Isn't this a very breaking change?

tomvanmele:
perhaps set the default to `None` and do nothing if that is the case?

Licini:
> Isn't this a very breaking change?

Yes it absolutely is 🥲, I'm thinking perhaps we can detect the arguments count of `after_rhino_install` of each plugin, then only send `version` in case the `after_rhino_install` is configured to accept this additional parameter.

Licini:
> perhaps set the default to `None` and do nothing if that is the case?

That will still require changes on each individual plugins when defining `after_rhino_install`. I think we just need to figure out a way to only send in version parameter if `after_rhino_install` is configured to accept it. I'm now just trying to figure out how this can be done with decorator wrapped functions

gonzalocasas:
@Licini would it work to not pass it as argument and instead rely on the type of detection that we do already for gh user components: https://github.com/compas-dev/compas/blob/main/src/compas_ghpython/components/__init__.py#L47

if so, that function would need to move to a more readily-available location.

Licini:
> @Licini would it work to not pass it as argument and instead rely on the type of detection that we do already for gh user components: https://github.com/compas-dev/compas/blob/main/src/compas_ghpython/components/__init__.py#L47
> 
> if so, that function would need to move to a more readily-available location.

Ha, yes this would indeed work for our scenarios too! The only outliner might happen in the future is when `compas_rhino.install` is used programatically without terminal arguments. But for now we can do the same for sure. 

Another idea I have is to save version as a global parameter when `install` is called, so it can be accessed later such as `compas_rhino.RHINO_VERSION` ?

src/compas_rhino/install.py
```diff
@@ -202,6 +204,9 @@ def install(version=None, packages=None, clean=False):
     if exit_code != 0:
         sys.exit(exit_code)
 
+    global INSTALLED_VERSION
```
Licini, line 207: @gonzalocasas @brgcode how about something like this?

tomvanmele: sounds good to me

src/compas_rhino/install.py
```diff
@@ -16,8 +16,10 @@
     'install',
     'installable_rhino_packages',
     'after_rhino_install',
+    'INSTALLED_VERSION'
```
gonzalocasas: I would not make it part of the official API, it's more an internal thing, so, I'd remove it from `__all__`. Other than that, LGTM!

tomvanmele: agreed about the public API thing

Licini: makes sense! adjusted now

tomvanmele (approved):
LGTM

PR #946, opened 2021-12-10
Assembly suggestions

gonzalocasas:
A few small changes to the assembly classes.

### Checklist


gonzalocasas:
@brgcode  Oh, this diff makes no sense because I merged `main` into this branch and now it looks gigantic... Either we merge `main` also on `assembly` to get a clean(er) diff, or you review just the last 4 individual commits which are pretty small.

src/compas/datastructures/assembly/assembly.py
```diff
@@ -78,7 +88,7 @@ def add_part(self, part, key=None, **kwargs):
         """
         key = self.graph.add_node(key=key, part=part, **kwargs)
         part.key = key
-        self.parts[part.guid] = part
+        self._parts[part.guid] = part
```
tomvanmele, line 93: i guess the point is to keep track of parts globally. so maybe we should check if a part already exists in the assembly before adding it?

tomvanmele (approved):

src/compas/datastructures/assembly/assembly.py
gonzalocasas: good point, added at https://github.com/compas-dev/compas/pull/946/commits/27729ed3a913feb1ce81fd1ec9418b27de302d79

PR #947, opened 2021-12-10
Clean up the application paths for Rhino/GH

tomvanmele:

### Checklist

### DIFF
--- a/src/compas_rhino/__init__.py
+++ b/src/compas_rhino/__init__.py
@@ -39,6 +39,38 @@ INSTALLABLE_PACKAGES = ['compas', 'compas_rhino', 'compas_ghpython']
 SUPPORTED_VERSIONS = ['5.0', '6.0', '7.0', '8.0']
 DEFAULT_VERSION = '7.0'
 
+IRONPYTHON_PLUGIN_GUID = '814d908a-e25c-493d-97e9-ee3861957f49'
+GRASSHOPPER_PLUGIN_GUID = 'b45a29b1-4343-4035-989e-044e8580d9cf'
+RHINOCYCLES_PLUGIN_GUID = '9bc28e9e-7a6c-4b8f-a0c6-3d05e02d1b97'
+
+
+__all__ = [
+    'PURGE_ON_DELETE',
+    'INSTALLABLE_PACKAGES',
+    'SUPPORTED_VERSIONS',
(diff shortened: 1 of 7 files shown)


src/compas_rhino/__init__.py
```diff
@@ -60,172 +92,223 @@ def _check_rhino_version(version):
...
-        python_plugins_path = _get_python_plugins_path_win32(version)
-
+        app = os.path.join(os.getenv('ProgramFiles'), 'Rhino {}'.format(version))
     elif compas.OSX:
-        python_plugins_path = _get_python_plugins_path_mac(version)
+        paths = {
+            '5': ['/', 'Applications', 'Rhinoceros.app'],
+            '6': ['/', 'Applications', 'Rhinoceros.app'],
+            '7': ['/', 'Applications', 'Rhino 7.app', ],
+            '8': ['/', 'Applications', 'RhinoWIP.app', ]
```
gonzalocasas, line 150: I think this part will need a re-work at some point. It probably needs a more elaborate check, like, if we're trying to install on `7`. We could search for `['Rhinoceros.app', `Rhino 7.app`, `Rhino7.app`] and maybe others, and then check in the metadata in that folder to make sure it really is what we want.

gonzalocasas (approved):
Generally LGTM, but I didn't have time to go deep into some of the details of the change. Do you want to get this merged and included in today's release?

tomvanmele:
not necessarily, but we could. as far as i can tell (and intended), it just does what already was happening before, but just in a much more readable and cleaner way...

src/compas_rhino/__init__.py
tomvanmele: yes, absolutely. but just to be clear, i haven't changed the functionality (also not this part) just cleaned it up...

tomvanmele:
i leave it up to you, considering the workshop on monday. there is still some time for testing though. i can install all possible versions on my windows laptop and on parallels to check if it still behaves as before...

gonzalocasas:
Let's go for it then...! 
I can do the testing on Windows, and you do the mac test (which you probably already did).

src/compas_rhino/__init__.py
gonzalocasas: yep yep, i know... it's just that every time I read it, i remember "uh, we should change this"

PR #952, opened 2021-12-14
Rhino plugin for nurbs surfaces

tomvanmele:

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Added `compas_rhino.geometry.surfaces.new_nurbssurface_from_points` plugin.
 * Added `compas_rhino.geometry.surfaces.new_nurbssurface_from_fill` plugin.
 * Added `compas_rhino.geometry.surfaces.new_nurbssurface_from_step` plugin.
+* Added `compas_rhino.conversions.RhinoSurface.to_compas`.
 
 ### Changed
 
@@ -31,6 +32,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Changed implementation of `compas.geometry.NurbsSurface.xyz`.
 * Fixed bug in `compas.geometry.NurbsSurface.to_mesh`.
 * Changed `compas_rhino.geometry.RhinoNurbsSurface.from_points` to use transposed points.
+* Fixed bug in `compas_rhino.conversions.RhinoSurface.to_compas_mesh`.
 
 ### Removed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Changed `compas_plotters.artists.MeshArtist.draw_vertexlabels` to use `compas_plotters.Plotter.fontsize`.
 * Changed `compas_plotters.artists.MeshArtist.draw_edgelabels` to use `compas_plotters.Plotter.fontsize`.
 * Changed `compas_plotters.artists.MeshArtist.draw_facelabels` to use `compas_plotters.Plotter.fontsize`.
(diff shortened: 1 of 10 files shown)


src/compas_rhino/geometry/__init__.py
```diff
@@ -94,5 +96,6 @@
     'RhinoSurface',
     'RhinoVector',
 
-    'RhinoNurbsCurve'
+    'RhinoNurbsCurve',
+    'RhinoNurbsSurface',
```
Licini, line 100: Sorry my first time looking at this part. I'm a bit confused here `RhinoSurface` and `RhinoNurbsSurface` are very similarly named, and they are more or less both wrappers of Rhino geometry instances, however they have quite different kind of APIs, what's the designed context for each of these classes?

tomvanmele: `RhinoSurface` is a backwards compatibility import from `compas_rhino.conversions` that allows Rhino surface objects and geometries to be converted to compas equivalents.

`RhinoNurbsSurface` is the Rhino plugin implementation of `compas.geometry.NurbsSurface`

src/compas_rhino/geometry/surfaces/nurbs.py
```diff
@@ -0,0 +1,490 @@
...
+from compas_rhino.conversions import point_to_compas
+from compas_rhino.conversions import vector_to_compas
+from compas_rhino.conversions import plane_to_compas_frame
+from compas_rhino.conversions import box_to_compas
+from compas_rhino.conversions import RhinoCurve
+
+import Rhino.Geometry
+
+
+def rhino_surface_from_parameters(points, weights, u_knots, v_knots, u_mults, v_mults, u_degree, v_degree, is_u_periodic=False, is_v_periodic=False):
```
Licini, line 20: Nitpicking here, just feel a bit strange of this function hanging here outside with a class, could it be part of class or in a collection of helpers in case it can be re-used elsewhere? (Same goes with RhinoNurbsCurve)

Licini (approved):
well, generally LGTM!

src/compas_rhino/geometry/__init__.py
Licini: I see!

CHANGELOG.md
```diff
@@ -21,6 +28,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Changed `compas_plotters.artists.MeshArtist.draw_vertexlabels` to use `compas_plotters.Plotter.fontsize`.
 * Changed `compas_plotters.artists.MeshArtist.draw_edgelabels` to use `compas_plotters.Plotter.fontsize`.
 * Changed `compas_plotters.artists.MeshArtist.draw_facelabels` to use `compas_plotters.Plotter.fontsize`.
+* Fixed bug in `compas_rhino.conversions.plane_to_compas_frame`.
+* Changed implementation of `compas.geometry.NurbsSurface.xyz`.
```
gonzalocasas, line 32: Not to change right now, but these messages should be a tad more specific, both `fixed bug` and `changed impl` are so generic that a user would need to go read the code to understand what's being talked about.

src/compas_rhino/conversions/surface.py
```diff
@@ -40,7 +41,30 @@ def geometry(self, geometry):
...
+        Returns
+        -------
+        :class:`compas_rhino.geometry.RhinoNurbsSurface`
+
+        Raises
+        ------
+        :class:`ConversionError`
+            If the surface BRep contains more than one face.
+        """
+        from compas.geometry import NurbsSurface
```
gonzalocasas, line 56: any reason for this import to be inlined instead of at the top?

gonzalocasas (approved):
sorry the late reviewing...
LGTM!!! 💯

src/compas_rhino/conversions/surface.py
tomvanmele: was originally because of a circular import but perhaps this is not the case anymore. will investigate...

PR #957, opened 2021-12-19
Artist docstrings

tomvanmele:

### Checklist

### DIFF
--- a/src/compas/artists/meshartist.py
+++ b/src/compas/artists/meshartist.py
@@ -60,7 +60,7 @@ class MeshArtist(Artist):
     edge_width : dict[tuple[int, int], float]
         Edge widths.
         Defaults to 1.
-        Visuzlization of edges with variable width is not available for all visualization contexts.
+        Visualization of edges with variable width is not available for all visualization contexts.
 
     Class Attributes
     ----------------
--- a/src/compas/artists/meshartist.py
+++ b/src/compas/artists/meshartist.py
@@ -56,7 +56,7 @@ class MeshArtist(Artist):
     vertex_size : dict[int, float]
         Vertex sizes.
         Defaults to 1.
-        Visuzlization of vertices with variable size is not available for all visualization contexts.
+        Visualization of vertices with variable size is not available for all visualization contexts.
     edge_width : dict[tuple[int, int], float]
         Edge widths.
         Defaults to 1.
--- a/src/compas/artists/meshartist.py
+++ b/src/compas/artists/meshartist.py
@@ -76,6 +76,7 @@ class MeshArtist(Artist):
         The default size of the vertices of the mesh.
     default_edgewidth : float
         The default width of the edges of the mesh.
+
     """
 
     default_color = (0.0, 0.0, 0.0)
@@ -305,6 +306,7 @@ class MeshArtist(Artist):
         -------
         list
             The identifiers of the objects representing the vertices in the visualization context.
+
         """
         raise NotImplementedError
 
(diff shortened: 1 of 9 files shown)


src/compas/artists/meshartist.py
```diff
@@ -16,57 +16,72 @@ class MeshArtist(Artist):
...
+    edge_text : dict[tuple[int, int], str]
+        Edge labels.
+        Defaults to the edge identifiers.
+    face_text : dict[int, tuple[float, float, float]]
+        Face labels.
+        Defaults to the face identifiers.
+    vertex_size : dict[int, float]
+        Vertex sizes.
+        Defaults to 1.
+        Visuzlization of vertices with variable size is not available for all visualization contexts.
```
gonzalocasas:
```suggestion
        Visualization of vertices with variable size is not available for all visualization contexts.
```

src/compas/artists/meshartist.py
```diff
@@ -16,57 +16,72 @@ class MeshArtist(Artist):
...
+    edge_width : dict[tuple[int, int], float]
+        Edge widths.
+        Defaults to 1.
+        Visuzlization of edges with variable width is not available for all visualization contexts.
```
gonzalocasas:
```suggestion
        Visualization of edges with variable width is not available for all visualization contexts.
```

src/compas/artists/robotmodelartist.py
```diff
@@ -362,19 +373,21 @@ def update_tool(self, joint_state=None, visual=True, collision=True, transformat
 
         Parameters
         ----------
-        joint_state : :obj:`dict`or :class:`compas.robots.Configuration`, optional
+        joint_state : dictor :class:`compas.robots.Configuration`, optional
```
gonzalocasas: this looks like a type, i think it was correct as it was 
```
        :obj:`dict` or :class:`compas.robots.Configuration`
```
optionally, if you wanna be super correct, this is a `Union[dict, compas.robots.Configuration]`, right?

src/compas/artists/robotmodelartist.py
```diff
@@ -362,19 +373,21 @@ def update_tool(self, joint_state=None, visual=True, collision=True, transformat
...
             A dictionary with joint names as keys and joint positions as values.
             Defaults to an empty dictionary.
         transformation : :class:`compas.geometry.Transformation`, optional
             The (absolute) transformation to apply to the entire tool's geometry.
-            If ``None`` is given, no additional transformation will be applied.
-            Defaults to ``None``.
+            If None is given, no additional transformation will be applied.
```
gonzalocasas, line 393: why the removal of the keyword styling?

gonzalocasas (approved):
LGTM! only some minor feedback

gonzalocasas:
Some of the tilde-prefixed links are not working (locally at least). I don't think this was affected by this PR, but since you're "there", maybe wanna fix it also:
image

src/compas/artists/robotmodelartist.py
tomvanmele: `Union` would be the typing version indeed. it just doesn't produce very legible results so i opted for the "or" version for those cases...

src/compas/artists/robotmodelartist.py
tomvanmele: i stopped doing :obj:`int` and :obj:`float` etc because it clutters the actual docstring and doesn't really add anything useful in the rendered version (in my opinion, but that is of course debatable).

similarly (and following the numpy style guide), i stopped adding those backticks on None, True, False... but also debatable

PR #958, opened 2021-12-19
Data docstrings

tomvanmele:

### Checklist

### DIFF
--- a/src/compas/data/data.py
+++ b/src/compas/data/data.py
@@ -121,17 +121,17 @@ class Data(object):
 
     @property
     def DATASCHEMA(self):
-        """:class:`schema.Schema` - The schema of the data of this object."""
+        """schema.Schema : The schema of the data of this object."""
         raise NotImplementedError
 
     @property
     def JSONSCHEMANAME(self):
-        """str - The schema of the data of this object in JSON format."""
+        """str : The schema of the data of this object in JSON format."""
         raise NotImplementedError
 
     @property
     def JSONSCHEMA(self):
-        """dict - The schema of the JSON representation of the data of this object."""
+        """dict : The schema of the JSON representation of the data of this object."""
         if not self._JSONSCHEMA:
             schema_filename = '{}.json'.format(self.JSONSCHEMANAME.lower())
             schema_path = os.path.join(os.path.dirname(__file__), 'schemas', schema_filename)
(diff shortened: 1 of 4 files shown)


src/compas/data/data.py
```diff
@@ -28,15 +28,77 @@
...
-        The schema of the data dict.
-    JSONSCHEMA : dict
-        The schema of the serialized data dict.
+    dtype : str, read-only
+        The type of the object in the form of a fully qualified module name and a class name, separated by a forward slash ("/").
+        For example: ``"compas.datastructures/Mesh"``.
     data : dict
-        The fundamental data describing the object.
-        The structure of the data dict is defined by the implementing classes.
+        The representation of the object as a dictionary containing only native Python data.
```
gonzalocasas: I would make this one very clear about what the dictionary should contain, since I remember being confused myself multiple times. Does `only native Python data` mean `only Python built-in data types`? Because an instance of a user-created class would still qualify as native Python in some definitions, so, if we want to avoid that, I would go for the "built-in data type" wording.

src/compas/data/data.py
```diff
@@ -28,15 +28,77 @@
...
+
     Attributes
     ----------
-    DATASCHEMA : :class:`schema.Schema`
-        The schema of the data dict.
-    JSONSCHEMA : dict
-        The schema of the serialized data dict.
+    dtype : str, read-only
+        The type of the object in the form of a fully qualified module name and a class name, separated by a forward slash ("/").
+        For example: ``"compas.datastructures/Mesh"``.
```
gonzalocasas, line 40: This is *not* a suggestion for a change in this PR at all, but maybe to start a conversation later on, but, what's the reason for using slash as separator instead of doing a really fully-qualified module & name with dots, ie. `compas.datastructures.Mesh`? Is it some legacy due to the way xfunc used to work?

src/compas/data/data.py
```diff
@@ -203,12 +257,12 @@ def from_jsonstring(cls, string):
         Parameters
         ----------
         string : str
-            The JSON string.
+            The object JSON string.
```
gonzalocasas:
```suggestion
            The object as a JSON string.
```

src/compas/data/data.py
```diff
@@ -59,16 +121,17 @@ def __setstate__(self, state):
 
     @property
     def DATASCHEMA(self):
```
gonzalocasas, line 123: These properties are not present in the doc output. At least locally, did I miss some config? I have latest versions of both sphinx and our sphinx theme.

image

gonzalocasas (approved):
LGTM + minor feedback!

src/compas/data/data.py
tomvanmele: think i got this from pickle and/or numpy

src/compas/data/data.py
tomvanmele: no indeed. figured they are to much of an experimental developer thing to expose in the docs for now...

PR #963, opened 2022-01-03
Data structure docstrings

tomvanmele:

### Checklist


src/compas/datastructures/__init__.py
```diff
@@ -15,7 +15,6 @@
     Datastructure
     Graph
     HalfEdge
-    HalfFace
```
gonzalocasas: why is this one removed from docs?

tomvanmele:
@gonzalocasas ready as well

src/compas/datastructures/mesh/geometry.py
```diff
@@ -14,15 +14,14 @@ def trimesh_face_circle(mesh, fkey):
...
+    fkey : int
         The face key.
 
     Returns
     -------
-    list
-        The centre coordinates, the radius value and the normal vector of the circle.
-    None
-        If the face is not a triangle.
+    Union[tuple[list[float], float, list[float]], None]
```
gonzalocasas: This is one of the few places where the `Union` syntax is used instead of plain English (`tuple[list[float], float, list[float]] or  None`). Perhaps we should use the same everywhere (and I kind of prefer `Union` vs the plain english version)

gonzalocasas (approved):
LGTM!

and with that, I think i've reviewed all of the pending PRs! ;)

PR #965, opened 2022-01-05
Geometry docstrings

tomvanmele:

### Checklist


tomvanmele:
@gonzalocasas ready

NOTES
```diff
@@ -0,0 +1,5 @@
+* polyline divide by length function does not make much sense (and returns an incorrect result)
```
gonzalocasas: is this doc really supposed to be committed?

src/compas/geometry/_core/normals.py
```diff
@@ -76,17 +78,19 @@ def normal_triangle(triangle, unitized=True):
...
     Returns
     -------
-    list
+    [float, float, float]
         The normal vector.
 
     Raises
     ------
-    ValueError
+    AssertionError
```
gonzalocasas, line 93: this is something we might reconsider at some point: `assert` statements are removed and `AssertionError` are not thrown when python is run with optimizations enabled (`python -o`). I haven't checked this in a long while, but as far as I remember, `assert` should only be used in test cases, not in the rest of the source code.

In any case, that change -if we ever do it- is for a different PR.

src/compas/geometry/transformations/translation.py
```diff
@@ -67,6 +73,14 @@ def __init__(self, matrix=None, check=True):
                     raise ValueError('This is not a proper translation matrix.')
         super(Translation, self).__init__(matrix=matrix)
 
+    @property
+    def translation_vector(self):
+        from compas.geometry import Vector
```
gonzalocasas, line 78: why the inlined import?

src/compas/geometry/primitives/frame.py
```diff
@@ -645,29 +690,29 @@ def to_local_coordinates(self, object_in_wcf):
...
-        if isinstance(object_in_wcf, list):
-            return Point(*object_in_wcf).transformed(T)
-        else:
-            return object_in_wcf.transformed(T)
+        if isinstance(obj, (list, tuple)):
+            return Point(*obj).transformed(T)
+        return obj.transformed(T)
 
-    def to_world_coordinates(self, object_in_lcf):
+    def to_world_coordinates(self, obj):
```
gonzalocasas: i would tend to keep the `lcf` (or `wcf` ) references in param names, sometimes this thing is very confusing and it is not bad to have it explicit in the name.

src/compas/geometry/transformations/rotation.py
```diff
@@ -69,25 +81,56 @@ def __init__(self, matrix=None, check=True):
                     raise ValueError('This is not a proper rotation matrix.')
         super(Rotation, self).__init__(matrix=matrix)
 
+    @property
+    def quaternion(self):
+        from compas.geometry import Quaternion
```
gonzalocasas, line 86: same thing here.. why are these imports inlined?

gonzalocasas (approved):
LGTM + comments!

NOTES
tomvanmele: nope will remove

src/compas/geometry/transformations/translation.py
tomvanmele: avoid circular import. the imports of transformations precede the imports of primitives etc.

src/compas/geometry/transformations/rotation.py
tomvanmele: same

tomvanmele: this problem might go away after restructuring but that is for the next set of PRs :)

gonzalocasas: 👍

PR #966, opened 2022-01-06
General geometry collection

tomvanmele:

### Checklist

### DIFF
--- a/src/compas/geometry/collections/collection.py
+++ b/src/compas/geometry/collections/collection.py
@@ -2,15 +2,124 @@ from __future__ import print_function
 from __future__ import absolute_import
 from __future__ import division
 
-__all__ = ['Collection']
+from compas.data import Data
 
 
-class Collection(object):
+class Collection(Data):
 
     __slots__ = ['_items']
 
-    def transform(self, X):
+    def __init__(self, items=None, **kwargs):
+        super(Collection, self).__init__(**kwargs)
+        self._items = []
+        self.items = items
+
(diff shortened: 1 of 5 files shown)


src/compas/geometry/collections/collection.py
```diff
@@ -2,15 +2,240 @@
...
+    # data
+    # ==========================================================================
+
+    @property
+    def DATASCHEMA(self):
+        """schema.Schema : Schema of the data representation."""
+        from schema import Schema
+        return Schema({
+            'itype': '.../...',
+            'items': lambda item: 'validate schema?'
```
gonzalocasas, line 92: This definitely needs an `itype` before merging, as well as something in the `items` stuff.

gonzalocasas: Actually... what is `itype`? :P `item_type` I assume, but it's treated used like the `dtype` of the data serialization framework with the same 2-level contraint, which I'm not sure makes so much sense here.

gonzalocasas (commented):
After going through this, I have one big question: why do we need `Collection` at all? It seems like a very restrictive array-like, which artificially limits items to be of the same type. I don't quite see where we use the class either... is this a leftover of older times?

tomvanmele:
it is a left-over. it could be useful for applying transformations etc to large collections of objects, and potentially in the future in relation to the viewer (it has a collection object for large numbers of small objects to make sure we don't create many small buffers).

but i am happy to remove it

PR #969, opened 2022-01-07
RPC docstrings

tomvanmele:

### Checklist

### DIFF
--- a/src/compas/rpc/dispatcher.py
+++ b/src/compas/rpc/dispatcher.py
@@ -25,15 +25,8 @@ except ImportError:
     from profile import Profile
 
 
-__all__ = ['Dispatcher']
-
-
 class Dispatcher(object):
     """Base class for remote services.
-
-    Examples
-    --------
-    >>>
 
     Notes
     -----
@@ -43,11 +36,24 @@ class Dispatcher(object):
     such that the errors can be rethrown on the client side.
 
     """
+
     def on_module_imported(self, module, newly_loaded_modules):
         """Event triggered when a module is successfully imported.
 
         Override this method when subclassing in order to handle the what happens
         after a module has been imported.
+
+        Parameters
+        ----------
+        module : module
+            A module object.
+        newly_loaded_modules : sequence[str]
+            ???
+
+        Returns
+        -------
+        None
+
         """
         pass
 
--- a/src/compas/rpc/dispatcher.py
+++ b/src/compas/rpc/dispatcher.py
@@ -48,7 +48,7 @@ class Dispatcher(object):
         module : module
             A module object.
         newly_loaded_modules : sequence[str]
-            ???
+            A list of module names that have been loaded.
 
         Returns
         -------
--- a/src/compas/rpc/proxy.py
+++ b/src/compas/rpc/proxy.py
@@ -23,9 +23,6 @@ except ImportError:
     from System.Diagnostics import Process
 
 
-__all__ = ['Proxy']
-
-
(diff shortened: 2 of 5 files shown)


src/compas/rpc/dispatcher.py
```diff
@@ -43,11 +36,24 @@ class Dispatcher(object):
...
 
         Override this method when subclassing in order to handle the what happens
         after a module has been imported.
+
+        Parameters
+        ----------
+        module : module
+            A module object.
+        newly_loaded_modules : sequence[str]
+            ???
```
gonzalocasas: This is a list of module names that have been loaded.

src/compas/rpc/proxy.py
```diff
@@ -37,28 +34,42 @@ class Proxy(object):
...
     ----------
-    package : :obj:`str`:, optional
+    package : str, optional
         The base package for the requested functionality.
-        Default is `None`, in which case a full path to function calls should be provided.
-    python : :obj:`str`:, optional
+        Default is None, in which case a full path to function calls should be provided.
+    python : str, optional
         The python executable that should be used to execute the code.
-        Default is ``'pythonw'``.
```
gonzalocasas, line 45: Is this (and all other "Defaults to") removed because the default is visible directly in the method signature?

src/compas/rpc/proxy.py
```diff
@@ -37,28 +34,42 @@ class Proxy(object):
...
+    autoreload : bool, optional
+        If True, automatically reload the proxied package if changes are detected.
         This is particularly useful during development. The server will monitor changes to the files
         that were loaded as a result of accessing the specified `package` and if any change is detected,
         it will unload the module, so that the next invocation uses a fresh version.
-    capture_output : :obj:`bool`, ``True`` to capture the stdout/stderr output of the remote process, otherwise ``False``.
-        In general, ``capture_output`` should be ``True`` when using a ``pythonw`` as executable (default).
+    capture_output : bool, optional
+        If True, capture the stdout/stderr output of the remote process.
+        In general, `capture_output` should be True when using a `pythonw` as executable (default).
```
gonzalocasas, line 58: I am not convinced about going to non-monospaced for these things.

```suggestion
        In general, ``capture_output`` should be True when using a ``pythonw`` as executable (default).
```

src/compas/rpc/server.py
```diff
@@ -10,12 +10,14 @@
...
-__all__ = ['Server']
-
-
 class Server(SimpleXMLRPCServer):
     """Version of a `SimpleXMLRPCServer` that can be cleanly terminated from the client side.
 
+    Notes
+    -----
+    This class has to be used by a service to start the XMLRPC server in a way
+    that can be pinged to check if the server is live, and can be cleanly terminated.
```
gonzalocasas:
```suggestion
    that can be pinged to check if the server is alive, and can be cleanly terminated.
```

gonzalocasas (approved):
LGTM + minor comments

src/compas/rpc/proxy.py
tomvanmele: yes

src/compas/rpc/proxy.py
tomvanmele: it is based on the numpydoc style guide, but we can change that if you like

gonzalocasas: Not sure... I like the style of monospaced font a bit better, but if numpy does it this way, so be it :P
image

PR #971, opened 2022-01-10
GH Python docstrings

tomvanmele:

### Checklist

### DIFF
--- a/src/compas_ghpython/artists/__init__.py
+++ b/src/compas_ghpython/artists/__init__.py
@@ -5,6 +5,12 @@ artists
 
 .. currentmodule:: compas_ghpython.artists
 
+.. rst-class:: lead
+
+Artists for visualizing (painting) COMPAS geometry, robots, and data structures in Grasshopper using the GH Python component.
+The artists in this package provide plugins for the pluggable methods of the base artists in :mod:`compas.artists`.
+Therefire, they can be used directly, from here, or through the base artists :class:`compas.artists.Artist`.
+
 
 Primitive Artists
 =================
--- a/src/compas_ghpython/artists/__init__.py
+++ b/src/compas_ghpython/artists/__init__.py
@@ -9,7 +9,7 @@ artists
 
 Artists for visualizing (painting) COMPAS geometry, robots, and data structures in Grasshopper using the GH Python component.
 The artists in this package provide plugins for the pluggable methods of the base artists in :mod:`compas.artists`.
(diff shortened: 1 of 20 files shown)


src/compas_ghpython/artists/__init__.py
```diff
@@ -5,6 +5,12 @@
 
 .. currentmodule:: compas_ghpython.artists
 
+.. rst-class:: lead
+
+Artists for visualizing (painting) COMPAS geometry, robots, and data structures in Grasshopper using the GH Python component.
+The artists in this package provide plugins for the pluggable methods of the base artists in :mod:`compas.artists`.
+Therefire, they can be used directly, from here, or through the base artists :class:`compas.artists.Artist`.
```
gonzalocasas:
```suggestion
Therefore, they can be used directly, from here, or through the base artists :class:`compas.artists.Artist`.
```

src/compas_ghpython/artists/robotmodelartist.py
```diff
@@ -36,5 +43,13 @@ def create_geometry(self, geometry, name=None, color=None):
...
+    # should this not have a return value?!
     def draw(self):
+        """Draw the robot model.
+
+        Returns
+        -------
+        None
+
+        """
         self.draw_visual()
```
gonzalocasas: yes, this is a bug. it should return `draw_visual()`  which is a generator of rhino meshes. 
```suggestion
    def draw(self):
        """Draw the visual meshes of the robot model.

        Returns
        -------
        generator[:rhino:`Rhino.Geometry.Mesh`]

        """
        return self.draw_visual()
```

And btw, the `draw_*` methods in the class `compas.artists.RobotModelArtist` are incorrectly documented as returning `None`, when most of them yield rhino meshes.

gonzalocasas: and 2nd btw: is there a convention for how to write generators in the returns? or should it be changed to `sequence[]` instead?

gonzalocasas (approved):
LGTM + minor details

src/compas_ghpython/artists/robotmodelartist.py
tomvanmele: if it yields something it should be a generator, if it an ordered random access collection it should be a sequence. the proper type hint would be `Generator[...]` and `Sequence[...]`, but in my opinion lowercase is fine as well in docstrings

gonzalocasas: 👍  (yep, it's a generator)

PR #973, opened 2022-01-13
Better color handling

tomvanmele:

### Checklist


src/compas/colors/__init__.py
```diff
@@ -0,0 +1,35 @@
...
+
+    Color
+    ColorMap
+
+"""
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+from .color import Color
```
gonzalocasas, line 29: I think also all the color-related utility functions in `compas.utilities` should be moved here (with the corresponding aliasing and deprecation notice)

src/compas/colors/color.py
```diff
@@ -0,0 +1,873 @@
...
+    >>> Color.lime().is_light
+    True
+    >>> Color.navy().is_light
+    False
+
+    """
+
+    def __init__(self, red, green, blue, alpha=1.0, **kwargs):
+        super(Color, self).__init__(**kwargs)
+        self.r = red
```
gonzalocasas, line 97: I would add a validation here, since r,g,b are expected to be within `0.0-1.0` bounds exclusively.

src/compas/colors/color.py
```diff
@@ -0,0 +1,873 @@
...
+    # data
+    # --------------------------------------------------------------------------
+
+    @property
+    def data(self):
+        return {'red': self.r, 'green': self.g, 'blue': self.b, 'alpha': self.a}
+
+    @data.setter
+    def data(self, data):
+        self.r = data['red']
```
gonzalocasas, line 112: also here validation, since this does not trigger the constructor

src/compas/colors/color.py
```diff
@@ -0,0 +1,873 @@
...
+    def lightness(self):
+        return self.hls[1]
+
+    @property
+    def brightness(self):
+        return self.hsv[2]
+
+    @property
+    def is_light(self):
+        return self.luminance > 0.179
```
gonzalocasas, line 201: move the constant to the top of the module and name it

src/compas/colors/color.py
```diff
@@ -0,0 +1,873 @@
...
+        y = self.luma
+        u, v = self.chroma
+        return y, u, v
+
+    @property
+    def luma(self):
+        """float :
+        The brightness of a yuv signal.
+        """
+        return 0.299 * self.r + 0.587 * self.g + 0.114 * self.b
```
gonzalocasas, line 217: I would name and make constants out of all of these magic numbers here and in related functions. Their values might be clear to you, but I have no idea what they stand for except assuming it's some coefficient of conversion.

src/compas/colors/color.py
```diff
@@ -0,0 +1,873 @@
...
+
+        """
+        r = y + 1.140 * v
+        g = y - 0.395 * u - 0.581 * v
+        b = y + 2.032 * u
+        return cls(r, g, b)
+
+    @classmethod
+    def from_i(cls, i):
+        """Construct a color from a single number in the range 0-1.
```
gonzalocasas, line 379: What kind of encoding is this one? Is there a "known" encoding (if so, please add some reference to the docs), or is it a compas-specific encoding? The name is very confusing, `from_i` implies on first read that is from an integer, but `i` is just the name of a variable that happens to be a float.

And tbh, not very convinced about the encoding method at all. It seems pretty arbitrary to me the way the float is interpreted. If the encoding is compas-specific, then why not use the fact that the float is encoded in 32-bits, break it up into 4 bytes and you have rgb+a in a very consistent and compact representation?

src/compas/utilities/colors.py
```diff
@@ -64,6 +64,9 @@
 def i_to_rgb(i, normalize=False):
     """Convert a number between 0.0 and 1.0 to an equivalent RGB tuple.
 
+    .. deprecated:: 1.14
```
gonzalocasas, line 45: I like this deprecation notice in the docs. However, all the other ones that were recently added to the docs were done just programmatically using `warnings` module. Should we do both everywhere?

gonzalocasas (approved):
Generally LGTM + some feedback

src/compas/colors/color.py
tomvanmele: it is compas-specific and needs to be included to provide backward compatible support for all the `i_to_xxx` functions currently in the utilities module. the original objective was to provide a method that can turn values in a normalized range into colors.

src/compas/utilities/colors.py
tomvanmele: i will add both everywhere in one big clean up docstring PR across all modules

PR #975, opened 2022-01-14
Rhino docstrings

tomvanmele:

### Checklist


src/compas_rhino/conduits/base.py
```diff
@@ -25,6 +25,21 @@ def __init__(self, refreshrate=1):
 
     @contextmanager
     def enabled(self):
+        """Create a safe context for the conduit with automatic enabling and disabling.
```
gonzalocasas: `safe` is a bit of an ambiguous term here, since our users might not know or care what are the possible unsafe aspects of a conduit, I would err on the side of being verbose and explain more in detail, e.g. "Create a context for the conduit with automatic enabling and disabling, that ensure all resources required for the it are deallocated properly when the conduit is no longer in use." (or something along those lines)

src/compas_rhino/conduits/base.py
```diff
@@ -34,30 +49,56 @@ def enabled(self):
...
+
+        Returns
+        -------
+        None
+
+        """
         self.Enabled = False
 
-    def redraw(self, k=0, pause=None):
+    def redraw(self, k=0, pause=0.0):
```
gonzalocasas, line 92: if the default changed to `0.0`, I would also change the line 106 below, to be a numeric check: `if pause > 0.0:` (which also hints at the fact we should validate pause not being negative)

src/compas_rhino/objects/__init__.py
```diff
@@ -55,10 +54,19 @@
 from compas.datastructures import Network
 from compas.datastructures import VolMesh
 
+import warnings
+
+warnings.warn(
+    "The objects module in compas_rhino is deprecated. Use the equivalent functionality from compas_ui instead",
```
gonzalocasas: `compas_ui`? That doesn't (yet?) exist, does it? Is it a typo?

gonzalocasas (approved):
LGTM, the only thing I'd definitely do before merge is sort out that confusion with the deprecation in favor of a module that does not yet exist (`compas_ui`)

src/compas_rhino/objects/__init__.py
tomvanmele: it is in the making :)

src/compas_rhino/ui/__init__.py
```diff
@@ -25,6 +25,14 @@
 from .mouse import Mouse
 from .rui import Rui
 
+import warnings
+
+warnings.warn(
```
gonzalocasas, line 30: @brgcode this one got merged and should also be deleted/commented out until `compas_ui` comes to life ;)

tomvanmele: okay, will do

PR #982, opened 2022-02-02
Cross platform use of colors

tomvanmele:

### Checklist


tomvanmele:
The example i use to test is this...
```python
from compas.artists import Artist
from compas.colors import Color

mesh1 = Mesh.from_meshgrid(10, 10)
mesh2 = mesh1.transformed(Translation.from_vector([11, 0, 0]))

mesh1.name = 'M1'
mesh2.name = 'M2'

Artist.clear()

artist1 = Artist(mesh1)

artist1.face_color = {face: Color.pink() for face in mesh1.face_sample(size=17)}

# artist1.draw()
artist1.draw_faces()
# artist1.draw_vertices()
# artist1.draw_vertexlabels()

artist2 = Artist(mesh2)

artist2.default_facecolor = Color.pink().lightened(50)
artist2.face_color = {face: '#0092D2' for face in mesh2.face_sample(size=17)}
artist2.draw_faces()

Artist.redraw()
```

tomvanmele:
result should be the same for Rhino and Blender
Screenshot 2022-02-02 at 08 57 23
Screenshot 2022-02-02 at 08 58 05

src/compas/artists/meshartist.py
```diff
@@ -3,11 +3,69 @@
...
+            item_color = defaultdict(lambda: default)
+
+            for item in value:
+                color = value[item]
+                if Color.is_rgb255(color):
+                    color = Color.from_rgb255(color[0], color[1], color[2])
+                elif Color.is_hex(color):
+                    color = Color.from_hex(color)
+                else:
+                    color = Color(color[0], color[1], color[2])
```
Licini: I would suggest something like 
```python
elif Color.is_rgbf(color):
    color = Color(color[0], color[1], color[2])
else:
    raise ValueError("Non-supported color value")
```

src/compas/artists/meshartist.py
```diff
@@ -3,11 +3,69 @@
...
+                item_color[item] = color
+            setattr(obj, self.private_name, item_color)
+
+        else:
+            if Color.is_rgb255(value):
+                color = Color.from_rgb255(value[0], value[1], value[2])
+            elif Color.is_hex(value):
+                color = Color.from_hex(value)
+            else:
+                color = Color(value[0], value[1], value[2])
```
Licini: Same with above : )

src/compas/artists/meshartist.py
```diff
@@ -79,16 +137,26 @@ class MeshArtist(Artist):
...
-    default_facecolor = (0.9, 0.9, 0.9)
+    color = Color.from_hex('#0092D2').lightened(50)
+
+    default_vertexcolor = Color.from_hex('#0092D2')
+    default_edgecolor = Color.from_hex('#0092D2')
+    default_facecolor = Color.from_hex('#0092D2').lightened(50)
+
+    vertex_color = ColorDict()
+    edge_color = ColorDict()
+    face_color = ColorDict()
```
Licini, line 88: just nitpicking, should it be `vertexcolor` here since it was `default_vertexcolor` above?

src/compas/artists/primitiveartist.py
```diff
@@ -50,10 +50,16 @@ def primitive(self, primitive):
...
-            self._color = self.default_color
+            self.color = self.default_color
         return self._color
 
     @color.setter
-    def color(self, color):
-        if is_color_rgb(color):
-            self._color = color
+    def color(self, c):
+        if not c:
```
Licini: very small thing, maybe `if c is not None`, cuz this is going to raise error when c is an np array

src/compas/artists/shapeartist.py
```diff
@@ -34,7 +34,7 @@ class ShapeArtist(Artist):
 
     """
 
-    default_color = (1, 1, 1)
+    default_color = Color.from_hex('#0092D2').rgb
```
Licini: Wouldn't it be better if the color attributes are actual Color instances than tuples?

Licini: Not sure if this was required by other implementations tho

Licini (commented):
Well overall this is awesome!

src/compas/artists/meshartist.py
Licini: nevermind...it might be too much a breaking change...

tomvanmele: well, i have been using the convention that the underscore form indicates a map between first and second part. so `vertex_color` is a map between `vertex` and `color`. but perhaps i was overthinking this :)

src/compas/artists/shapeartist.py
tomvanmele: yes, i have only done it for mesh for now as a prototype...

tomvanmele: well and a little test here and there :)

src/compas/artists/meshartist.py
```diff
@@ -360,17 +318,17 @@ def draw_faces(self, faces=None, color=None, text=None):
         """
         raise NotImplementedError
 
-    @abstractmethod
-    def draw_mesh(self):
-        """Draw the mesh of the mesh.
+    # @abstractmethod
```
gonzalocasas: Why is this commented out?

src/compas/artists/meshartist.py
```diff
@@ -405,14 +363,14 @@ def clear_faces(self):
         """
         raise NotImplementedError
 
-    def clear(self):
-        """Clear all components of the mesh.
+    # def clear(self):
```
gonzalocasas: Same here.. are they not part of the expected implementation anymore? (although, here I understand it a bit more, because GHPython's artists don't have stuff to clear, but still)

src/compas/artists/colordict.py
```diff
@@ -0,0 +1,59 @@
...
+        self.default_name = 'default_' + ''.join(name.split('_'))
+
+    def __get__(self, obj, otype=None):
+        return getattr(obj, self.private_name) or defaultdict(lambda: getattr(obj, self.default_name))
+
+    def __set__(self, obj, value):
+        if not value:
+            return
+
+        if isinstance(value, dict):
```
gonzalocasas: not totally clear what exactly you are doing in this one, but, shouldn't this be checking for `mapping` or one of those more general interfaces?

src/compas/artists/colordict.py
```diff
@@ -0,0 +1,59 @@
...
+        self.public_name = name
+        self.private_name = '_' + name
+        self.default_name = 'default_' + ''.join(name.split('_'))
+
+    def __get__(self, obj, otype=None):
+        return getattr(obj, self.private_name) or defaultdict(lambda: getattr(obj, self.default_name))
+
+    def __set__(self, obj, value):
+        if not value:
+            return
```
gonzalocasas: Despite being internal, I would document what each of these methods is doing because it's quite cryptic.

src/compas/artists/colordict.py
```diff
@@ -0,0 +1,59 @@
...
+            for item in value:
+                color = value[item]
+                if Color.is_rgb255(color):
+                    color = Color.from_rgb255(color[0], color[1], color[2])
+                elif Color.is_hex(color):
+                    color = Color.from_hex(color)
+                else:
+                    color = Color(color[0], color[1], color[2])
+                item_color[item] = color
+            setattr(obj, self.private_name, item_color)
```
gonzalocasas: this block is duplicated below, maybe move to a private function and call...?

src/compas/artists/shapeartist.py
```diff
@@ -58,13 +58,19 @@ def shape(self, shape):
...
-            self._color = color
+    def color(self, c):
+        if not c:
+            return
+        if Color.is_rgb255(c):
+            self._color = Color.from_rgb255(c[0], c[1], c[2])
+        elif Color.is_hex(c):
+            self._color = Color.from_hex(c)
+        else:
+            self._color = Color(c[0], c[1], c[2])
```
gonzalocasas: this behavior seems to be re-used multiple times, maybe move to a classmethod/staticmethod somewhere else?

src/compas/numerical/linalg.py
```diff
@@ -708,10 +708,7 @@ def spsolve_with_known(A, b, x, known):
     >>> A = array([[2, 1, 3], [2, 6, 8], [6, 8, 18]])
     >>> b = array([[1], [3], [5]])
     >>> x = array([[0.3], [0], [0]])
-    >>> solve_with_known(A, b, x, [0])
-    array([[ 3.00000000e-01],
-           [ 4.00000000e-01],
-           [-6.66247443e-17]])
+    >>> x = solve_with_known(A, b, x, [0])
```
gonzalocasas: I saw this one also failing occasionally. It could also be fixed with `allclose` because it fails due to rounding.

gonzalocasas (commented):
I haven't finished reviewing this, but before I continue: is this for 2.x alpha? Or still aiming at 1.x? I think there are a few breaking changes in there.. or I am confused?

gonzalocasas:
and, this could definitely use some unit testing ;)

src/compas/artists/meshartist.py
tomvanmele: already defined on the base artist. was only reformulated for the mesh because of this awkward definition of `draw`, which i would now replace by what used to be `draw_mesh`, since that is the expected behaviour in almost all cases...

src/compas/artists/meshartist.py
tomvanmele: drawing the mesh should actually be the default behaviour, which can be replaced by combinations of `draw_vertices`, `draw_edges`, `draw_faces` in cases where the visualisation of individual components is needed...

so `draw_mesh` can simply be replaced by `draw`, which then draws the mesh instead of individual vertices, edges, and faces...

src/compas/artists/colordict.py
tomvanmele: yes true

tomvanmele: not sure if that will work in Rhino though

src/compas/numerical/linalg.py
tomvanmele: yes, true, and in any case it shouldn't ne part of this PR...

tomvanmele:
> I haven't finished reviewing this, but before I continue: is this for 2.x alpha? Or still aiming at 1.x? I think there are a few breaking changes in there.. or I am confused?

well it is a bit WIP still so perhaps i should convert it to draft. but ideally this can still be included in the current version since it solves (or should solve) the awkwardness of dealing with colors in Rhino v. Blender...

src/compas/artists/meshartist.py
gonzalocasas: aha! got it now... 👍 sounds perfect then..! it would still make sense to include a deprecated alias `draw_mesh` to not break the significant bunch of code and examples that use it.

gonzalocasas:
> well it is a bit WIP still so perhaps i should convert it to draft. but ideally this can still be included in the current version since it solves (or should solve) the awkwardness of dealing with colors in Rhino v. Blender...

👍  I suspected it was the case, but wasn't sure ;)

tomvanmele:
@gonzalocasas @Licini this is ready now. there are some stubs for testing in `docs/tutorial/artists` which will eventually make it into a proper tutorial...

src/compas/artists/colordict.py
```diff
@@ -0,0 +1,108 @@
...
+
+    def __get__(self, obj, otype=None):
+        """Get the color dict stored in the private attribute corresponding to the public attribute name of the descriptor.
+
+        Parameters
+        ----------
+        obj : object
+            The object owning the instance of the descriptor.
+        otype : ???, optional
+            ???
```
gonzalocasas: missing details ;)

src/compas/artists/colordict.py
gonzalocasas: image

yep, this works on ironpython

src/compas/datastructures/network/network.py
```diff
@@ -209,6 +209,30 @@ def from_nodes_and_edges(cls, nodes, edges):
 
         return network
 
+    @classmethod
+    def from_pointcloud(cls, cloud, degree=3):
+        """Construct a network from random connections between the points of a pointcloud.
```
gonzalocasas, line 214: Interesting method. I think the name does not clearly imply this is a random network, I'm not sure if it's needed to either.  Perhaps it wouldn't be a bad idea to add a way to create a network from point cloud, but not connect it randomly by default. However, I do see that it's also too trivial to justify.

gonzalocasas (approved):
LGTM!

Suggestion for future PRs: when you do sweeping changes using regex, please share the regex used in the PR description. That way a) we could review the origin of the change instead of every instance, and b) we could also apply the same to other compas packages.

tomvanmele:
sorry didn't see the comments. will push directly to main...

PR #987, opened 2022-02-05
Fix doctests

gonzalocasas:
Fixes all doctests (and also a bunch of docstrings) so that we can re-enable running them on CI (see https://github.com/compas-dev/compas-actions.docs/pull/1)

### Checklist


src/compas/data/data.py
```diff
@@ -83,20 +83,20 @@ class Data(object):
...
-    >>> s = json_dumps(a)
-    >>> b = json_loads(s)
-    >>> a is b
+    >>> from compas.data import json_dumps, json_loads        # doctest: +SKIP
+    >>> s = json_dumps(a)                                     # doctest: +SKIP
+    >>> b = json_loads(s)                                     # doctest: +SKIP
+    >>> a is b                                                # doctest: +SKIP
     False
-    >>> a == b
+    >>> a == b                                                # doctest: +SKIP
```
tomvanmele: do these also need to be skipped if we remove the blank line above?

src/compas/data/encoders.py
```diff
@@ -78,7 +79,7 @@ class DataEncoder(json.JSONEncoder):
     >>> from compas.data import json_dump
     >>> from compas.geometry import Point
     >>> point = Point(0, 0, 0)
-    >>> json_dump(point, 'point.json')
+    >>> json_dump(point, compas.get('point.json'))
```
tomvanmele, line 82: not sure why `compas.get` is needed, but if we add it, we should add `import compas` as well...

src/compas/data/encoders.py
```diff
@@ -155,15 +156,16 @@ class DataDecoder(json.JSONDecoder):
...
-    >>> with open('point.json') as f:
+    >>> with open(compas.get('point.json'), 'r') as f:
     ...     point = json.load(f, cls=DataDecoder)
     ...
 
     Implicit use case.
 
     >>> from compas.data import json_load
-    >>> point = json_load('point.json')
+    >>> point = json_load(compas.get('point.json'))
```
tomvanmele, line 168: same here

src/compas/datastructures/graph/graph.py
```diff
@@ -436,7 +436,9 @@ def add_node(self, key=None, attr_dict=None, **kwattr):
         Examples
         --------
         >>> graph = Graph()
-        >>> graph.add_node()
+        >>> key = graph.add_node()
```
tomvanmele: i would prefer not to use `key` anymore, since it is a reference to an implementation detail.
can it not be `>>> node = graph.add_node()`?

tomvanmele: for example, in potential subsequent functions, i find it makes a difference...
```python
x = graph.node_attribute(node, 'x')
```

tomvanmele: btw, would also like to refactor this in the names of function and method params...

src/compas/datastructures/mesh/operations/split.py
```diff
@@ -193,7 +193,7 @@ def mesh_split_face(mesh, fkey, u, v):
     >>> mesh = Mesh.from_obj(compas.get("faces.obj"))
     >>> fkey = mesh.get_any_face()
```
tomvanmele: technically, i guess this should become `face = mesh.face_sample(size=1)[0]`

src/compas/datastructures/mesh/operations/split.py
```diff
@@ -193,7 +193,7 @@ def mesh_split_face(mesh, fkey, u, v):
...
     >>> # u and v defines the new edge after splitting
-    >>> u = mesh.get_any_face_vertex(fkey)
+    >>> u = mesh.face_vertices(fkey)[1]
```
tomvanmele: why not `[0]`?

src/compas/datastructures/mesh/operations/split.py
tomvanmele: also, to make this example more robust we should exclude faces on the boundary

tomvanmele: i guess the sample functions should be updated with an `exclude` parameter.
```python
face = mesh.face_sample(size=1, exclude=mesh.faces_on_boundary())[0]
```

src/compas/geometry/_core/distance.py
```diff
@@ -609,7 +609,7 @@ def closest_points_in_cloud_numpy(points, cloud, threshold=10**7, distances=True
     >>> points = [[0, 0, 0], [1, 0, 0], [1, 1, 0], [0, 1, 0]]
     >>> cloud = [[0, 0, 0], [1, 0, 0], [1, 1, 0], [0, 1, 0]]
     >>> closest_points_in_cloud_numpy(points, cloud, distances=True)
-    (array([0, 1, 2, 3]), array([[0.        , 1.        , 1.41421356, 1.        ],
+    (array([0, 1, 2, 3], dtype=int64), array([[0.        , 1.        , 1.41421356, 1.        ],
```
tomvanmele, line 612: perhaps we could consider rewriting the line above to make things more readable
```python
>>> closest, distances = closest_point_in_cloud_numpy(points, cloud, distances=True)
>>> closest
array([0, 1, 2, 3], dtype=int64)
>>> distances
array([[...]])
```

tomvanmele: note to self: add to pointcloud object :)

src/compas/robots/model/base.py
```diff
@@ -90,7 +90,8 @@ def from_urdf(cls, attributes, elements=None, text=None):
         Examples
         --------
         >>> attributes = {'rpy': '0.0 1.57 0.0', 'xyz': '0.0 0.13 0.0'}
-        >>> FrameProxy.from_urdf(attributes, [], None)
+        >>> f = FrameProxy.from_urdf(attributes, [], None)
+        >>> print(f)
```
tomvanmele: this can't be handled with forwarding the `__repr__`?

tomvanmele (commented):
cool! small things here and there :)

src/compas/data/data.py
gonzalocasas: actually, the only thing that really needs to be skipped is the loading, because it does not find the `class Point` type defined above in the doctest. I'll re-enable the dumping part that works.

src/compas/data/encoders.py
gonzalocasas: since tests are run in shuffled order, the load test below can run before the dump test above, so, I created the `point.json` file in the `data` folder and the `compas.get` just makes sure it's getting that one. the import is right above

src/compas/datastructures/graph/graph.py
gonzalocasas: 👍

src/compas/datastructures/mesh/operations/split.py
gonzalocasas: 👍

src/compas/datastructures/mesh/operations/split.py
gonzalocasas: true, no idea why i used 1, changing to `0`

src/compas/geometry/_core/distance.py
gonzalocasas: agree.. I just wanted to make this one pass ;)

src/compas/robots/model/base.py
gonzalocasas: yep, you're right. fixed

gonzalocasas:
good to go (if tests are all green)! Ideally https://github.com/compas-dev/compas-actions.docs/pull/1 is merged first, but the doctests are green on my machine

tomvanmele (approved):
LGTM

PR #990, opened 2022-02-09
realpath_but_for_real_this_time

gonzalocasas:
As discussed, this fixes the `compas.DATA` and related stuff for dev installs that use symlinks. `compas.DATA` still does not work on a non-dev install because the folder is not even copied on install, but that's a separate issue. As a side effect, this does fix the `compas.__version__` output on dev installs to include the git commit hash.

 
### Checklist

### DIFF
--- a/src/compas/__init__.py
+++ b/src/compas/__init__.py
@@ -46,7 +46,7 @@ __version__ = '1.14.0'
 version = LooseVersion(compas.__version__)
 versionstring = version.vstring.split('-')[0]
 
-HERE = os.path.dirname(__file__)
+HERE = compas._os.realpath(os.path.dirname(__file__))
 """str: Path to the location of the compas package."""
 
 HOME = compas._os.absjoin(HERE, '../..')
--- a/src/compas/_os.py
+++ b/src/compas/_os.py
@@ -287,7 +287,7 @@ def _realpath_ipy_win(path):
         if match_name == dirname:
             return match_link
 
-    raise ValueError('path not found: {}'.format(path))
+    return path
 
 
 def _realpath_ipy_posix(path):
--- a/src/compas/_os.py
+++ b/src/compas/_os.py
@@ -256,12 +256,13 @@ def absjoin(*parts):
 
 
 def realpath(path):
-    """Return the canonical path of the specified filename, eliminating any symbolic links encountered in the path.
+    """Return the canonical path of the specified filename, resolving any symbolic links encountered in the path.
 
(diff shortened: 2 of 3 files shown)


gonzalocasas:
I did not test this on OS X (only did a quick test on `ipy` for Linux which worked, so I assume osx will work as well).

src/compas/__init__.py
```diff
@@ -46,7 +46,7 @@
 version = LooseVersion(compas.__version__)
 versionstring = version.vstring.split('-')[0]
 
-HERE = os.path.dirname(__file__)
+HERE = compas._os.realpath(os.path.dirname(__file__))
```
tomvanmele, line 49: i think we can actually rename `compas._os` to `compas.os`...

tomvanmele: because of the absolute paths, there is no risk of confusion...

src/compas/_os.py
```diff
@@ -252,6 +255,48 @@ def absjoin(*parts):
     return os.path.abspath(os.path.join(*parts))
 
 
+def realpath(path):
+    """Return the canonical path of the specified filename, eliminating any symbolic links encountered in the path.
```
tomvanmele: eliminating => resolving?

src/compas/_os.py
```diff
@@ -252,6 +255,48 @@ def absjoin(*parts):
...
+
+    This function uses Python's stdlib `os.path.realpath` in most cases,
+    except when inside IronPython because (guess what?) it is broken and
+    doesn't really eliminate sym links, so, we fallback to a different
+    way to identifying symlinks in that situation."""
```
tomvanmele: for consistency with the recently cleaned up docstrings (and personal preference :)), i would do
```python
... situation.

"""
```

tomvanmele (approved):
LGTM, with some minor comments...

src/compas/__init__.py
gonzalocasas: While I do agree there's no risk of confusion, I would prefer to keep it underscored to indicate this whole module is not part of the public API.

src/compas/_os.py
gonzalocasas: 👍

src/compas/_os.py
gonzalocasas: 👍

src/compas/__init__.py
gonzalocasas: If you want to go for this renaming to `compas.os`, then let's do it in a different PR. For now, I will go ahead and merge this.

PR #991, opened 2022-02-10
Move data files

gonzalocasas:
Why do we keep struggling about this? Let's move data files into `src/compas/data` and have them shipped for the non-dev installs. This resolves a whole lot of little annoyances and differences between conda/pip installs vs dev installs.

### Checklist


gonzalocasas:
This is to be merged into `realpath` branch (or `main` after that one is merged). That's why the CI tests have not been triggered.

tomvanmele:
sure, but perhaps we should then revise the selection of sample files and give them somewhat more descriptive names. the current set just randomly grew out of stuff i had on my computer at some point in time. especially if we want them to be really useful for examples and test they need to be curated a bit more carefully.

in addition, we then perhaps also need to add a location for slightly larger files. otherwise we will needlessly increase the size of the installable package...

tomvanmele:
also, perhaps we could/should rename the `get` function...

tomvanmele:
other than that, i am all for it!

gonzalocasas:
> especially if we want them to be really useful for examples and test they need to be curated a bit more carefully.

They look relatively decent to me.. (except `butt_model.obj`, really hoping it's a butt joint model... 😛 )


> we then perhaps also need to add a location for slightly larger files

Good point. I realize I forgot to fix the `get` function, because now it would still try to load from `https://github.com/compas-dev/compas/raw/main/data/`, so what I could do is change `get` to 1) first check locally in the new location `src/data/samples`, if not found 2) go to the github repo, but not try to get them from `src/data/samples` of the repo, instead, maybe use `https://github.com/compas-dev/compas/raw/main/samples/` and we add a top-level dir for large(er) files.
Sounds good?



> also, perhaps we could/should rename the `get` function...

I actually got so used to `get` that now it feels odd to rename, but here are some alternatives, let me know if you like any?
- `compas.get_sample_path()`
- `compas.get_example_path()`
- `compas.sample_path()`
- `compas.get_sample()`
- `compas.get_example()`

or something else?

tomvanmele:
> They look relatively decent to me.. (except `butt_model.obj`, really hoping it's a butt joint model... 😛 )

no unfortunately it is just a surface that looks like a but and was used to test bestfit algorithms...

tomvanmele:
> Good point. I realize I forgot to fix the `get` function, because now it would still try to load from `https://github.com/compas-dev/compas/raw/main/data/`, so what I could do is change `get` to 1) first check locally in the new location `src/data/samples`, if not found 2) go to the github repo, but not try to get them from `src/data/samples` of the repo, instead, maybe use `https://github.com/compas-dev/compas/raw/main/samples/` and we add a top-level dir for large(er) files.
> Sounds good?

yup

tomvanmele:
> I actually got so used to `get` that now it feels odd to rename, but here are some alternatives, let me know if you like any?

let's maybe leave that for compas 2.0 indeed :)

gonzalocasas:
@brgcode better now?

tomvanmele (approved):
LGTM

PR #999, opened 2022-03-10
Add support for copy and pickle to artists

tomvanmele:

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Added `compas.datastructures.VolMesh.from_meshgrid`.
 * Added `vertices_where`, `vertices_where_predicate`, `edges_where`, `edges_where_predicate` to `compas.datastructures.HalfFace`.
 * Added `faces_where`, `faces_where_predicate`, `cells_where`, `cells_where_predicate` to `compas.datastructures.HalfFace`.
+* Added `compas.artists.Artist.__init__` to set shared `item` attribute.
+* Added `compas.artists.Artist.__copy__` to support shallow copies.
+* Added `compas.artists.Artist.__deepcopy__` to support deep copies.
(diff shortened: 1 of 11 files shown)


gonzalocasas:
In terms of code, this looks good but i don't understand the context in which this is useful. In principle it even sounds counter intuitive to have serializable components in charge of visualization, and it adds additional requirements to implement new artists, but i guess there's a good use case for it. Could you add a couple of details about the situation in which you use this?

tomvanmele:
we use this for serializing the state of an app, like FormFinder or RV2, such that it can be reloaded at a later point in time. artists and objects can have many settings that are a bit painful to serialize in any other way. the copy and deepcopy interface allow various independent states of live objects to be stored for undo/redo...

we tried to be more precise with the serialization by using a similar infrastructure as for the data classes, but it is difficult to get it right, whereas `pickle` and `shelve` provide this functionality out of the box.

```python
import time
import compas
import compas_rhino
from compas_ui.app import App

compas_rhino.clear()

app = App(name='Test')
app.scene.clear()

mesh = Mesh.from_obj(compas.get('tubemesh.obj'))
mesh.name = 'TubeMesh'

obj = app.scene.add(mesh)
app.scene.update()
app.record()

time.sleep(1)

x, y, z = mesh.centroid()
obj.location = Point(-x, -y, 0)
app.scene.update()
app.record()

time.sleep(1)

app.undo()

app.save()
```

in the above, the scene creates UI objects and artists on the fly, which can have all sorts of settings that are object specific. keeping track of those states is difficult through, for example, the data mechanism...

tomvanmele:
i can add this behaviour in `compas_ui` specifically if that is preferred

Licini (approved):
Looks good to me!

tomvanmele:
gonna try this with a mixin in `compas_ui` first, we can see where to take it from there afterwards...

tomvanmele:
works with a mixin in `compas_ui` so we can use it locally there for now. the other changes are still relevant though...

Issue #1008, opened 2022-03-14
Help user understand cases where plugin is not found


gonzalocasas:
# Feature Request

As a user, I want know what to do when a compas plugin is not found (ie. not installed) so that I can be a happy person.

## Details

The current error message when a plugin is not found is really low level and it doesn't indicate the user how to resolve or why the error is occurring at all. This is the case for remeshing, boolean operation plugins, and also for artists (when scripts are run from the command line).

There are two concrete issues:
 * The message contains a URL that is not resolvable, eg. https://plugins.compas.dev/trimesh/trimesh_remesh
 * The English error message doesn't provide guidance on how to resolve it.

Consequently, there are two actions needed:
* The idea of the URL was to point to a webpage that would show specific info about this pluggable. This needs to be configured in our DNS and we need to prepare a webpage as target for it. I am not sure if the current sphinx page will do (maybe it does).
* The error messages should be more helpful, telling user what are their options (like what happens when you type a command that is not installed on linux and `apt` will suggest immediately `Maybe you need to: sudo apt install whatever`

PR #1009, opened 2022-03-15
Improve error messages for artist creation

gonzalocasas:
Tiny PR, partially addresses #1008 but only for a very specific sub-set of the cases (the `Artist`).

### Checklist

### DIFF
--- a/src/compas/artists/artist.py
+++ b/src/compas/artists/artist.py
@@ -45,6 +45,13 @@ def _get_artist_cls(data, **kwargs):
     else:
         Artist.CONTEXT = identify_context()
 
+    if Artist.CONTEXT is None:
+        error_message = 'No context defined.'
+        error_message += '\n\nThis usually means that the script that you are running requires'
+        error_message += '\na CAD environment but it is being ran as a standalone script'
+        error_message += '\n(ie. from the command line or code editor).'
+        raise Exception(error_message)
+
     dtype = type(data)
     cls = None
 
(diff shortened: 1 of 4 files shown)


src/compas/artists/artist.py
```diff
@@ -45,6 +45,13 @@ def _get_artist_cls(data, **kwargs):
     else:
         Artist.CONTEXT = identify_context()
 
+    if Artist.CONTEXT is None:
+        error_message = 'No context defined.'
+        error_message += '\n\nThis usually means that the script that you are running requires'
+        error_message += '\na CAD environment but it is being ran as a standalone script'
+        error_message += '\n(ie. from the command line or code editor).'
+        raise Exception(error_message)
```
tomvanmele: defining custom exception would make the formatting of the message perhaps easier/more elegant? for example `NoArtistContextError`

gonzalocasas: good point, like this? https://github.com/compas-dev/compas/pull/1009/commits/0f9b3ff34bf22b80984eeeaea05bb41f5ec01538

the other one is really a `ValueError`, right? Or do you suggest to also create a custom exception for that?

src/compas/artists/artist.py
```diff
@@ -89,6 +93,10 @@ def __new__(cls, *args, **kwargs):
         if not Artist.__ARTISTS_REGISTERED:
             register_artists()
             Artist.__ARTISTS_REGISTERED = True
+
+        if not args or len(args) < 1 or args[0] is None:
```
tomvanmele: don't think the length check is needed (already covered by `not args`).
we could also structure this differently:

1. `def __new__(cls, item, *args, **kwargs)` since `item` is mandatory
2. forget about `if not args`
3. add `if item is None` to raise ValueError

gonzalocasas: yep, true: fixed here https://github.com/compas-dev/compas/pull/1009/commits/813c4674a138f31d8e87094cef1da9ca798747c5

tomvanmele (approved):

PR #1010, opened 2022-03-16
fix: exception calling `geometry` of compas.datastructures.Part`

gonzalocasas:
If the part has no features assigned, the call to `.geometry` causes an exception, but additionally, it's very inefficient to do all those transformations from shape to mesh back to polyhedron which, in the case of no attached features to be boolean'ed into the geometry, is not needed.

### Checklist

### DIFF
--- a/src/compas/datastructures/assembly/part.py
+++ b/src/compas/datastructures/assembly/part.py
@@ -144,13 +144,17 @@ class Part(Datastructure):
     def geometry(self):
         # TODO: this is a temp solution
         # TODO: add memoization or some other kind of caching
-        A = Mesh.from_shape(self.shape)
-        for shape, operation in self.features:
-            A.quads_to_triangles()
-            B = Mesh.from_shape(shape)
-            B.quads_to_triangles()
-            A = Part.operations[operation](A.to_vertices_and_faces(), B.to_vertices_and_faces())
-        geometry = Shape(*A)
+        if self.features:
+            A = Mesh.from_shape(self.shape)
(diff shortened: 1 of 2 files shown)


src/compas/datastructures/assembly/part.py
```diff
@@ -144,13 +144,17 @@ def frame(self, frame):
...
-            A = Part.operations[operation](A.to_vertices_and_faces(), B.to_vertices_and_faces())
-        geometry = Shape(*A)
+        if self.features:
+            A = Mesh.from_shape(self.shape)
+            for shape, operation in self.features:
+                A.quads_to_triangles()
+                B = Mesh.from_shape(shape)
+                B.quads_to_triangles()
+                A = Part.operations[operation](A.to_vertices_and_faces(), B.to_vertices_and_faces())
+            geometry = Shape(*A)
```
tomvanmele: i think this can be further reduced since geometrical shapes can be converted to triangulated polyhedrons directly.

```python
A = self.shape.to_vertices_and_faces(triangulated=True)
for shape in operation in self.features:
    B = shape.to_vertices_and_faces(triangulated=True)
    A = Part.operations[operation](A, B)  # this is always a triangulation...
geometry = Shape(*A)
```

gonzalocasas: done! https://github.com/compas-dev/compas/pull/1010/commits/510bb4362766a0ef84479e9c9702f09438a8a200

tomvanmele (approved):

PR #1015, opened 2022-03-25
Extend shorten polyline2

mattiskoh:
I took the comments from https://github.com/compas-dev/compas/pull/884 and re-did the implementation.

### Checklist


CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+* Added `Polyline.extend` and `Polyline.shorten`.
 ### Added
 
 * Added optional `triangulated` flag to `Mesh.to_vertices_and_faces`.
```
gonzalocasas:
```suggestion
### Added

* Added `Polyline.extend` and `Polyline.shorten`.
* Added optional `triangulated` flag to `Mesh.to_vertices_and_faces`.
```

gonzalocasas: Moved under the correct headline.

src/compas/geometry/primitives/polyline.py
```diff
@@ -413,3 +413,68 @@ def divide_polyline_by_length(self, length, strict=True, tol=1e-06):
...
+        -------
+        :class:`~compas.geometry.Polyline`
+
+        """
+        try:
+            start, end = length
+            self.points[0] = self.points[0] + self.lines[0].vector.unitized().scaled(-start)
+        except TypeError:
+            start = end = length
+        self.points[-1] = self.points[-1] + self.lines[-1].vector.unitized().scaled(end)
```
gonzalocasas, line 436: I think you could move this line down (outside the `try` block), right?
```suggestion
        except TypeError:
            start = end = length
        self.points[0] = self.points[0] + self.lines[0].vector.unitized().scaled(-start)
        self.points[-1] = self.points[-1] + self.lines[-1].vector.unitized().scaled(end)
```

gonzalocasas (approved):
generally looks good to me! I added a couple of small suggestions only

src/compas/geometry/primitives/polyline.py
mattiskoh: I kept in inside on purpose. My reasoning being: The standard behavior of the function is to only extend the line from the end this was also taking Tom's comments into account. So by keeping it within the `try` gives us option to also extend from the beginning. In the rare case that you would only like to extend from the beginning you can simply input `(length, 0)` into the function. Extending both start and end `(length0, length1)` and extending just the end would be either `length` or `(0, length)`

src/compas/geometry/primitives/polyline.py
```diff
@@ -413,3 +413,68 @@ def divide_polyline_by_length(self, length, strict=True, tol=1e-06):
...
+        """Extends a polyline by a given length, by moving the first and/or last point tangentially.
+
+        Parameters:
+        -----------
+        length: float or tuple[float, float]
+            end_extension or tuple[start_extension, end_extension]
+
+        Returns
+        -------
+        :class:`~compas.geometry.Polyline`
```
tomvanmele: the return value should be `None` since "extend" implies the polyline is modified in-place. "extended" would return a new polyline. probably both methods should be added. see, for example, "transform" vs. "transformed"...

src/compas/geometry/primitives/polyline.py
```diff
@@ -413,3 +413,68 @@ def divide_polyline_by_length(self, length, strict=True, tol=1e-06):
...
 
     divide_by_length = divide_polyline_by_length
+
+    def extend(self, length):
+        """Extends a polyline by a given length, by moving the first and/or last point tangentially.
+
+        Parameters:
+        -----------
+        length: float or tuple[float, float]
+            end_extension or tuple[start_extension, end_extension]
```
tomvanmele:
```
length : float | tuple[float, float]
    A single length value to extend the polyline only at the end,
    or two length values to extend at both ends.
```

src/compas/geometry/primitives/polyline.py
```diff
@@ -413,3 +413,68 @@ def divide_polyline_by_length(self, length, strict=True, tol=1e-06):
...
+        """Shortens a polyline by a given length.
+
+        Parameters:
+        -----------
+        length: float or tuple[float, float]
+            end_shorten or tuple[start_shorten, end_shorten]
+
+        Returns
+        -------
+        :class:`~compas.geometry.Polyline`
```
tomvanmele: same here:
"shorten" => no return value, the polyline is modified in-place
"shortened" => return a new polyline

src/compas/geometry/primitives/polyline.py
```diff
@@ -413,3 +413,68 @@ def divide_polyline_by_length(self, length, strict=True, tol=1e-06):
...
+
+        return self
+
+    def shorten(self, length):
+        """Shortens a polyline by a given length.
+
+        Parameters:
+        -----------
+        length: float or tuple[float, float]
+            end_shorten or tuple[start_shorten, end_shorten]
```
tomvanmele: same comment as above

src/compas/geometry/primitives/polyline.py
```diff
@@ -413,3 +413,68 @@ def divide_polyline_by_length(self, length, strict=True, tol=1e-06):
...
+                    self.points.pop(0)
+                    break
+                else:
+                    self.points[0] = line.end + line.vector.unitized().scaled(-(total_length-start))
+                    break
+        except TypeError:
+            start = end = length
+
+        total_length = 0
+        for line in reversed(self.lines):
```
tomvanmele: this makes a copy of the lines, which seems overkill. how about
```python
for i in range(len(self.lines)):
    line = self.lines[-(i+1)]
```

src/compas/geometry/primitives/polyline.py
```diff
@@ -413,3 +413,68 @@ def divide_polyline_by_length(self, length, strict=True, tol=1e-06):
...
+            total_length += line.length
+            if total_length < end:
+                self.points.pop(-1)
```
tomvanmele: since you are not using the popped point, this would be
```python
del self.points[-1]
```

src/compas/geometry/primitives/polyline.py
```diff
@@ -413,3 +413,68 @@ def divide_polyline_by_length(self, length, strict=True, tol=1e-06):
...
+            elif total_length == end:
+                self.points.pop(-1)
```
tomvanmele: same

src/compas/geometry/primitives/polyline.py
```diff
@@ -413,3 +413,104 @@ def divide_polyline_by_length(self, length, strict=True, tol=1e-06):
...
+
+        Parameters:
+        -----------
+        length: float or tuple[float, float]
+            A single length value to extend the polyline only at the end,
+            or two length values to extend at both ends.
+
+        Returns
+        -------
+        :class:`~compas.geometry.Polyline`
```
tomvanmele: None

src/compas/geometry/primitives/polyline.py
```diff
@@ -413,3 +413,104 @@ def divide_polyline_by_length(self, length, strict=True, tol=1e-06):
...
+        try:
+            start, end = length
+            self.points[0] = self.points[0] + self.lines[0].vector.unitized().scaled(-start)
+        except TypeError:
+            start = end = length
+        self.points[-1] = self.points[-1] + self.lines[-1].vector.unitized().scaled(end)
+
+    def extended(self, length):
+        """Extends a polyline by a given length, by modifying the first and/or last point tangentially.
+        Returns a copy of the original polyline.
```
tomvanmele: Only the first line is recognized as the short summary. Therefore this should be something like "Returns a copy of this polyline extended by a given length."

src/compas/geometry/primitives/polyline.py
```diff
@@ -413,3 +413,104 @@ def divide_polyline_by_length(self, length, strict=True, tol=1e-06):
...
+                del self.points[-1]
+            elif total_length == end:
+                del self.points[-1]
+                break
+            else:
+                self.points[-1] = line.start + line.vector.unitized().scaled(total_length-end)
+                break
+
+    def shortened(self, length):
+        """Shortens a polyline by a given length. Returns a copy of the original Polyline.
```
tomvanmele: Returns a copy of this polyline shortened by a given length.

tomvanmele (commented):
almost there...

tomvanmele (approved):

mattiskoh:
No clue why it's not passing the checks now..

tomvanmele:
there is something wrong with the workflows. it is happening everywhere...

PR #1020, opened 2022-03-29
* Set `jinja >= 3.0` to dev dependencies to fix docs build error.

Licini:
Fixing this: https://github.com/compas-dev/compas/runs/5721182179?check_suite_focus=true


gonzalocasas (approved):
LGTM!

PR #1021, opened 2022-03-29
Data hash value

tomvanmele:

### Checklist

### DIFF
--- a/src/compas/data/data.py
+++ b/src/compas/data/data.py
@@ -4,6 +4,7 @@ from __future__ import division
 
 import os
 import json
+import hashlib
 from uuid import uuid4
 from copy import deepcopy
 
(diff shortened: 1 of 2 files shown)


src/compas/data/data.py
```diff
@@ -112,12 +113,12 @@ def __init__(self, name=None):
 
     def __getstate__(self):
         """Return the object data for state serialization with older pickle protocols."""
-        return {'__dict__': self.__dict__, 'dtype': self.dtype, 'data': self.data}
+        return {"__dict__": self.__dict__, "dtype": self.dtype, "data": self.data}
```
gonzalocasas, line 116: 🖤 black! yay!

src/compas/data/data.py
```diff
@@ -350,3 +360,34 @@ def validate_json(self):
             print("Validation against the JSON schema of this object failed.")
             raise e
         return jsonstring
+
+    def hash(self, as_string=False):
```
gonzalocasas: This is too close to the built-in `__hash__` magic method, but has different semantics, I would suggest a different name, maybe `sha256hash` or `unique_hash`, or `fingerprint`, or something else

gonzalocasas (changes requested):
the idea looks great to me, I would only change the method name to avoid confusion with the built-in

src/compas/data/data.py
Licini: I vote for `sha256` or `fingerprint`

gonzalocasas (approved):
LGTM

PR #1023, opened 2022-03-31
Add optional search path to proxy

tomvanmele:

### Checklist

### DIFF
--- a/src/compas/rpc/dispatcher.py
+++ b/src/compas/rpc/dispatcher.py
@@ -86,13 +86,12 @@ class Dispatcher(object):
             * `'profile'` : A profile of the function execution.
 
         """
-        odict = {
-            'data': None,
-            'error': None,
-            'profile': None
-        }
+        odict = {"data": None, "error": None, "profile": None}
 
-        parts = name.split('.')
+        if len(args) > 1:
+            sys.path.insert(0, args[1])
+
+        parts = name.split(".")
 
         functionname = parts[-1]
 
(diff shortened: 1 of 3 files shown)


tomvanmele:
in file `/Users/vanmelet/Desktop/customrpc.py` define function `mycoolstuff`

```python
def mycoolstuff():
    return "woohoo"
``` 

in caller do

```python
from compas.rpc import Proxy

proxy = Proxy("customrpc", path="/Users/vanmelet/Desktop")

result = proxy.mycoolstuff()
print(result)
```

tomvanmele:
maybe we should also add the option to select a custom environment...

tomvanmele:
would like to rethink the need for a bootstrapper...

src/compas/rpc/dispatcher.py
```diff
@@ -86,13 +86,12 @@ def _dispatch(self, name, args):
...
-        odict = {
-            'data': None,
-            'error': None,
-            'profile': None
-        }
+        odict = {"data": None, "error": None, "profile": None}
 
-        parts = name.split('.')
+        if len(args) > 1:
+            sys.path.insert(0, args[1])
```
gonzalocasas: Add check otherwise it will insert the path on every function call:
```suggestion
            if args[1] not in sys.path:
                sys.path.insert(0, args[1])
```

gonzalocasas (approved):
Besides that suggestion I added, it LGTM! Works on Windows 👍

PR #1032, opened 2022-04-06
Small fixes/improvements to Assembly

gonzalocasas:
A couple of fixes/improvements to `Assembly`:
* Fixed the `find` method which relied on the `self._parts` dictionary to map guids to parts, however, that dictionary was not restored after deserialization. 
* Added a matching `find_by_key` to `Assembly` because it was not possible to lookup parts by their keys.

### Checklist

### DIFF
--- a/src/compas/datastructures/assembly/assembly.py
+++ b/src/compas/datastructures/assembly/assembly.py
@@ -30,9 +30,10 @@ class Assembly(Datastructure):
 
     def __init__(self, name=None, **kwargs):
         super(Assembly, self).__init__()
-        self.attributes = {"name": name or "Assembly", "guid_key_map": {}}
+        self.attributes = {"name": name or "Assembly"}
         self.attributes.update(kwargs)
         self.graph = Graph()
+        self._parts = {}
 
     # ==========================================================================
     # data
@@ -65,6 +66,7 @@ class Assembly(Datastructure):
     def data(self, data):
         self.attributes.update(data["attributes"] or {})
         self.graph.data = data["graph"]
+        self._parts = {part.guid: part.key for part in self.parts()}
 
     # ==========================================================================
     # properties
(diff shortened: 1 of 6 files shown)


src/compas/datastructures/assembly/assembly.py
```diff
@@ -30,10 +30,9 @@ class Assembly(Datastructure):
 
     def __init__(self, name=None, **kwargs):
         super(Assembly, self).__init__()
-        self.attributes = {'name': name or 'Assembly'}
+        self.attributes = {"name": name or "Assembly", "guid_key_map": {}}
```
tomvanmele: are you adding this here because of serialization?
if so, it will not help, because, currently, the guids are not part of the serialized data of a data object...

in addition, i find it a bit weird to "hide" such a key component of the mapping infrastructure in the "general attributes" dict

gonzalocasas: you're right about hiding it, I will rework this, but I'm confused about the guids not being part of serialized objects: does that mean as a general rule that guids are only meaningful while in-memory but are not intended to be stable across serialization?

tomvanmele: yes that is currently unfortunately indeed the case

tomvanmele: we could add it to the serialized version of the object, like we do with `dtype`, which is also not part of the data dict...

tomvanmele: it would solve a lot of issues we are having with the serialization of state and pull/push to the cloud (speckle)

gonzalocasas: addressed the issue at https://github.com/compas-dev/compas/pull/1032/commits/b589f0ef4e0f7a978e9c49fba49feeb75a53a31e 
now it does not depend on guid serialization anymore and mapping dict is rebuilt on deserialization but not in the custom attrs

tomvanmele (approved):
LGTM

PR #1048, opened 2022-05-31
Fix/uninstall ghplugin

arpastrana:

### Checklist

### What are the issues?

The source directory path in `compas_ghpython.uninstall` to uninstall the COMPAS gh plugin was missing the "ghuser"  subfolder. Thus, there is a mismatch with the install path specified by `compas_ghpython.install`.

Most relevantly, `compas_ghpython.components.uninstall_userobjects` internally discarded the input list with the names of the .ghuser objects to uninstall. 

This silent bug did not cause any apparent trouble with the installation of the main library since the names of the COMPAS gh components were prefixed with "compas".  However, this bug makes the gh components of other COMPAS extensions (e.g. COMPAS CEM) to not be uninstalled if they happen to follow a different naming convention (e.g. CEM_AddTrailEdge).

As a result, I propose to gather all the gh components whose names start with "compas" during the uninstall as the fallback option, not the main one, particularly since we are already searching for all the gh components names --prefixed by "compas" or not-- in line 18 of `compas_ghpython.uninstall`.

### DIFF
--- a/src/compas_ghpython/uninstall.py
+++ b/src/compas_ghpython/uninstall.py
@@ -14,7 +14,7 @@ def after_rhino_uninstall(uninstalled_packages):
     if 'compas_ghpython' not in uninstalled_packages:
         return []
 
-    srcdir = os.path.join(os.path.dirname(__file__), 'components')
+    srcdir = os.path.join(os.path.dirname(__file__), 'components', 'ghuser')
     userobjects = [os.path.basename(ghuser) for ghuser in glob.glob(os.path.join(srcdir, '*.ghuser'))]
     uninstalled_objects = uninstall_userobjects(userobjects)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Moved from `autopep8` to `black`
 * Fixed bug in `compas.utilities.linspace` for number series with high precision start and stop values.
 * Fixed uncentered viewbox in `Plotter.zoom_extents()`
+* Fixed source directory path in `compas_ghpython.uninstall` plugin.
 
 ### Removed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Fixed bug in `compas.utilities.linspace` for number series with high precision start and stop values.
 * Fixed uncentered viewbox in `Plotter.zoom_extents()`
 * Fixed source directory path in `compas_ghpython.uninstall` plugin.
+* Fixed bug in`compas_ghpython.components`that ignored input list of `.ghuser` objects to uninstall.
 
 ### Removed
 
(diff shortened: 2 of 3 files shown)


src/compas_ghpython/uninstall.py
```diff
@@ -14,7 +14,7 @@ def after_rhino_uninstall(uninstalled_packages):
     if 'compas_ghpython' not in uninstalled_packages:
         return []
 
-    srcdir = os.path.join(os.path.dirname(__file__), 'components')
+    srcdir = os.path.join(os.path.dirname(__file__), 'components', 'ghuser')
```
gonzalocasas, line 17: The rest of the PR looks great, but list line I'm not sure. In which cases does it currently fail? I did a check test with a dev install (ie. from source) and one with a conda install, and both seem to uninstall properly without this change. However, we did have this bug in COMPAS FAB recently (https://github.com/compas-dev/compas_fab/pull/350/files) so, I'm wondering if this will not actually break stuff on core in an attempt to fix it for extension packages?

gonzalocasas (commented):
Most things look good, but see below for a big doubt I have on the change of path for uninstall.

src/compas_ghpython/uninstall.py
arpastrana: It does not fail for COMPAS core because of the "fallback" mechanism in lines 100-101 of `compas_ghpython.components.uninstall_userobjects()`, which searches for .ghuser components whose names start with "compas":

```python
# compas_ghpython.components.__init__.py

userobjects = []   # at the moment, this silently overwrites the userobjects in the function's signature
for name in os.listdir(dstdir):
      if name.lower().startswith('compas'):
      userobjects.append(name)
```

The fallback makes sense because `userobjects` below is currently an empty list. Why? Because the core .ghuser components are installed by `compas_ghpython.install_userobjects` in `os.path.dirname(__file__)/components/ghuser` not in `os.path.dirname(__file__)/components`.
If you install the COMPAS core GH components --either from a source or from a conda installation--, and then uninstall them, the following check fails.

```python
# compas_ghpython.uninstall.py

srcdir = os.path.join(os.path.dirname(__file__), 'components')
userobjects = [os.path.basename(ghuser) for ghuser in glob.glob(os.path.join(srcdir, '*.ghuser'))]
assert len(userobjects) > 0   # it assumes ghuser components were first installed 
uninstalled_objects = uninstall_userobjects(userobjects)
```

If we change the path for uninstall as proposed in this PR, the check passes, and the COMPAS core and extensions packages ghuser components are uninstalled without breaking :)

arpastrana:
Thanks, @gonzalocasas! I have explained above why I think changing the path to uninstall the components is a good idea.

src/compas_ghpython/uninstall.py
arpastrana: What do you think, @gonzalocasas? Reasonable or should I roll back this one line to help merging this PR? 🙂

tomvanmele: think he is busy with a workshop...

arpastrana: @brgcode Thanks for the heads up!

gonzalocasas (approved):
sorry the delay, I agree with the comments, let's go ahead and merge!

PR #1050, opened 2022-06-03
Lock sphinx version

Licini:
The latest sphinx 5.0 released few days ago is breaking the search box on our docs site. Let's lock it to 4.5 for now until they get it fixed (or until we figure out what we need to change to work with latest version).


gonzalocasas (approved):
LGTM!

PR #1052, opened 2022-06-09
Rhino box bug

chenkasirer:
This is to address two issues related to conversion to/from RhinoBox.

**CompasBox => RhinoBox** 
Was missing the offsetting of the frame point to account for the difference in frame location in box in COMPAS and Rhino (center vs. corner)

**RhinoBox => CompasBox**
Was missing scaling the offset values by the axes, causing a discrepancy when converting a transformed (rotated) Box. 

This might be considered a breaking change since client code compensating for this will now behave differently. But open for debate.

- [x] Breaking change: bug fix or new feature that involve incompatible API changes. (?)

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Fixed uncentered viewbox in `Plotter.zoom_extents()`
 * Changed `RobotModelArtists.atteched_tool_models` to dictionary to support multiple tools.
 * Locked `sphinx` to 4.5.
+* Fixed conversion bug of transformed box in `compas_rhino.conversions._shapes`
 
 ### Removed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,7 +31,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Rebuild part index after deserialization in `Assembly`.
 * Fixed bug in `compas.artists.colordict.ColorDict`.
 * Change `Mesh.mesh_dual` with option of including the boundary. 
-* Fixed type error in `compas_rhino.conversions._shapes.box_to_rhino`.
+* Fixed type error in `compas_rhino.conversions.box_to_rhino`.
 * Moved from `autopep8` to `black`
 * Fixed bug in `compas.utilities.linspace` for number series with high precision start and stop values.
 * Fixed uncentered viewbox in `Plotter.zoom_extents()`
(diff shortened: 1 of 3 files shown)


tomvanmele:
this is not really related to the content of this PR, but we should find a way to test this kind of thing as part of the automated testing framework. since `rhino3dm` is installable from PyPi (https://pypi.org/project/rhino3dm/), perhaps we can define some sample data that can be used for this?

chenkasirer:
> this is not really related to the content of this PR, but we should find a way to test this kind of thing as part of the automated testing framework. since `rhino3dm` is installable from PyPi (https://pypi.org/project/rhino3dm/), perhaps we can define some sample data that can be used for this?

I agree. Wasn't aware of `rhino3dm`, pretty cool. I'll look into it and try to add some tests on this PR.

chenkasirer:
So this is trickier than I imagined. I could mock `Rhino.Geometry` and get it to import some mock package which actually just wraps `rhino3dm`, but the types in `rhion3dm` don't expose the full API available for the types inside Rhino. For example, the constructor used in `compas_rhino.conversions` to create a Rhino `Box` is commented out in the binding source of `rhino3dm`.
```cpp
# bnd_box.h

  BND_Box(const class BND_BoundingBox& bbox);
  //BND_Box(Plane basePlane, Interval xSize, Interval ySize, Interval zSize)
  //BND_Box(Plane basePlane, IEnumerable<Point3d> points)
  //BND_Box(Plane basePlane, GeometryBase geometry)
  //BND_Box(Plane basePlane, BoundingBox boundingbox)
```
Maybe this is due to lack of whatever dark magic layer which wraps around constructor overloading. But nor sure what can be done about it. We could maybe compile our own shared library but that seems like a lot of work, and will not be available on the test pipeline. @gonzalocasas @brgcode what do you think?

tomvanmele:
not sure we talking about the same thing. what i had in mind is that we export some geometries from Rhino as 3dm files and add them to the collection test samples that can be read during the testing process using `rhino3dm`, but perhaps that is not possible...

CHANGELOG.md
```diff
@@ -37,6 +37,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Fixed uncentered viewbox in `Plotter.zoom_extents()`
 * Changed `RobotModelArtists.atteched_tool_models` to dictionary to support multiple tools.
 * Locked `sphinx` to 4.5.
+* Fixed conversion bug of transformed box in `compas_rhino.conversions._shapes`
```
gonzalocasas: A technicality, but the public API is only up to the second level imports, so, users don't know anything about `compas_rhino.conversions._shapes`, we should simply refer to `compas_rhino.conversions` in this case.

gonzalocasas (approved):
LGTM!

About the unit testing conversation, let's keep it separate from this PR, maybe move to a github issue or discussion.

PR #1057, opened 2022-07-13
Geometry brep

chenkasirer:
WIP: added initial implementation of a `compas.geometry.BRep` with a Rhino backend in `compas_rhino.geometry.brep`.

1. Copied over the interface for BRep from `compas_occ` into `compas.geometry.BRep` and generalized.
2. Started implementing the package `compas_rhino.geometry.brep` with wrapping types for brep, edge, vertex and loop.
3. Implemented the serializing part of `data` (in some also the setter but not tested)

I could serialize a simple box brep with:
```python
from pprint import pprint
from compas.geometry import BRep
from compas.data import json_dumps

box = Box.from_width_height_depth(5., 5., 10.)
brep = BRep.from_box(box)

pprint(json_dumps(brep))
```

Opening this draft PR to start a conversation about the approach.

### Checklist


src/compas/geometry/brep/brep.py
```diff
@@ -0,0 +1,410 @@
...
+    frame : :class:`~compas.geometry.Frame`, read-only
+        The local coordinate system of the BRep.
+    area : float, read-only
+        The surface area of the BRep.
+    volume : float, read-only
+        The volume of the regions contained by the BRep.
+
+    Other Attributes
+    ----------------
+    occ_shape : ``TopoDS_Shape``
```
gonzalocasas: this one probably doesn't belong here

src/compas/geometry/brep/brep.py
```diff
@@ -0,0 +1,410 @@
...
+        The underlying OCC shape of the BRep.
+    type : {TopAbs_COMPOUND, TopAbs_COMPSOLID, TopAbs_SOLID, TopAbs_SHELL, TopAbs_FACE, TopAbs_WIRE, TopAbs_EDGE, TopAbs_VERTEX, TopAbs_SHAPE}, read-only
+        The type of BRep shape.
+    orientation : {TopAbs_FORWARD, TopAbs_REVERSED, TopAbs_INTERNAL, TopAbs_EXTERNAL}, read-only
```
gonzalocasas: also this constants seem to be `occ` specific, right?

src/compas/geometry/brep/brep.py
```diff
@@ -0,0 +1,410 @@
...
+        raise NotImplementedError
+
+    # ==============================================================================
+    # Converters
+    # ==============================================================================
+
+    def to_json(self, filepath):
+        raise NotImplementedError
+
+    def to_step(self, filepath, schema="AP203", unit="MM"):
```
gonzalocasas: `"AP203"` and `"MM"` should go somewhere up as constants. Is it ok / consistent with the rest of COMPAS to default to mm units instead of meters?

src/compas/geometry/brep/brep.py
chenkasirer: Right. It's copied over from OCC. Just wanted to see if that's really what we discussed before I re-write all the docstrings.

jf---: tricky one, this would make pythonocc a core dependency, making a massive CAD kernel a core dependency...
its a nice contribution ( have devoted a **lot** of time and energy on pythonocc ) but better make the dependency explicit and have this kind of functionality baked in `compas_occ`?

gonzalocasas: 👍  I guess the contained shape attribute is not part of the interface and instead is something specific to each implementation

tomvanmele: well, an underlying native representation will be present in both cases. in case of nurbs, this is currently referred to as, for example, `occ_curve` and `rhino_curve`, which turns out to be a bit annoying. perhaps better would be to have both refer to an underlying `_curve`.

same could be said about the breps...

src/compas/geometry/brep/brep.py
tomvanmele: yes they are. we could wrap this and make it more general so it applies to both/all backends...

src/compas/geometry/brep/brep.py
tomvanmele: should be removed anyway, because in my experience, using these options results in failed step conversions. their usage is commented out in the occ implementation, but i had not yet removed the params...

src/compas/geometry/brep/brep.py
jf---: @brgcode so from the import statement it would be obvious whether the `brep` backend is `occ` || `rhino`?
cool & ambitious to get `brep` support 🦾💎

tomvanmele: no this would work in the same way as with the nurbs plugins. the constructor methods will hand off the actual construction process to a pluggable. the pluggable will receive plugins from `compas_rhino` and `compas_occ`. depending on the execution context, the plugin manager will then decide which plugin to use...

src/compas/geometry/brep/__init__.py
```diff
@@ -0,0 +1,3 @@
+from .brep import BRep
+
+__all__ = [_ for _ in dir() if not _.startswith("_")]
```
tomvanmele: need to make these explicit for typing infrastructure...

src/compas_rhino/geometry/brep/__init__.py
```diff
@@ -0,0 +1,13 @@
+from compas.plugins import plugin
+
+from .brep import RhinoBRep
+
+
+@plugin(category="factories", requires=["Rhino"])
+def new_brep_from_shape(cls, *args, **kwargs):
+    return RhinoBRep.from_shape(*args, **kwargs)
```
tomvanmele: in `compas_occ` this is a constructor for making a BRep directly from the underlying OCC shape. perhaps a more generic name that also makes sense in Rhino is `..._from_brep`. this would then directly populate the `_brep` attribute, rather than `occ_shape` and `rhino_brep`...

src/compas/geometry/__init__.py
```diff
@@ -1201,7 +1205,9 @@
     'NurbsCurve',
 
     'Surface',
-    'NurbsSurface'
+    'NurbsSurface',
+
+    "Brep",
```
tomvanmele, line 1240: will need to expose vertex, edge, loop, and face as well...

src/compas_rhino/geometry/brep/brep.py
```diff
@@ -0,0 +1,82 @@
...
+from .face import RhinoBRepFace
+from .edge import RhinoBRepEdge
+from .vertex import RhinoBRepVertex
+from .loop import RhinoBRepLoop
+
+
+class RhinoBRep(Geometry):
+    def __init__(self):
+        super(RhinoBRep, self).__init__()
+        self.rhino_brep = None
```
tomvanmele: `self._brep`?
and the same for `compas_occ` then, instead of `self.occ_shape`?

src/compas_rhino/geometry/brep/brep.py
```diff
@@ -0,0 +1,82 @@
...
+    def data(self, data):
+        faces = []
+        for facedata in data["faces"]:
+            face = RhinoBRepFace.from_data(facedata)
+            faces.append(face)
+        self.rhino_brep = BRep.from_faces(faces).rhino_brep
+
+    @classmethod
+    def from_faces(cls, faces):
+        brep = Brep()
```
tomvanmele: ?

src/compas_rhino/geometry/brep/brep.py
```diff
@@ -0,0 +1,82 @@
...
+            return [RhinoBRepLoop(l) for l in self.rhino_brep.Loops]
+
+    @property
+    def faces(self):
+        if self.rhino_brep:
+            return [RhinoBRepFace(f) for f in self.rhino_brep.Faces]
+
+    @property
+    def frame(self):
+        """TODO: what to do here? Brep has no Frame/plane information"""
```
tomvanmele: in Rhino?
i guess that means the frame is `Frame.worldXY()`

src/compas_rhino/geometry/brep/brep.py
```diff
@@ -0,0 +1,82 @@
...
+        if self.rhino_brep:
+            return self.rhino_brep.GetArea()
+
+    @property
+    def volume(self):
+        if self.rhino_brep:
+            return self.rhino_brep.GetVolume()
+
+    @classmethod
+    def from_shape(cls, shape):
```
tomvanmele: i think we should rename this to `from_brep`

src/compas_rhino/geometry/brep/edge.py
```diff
@@ -0,0 +1,35 @@
...
+from compas_rhino.conversions import line_to_rhino_curve
+
+
+from .vertex import RhinoBRepVertex
+
+
+class RhinoBRepEdge(Data):
+    def __init__(self, rhino_edge=None):
+        super(RhinoBRepEdge, self).__init__()
+        self._rhino_edge = None
```
tomvanmele: `self._edge`?

src/compas_rhino/geometry/brep/face.py
```diff
@@ -0,0 +1,41 @@
+from compas.data import Data
+from compas_rhino.geometry import RhinoNurbsSurface
+
+from .loop import RhinoBRepLoop
+
+
+class RhinoBRepFace(Data):
+
+    TOLERANCE = 1e-6
```
tomvanmele: need a better solution for this.
also just in general, we need a formal location to store `compas`-wide configuration options...

src/compas_rhino/geometry/brep/face.py
```diff
@@ -0,0 +1,41 @@
...
+
+    def __init__(self, rhino_face=None):
+        super(RhinoBRepFace, self).__init__()
+        self.loops = None
+        self.surface = None
+        if rhino_face:
```
tomvanmele, line 32: same :)

tomvanmele (commented):
good start! :)

src/compas_rhino/geometry/brep/__init__.py
chenkasirer: I agree. May I suggest we drop verboseness a notch and just call them `from_x`? otherwise we're looking at a `new_brep_from_brep`..

Also, I'd keep a `new_brep` plugin which would do something like `RhinoBRep.from_brep(Rhino.Geometry.Brep())` to essentially create an empty `Brep` which can be filled later on. This will allow:
```python3
from compas.geometry import BRep
b = BRep()
```
And in the future we could offer adding faces etc. What do you think?

And since we're discussing names, I went with `BRep` cause of `compas_occ` but I like `Brep` much better. What would you say about changing?

tomvanmele: > I agree. May I suggest we drop verboseness a notch and just call them `from_x`? otherwise we're looking at a `new_brep_from_brep`..

sounds good

tomvanmele: well and all the other stuff sounds good as well :)

src/compas_rhino/geometry/brep/brep.py
chenkasirer: To be implemented. 

While we're here, this part will have to be a bit different than `compas_occ`. `BrepEdge`, `BrepFace` and even `BrepVertex` all don't have a public constructor and can only be created via the `Brep` instance (e.g. `Brep().Vertices.Add(Point3d)`). 

Consequently, de-serializing a `compas_rhino.geometry.RhinoBRepVertex` alone does not result in a `Rhino.Geometry.BrepVertex` instance. Therefore, I imagine the de-serialized building blocks will be containing COMPAS types and the actual reconstruction of the native Rhino `Brep` will take place here, in `from_faces`. It can then also "inject" the native types to each of the underlying objects.

Do you see a better way to go about it?

tomvanmele: no, that makes sense to me. in fact, also in `compas_occ` we should try to funnel everything through the `Brep` class, and expose the other types only through the element accessors (`.vertices`, `.edges`, `.loops`, `.faces`), or for development pruposes.

src/compas/geometry/brep/brep.py
chenkasirer: As @gonzalocasas points out, I think it's worth discussing if this is part of the public interface, named one way or another. 

The main reason I kept it decorated with `property` is to allow replacing (or lazy setting) the native type after instantiation. The setter then already extracts the `data`. That way the other accessors are not dependent on the native instance being present, which is sometimes difficult due to reason mentioned in https://github.com/compas-dev/compas/pull/1057#discussion_r920874977.

Having said that, I guess there's no technical reason not to prefix the property's name with `_`, don't know how conventional that is though. Alternatively, could have an explicit `_set_native_instance()`.

Should the underlying native instance be in any way public?

src/compas_rhino/geometry/brep/brep.py
chenkasirer: So what happens if the input Brep is transformed?
Is frame information implicit in Brep the same way it is for, let's say, `Polyhedron` (i.e. the vertices get transformed)?

Working with `Polyhedron` I found it difficult when doing `Transformation.from_frame_to_frame()` if what I was transforming didn't have frame information. 
I saw two alternatives back then:
1. Add explicit `Frame` to keep track of the transformations
2. Avoid transforming in-place and always return a copy transformed from `Frame.worldXY()`

Or did I miss anything?

tomvanmele: i think it will be difficult to use a local frame for the brep objects, because it means not only the brep topology elements but also the underlying (nurbs) geometry will have to be expressed wrt to this local frame.

for the shapes, we have to make a distinction between parametric shapes (box, sphere, ...) and the non-parametric ones (i guess this is only polyhedron, which is a special case also in other regards). all the parametric ones should have a local frame. transformations are easy because you just transform the frame, and world coordinates can be calculated based on the parameters on the fly.

data structures (and polyhedron) are defined explicitly wrt to the global coordinate system.

tomvanmele: would do you mean by "keeping track of transformations"? transformations defining subsequent movements of a geometry in space? currently we differentiate between the geometry of an object and the viewobject (a bit like in Rhino and Blender). the viewobject can store a transformation stack that can keep track of such transformations...

chenkasirer: Ah was not aware of this differentiation between parametric and non-parametric shapes, explains a lot :)

So maybe this is outside the scope of the `Brep` implementation, but how would you handle subsequent transformation of a `Brep`?
First transformation would be from `Frame.worldXY()` to `frame01`
The next transformation would have to be from `frame01` to `frame02`, but `frame01` is lost, or at least not part of `Brep`.

src/compas_rhino/geometry/brep/face.py
chenkasirer: @gonzalocasas mentioned once these can be handled with some global COMPAS precision system. I'll look into it.

tomvanmele: well, i wouldn't necessarily call it a system, but there is indeed a precision setting `compas.PRECISION` :) 
a more robust version of this that also allows for other settings would be a good thing...

src/compas_rhino/geometry/brep/brep.py
tomvanmele: well i would argue you calculate the actual frame to frame transformation for each of those, and store the matrices in a transformation stack. but, again, not sure if this is something that should be added to the geometry itself...

tomvanmele: about the (non-)parametric shapes. this goes a bit in the direction of the discussion about adding a reference frame to all shapes.

i think in COMPAS 2.0 (or in the LTS version of 1.0, but i don't know if we manage to do that without breaking changes) we should differentiate between the following objects.

* Primitives: point, vector, line, plane, frame, polyline, polygon (defined by point locations), polyhedron (defined by polygons)
* Parametric shapes: circle, ellipse, regular polygon (defined by a radius and the number of sides), regular polyhedron (more or less the platonic solids), box, sphere, torus, cylinder, cone, ellipsoid
* Nurbs: curve, surface
* Boundary representations
* Function representations (not sure about this one)
* Transformations

the parametric shapes are implicitly defined wrt a local frame. all other objects are explicitly defined wrt the global coordinate system.

anyway, i know this is not part of this PR, but since we were talking about it...

gonzalocasas: sounds like a good categorization, we should eventually (ie. outside this PR) add `ParametricShape` and `Nurb` as a base classes. I believe all others already exist as classes. re. function representation, is that `compas_vol` sort of thing?

tomvanmele: yes indeed. perhaps it is a bit too specific, but some base infrastructure could easily be added in core...

src/compas/geometry/brep/brep.py
chenkasirer: Question about this: do we need an external/interface representation for these? In `compas_occ` I've only seen them being used via `is_compound()`, `is_shell()` etc.. If that's the case we don't really need a generic representation. Do you see a need?

tomvanmele: no, no need...

src/compas_rhino/geometry/brep/brep.py
```diff
@@ -106,7 +122,71 @@ def volume(self):
             return self._brep.GetVolume()
 
     def _create_rhino_brep(self, faces):
+        """
+          Things need to be defined in a valid brep:
+           1- Vertices
+           2- 3D Curves (geometry)
+           3- Edges (topology - reference curve geometry)
+           4- Surface (geometry)
```
tomvanmele: surfaceS

src/compas_rhino/geometry/brep/brep.py
```diff
@@ -106,7 +122,71 @@ def volume(self):
...
+           5- Faces (topology - reference surface geometry)
+           6- Loops (2D parameter space of faces)
+           4- Trims and 2D curves (2D parameter space of edges)
```
tomvanmele: typically you don;t have to do this explicitly. the edge curves are converted to surface trims automatically, provided that is possible of course.

src/compas/geometry/brep/brep.py
```diff
@@ -0,0 +1,478 @@
...
+            faces.append(face.data)
+        return {"faces": faces}
+
+    @data.setter
+    def data(self):
+        raise NotImplementedError
+
+    @property
+    def orientation(self):
+        raise NotImplementedError
```
tomvanmele: why is this in the data section?

src/compas/geometry/brep/brep.py
```diff
@@ -0,0 +1,478 @@
...
+
+    # ==============================================================================
+    # Properties
+    # ==============================================================================
+
```
tomvanmele, line 232: perhaps the first (and foremost) property be `brep` or `shape`, referring to the underlying Brep object of the implementing backend

src/compas/geometry/brep/brep.py
```diff
@@ -0,0 +1,478 @@
...
+    # Constructors
+    # ==============================================================================
+
+    @classmethod
+    def from_brep(cls, brep):
+        return from_brep(brep)
+
+    @classmethod
+    def from_step_file(cls, filename):
+        return from_step_file(filename)
```
tomvanmele: this is not for now, but i was thinking that we should start naming methods that load data from files (and save data to files) differently than the other "constructors", because they do something fundamentally different. especially in the "to_..." variations the naming is not great. all the other "to" methods return a different representation of the object, whereas methods who save to a file do not...

src/compas/geometry/brep/brep.py
```diff
@@ -0,0 +1,478 @@
...
+    def from_torus(cls, torus):
+        return from_torus(torus)
+
+    @classmethod
+    def from_mesh(cls, mesh):
+        return from_mesh(mesh)
+
+    @classmethod
+    def from_faces(cls, faces):
+        raise NotImplementedError
```
tomvanmele: perhaps we should rename this to `from_brepfaces`

src/compas/geometry/brep/brep.py
```diff
@@ -0,0 +1,478 @@
...
+
+    # ==============================================================================
+    # Converters
+    # ==============================================================================
+
+    def to_json(self, filepath):
+        raise NotImplementedError
+
+    def to_step(self, filepath):
+        raise NotImplementedError
```
tomvanmele: again, perhaps we should start naming there `save(as)_json` and `save(as)_step` or just `save` and determine the behaviour based on the file type

src/compas/geometry/brep/brep.py
```diff
@@ -0,0 +1,478 @@
...
+
+    def to_tesselation(self, linear_deflection=LINEAR_DEFLECTION):
+        raise NotImplementedError
+
+    def to_meshes(self, u=16, v=16):
+        raise NotImplementedError
+
+    def to_viewmesh(self):
```
tomvanmele: this should have some kind of precision setting

src/compas/geometry/brep/brep.py
```diff
@@ -0,0 +1,478 @@
...
+    # trim
+    # rotate
+    # translate
+    # unjoin edges
+
+    def make_solid(self):
+        raise NotImplementedError
+
+    def check(self):
+        raise NotImplementedError
```
tomvanmele: rename this to `is_valid`?

src/compas/geometry/brep/brep.py
```diff
@@ -0,0 +1,478 @@
...
+        None
+
+        """
+        NotImplementedError
+
+    def transform(self, matrix):
+        raise NotImplementedError
+
+    def transformed(self, matrix):
+        raise NotImplementedError
```
tomvanmele: this can already have an implementation
```python
brep = self.copy()
brep.transform(matrix)
return brep
```

src/compas/geometry/brep/brep.py
```diff
@@ -0,0 +1,478 @@
...
+
+    def contours(self, planes):
```
tomvanmele: just noticing this now, but we should add a description in the docstring of the expected parameters and returns. ideally we do this with type hints but of course this doesn't work in Rhino...

src/compas/geometry/brep/edge.py
```diff
@@ -0,0 +1,228 @@
...
+    """
+
+    # ==============================================================================
+    # Properties
+    # ==============================================================================
+
+    @property
+    def is_line(self):
+        """
+        Returns True if this edge is a line, False otherwise.
```
tomvanmele: "if the geometry of the edge is a line"

tomvanmele: same "the geometry of" everywhere else

docs/tutorial/brep.rst
```diff
@@ -0,0 +1,138 @@
+***********************
+Boundary Representation
+***********************
+
+.. rst-class:: lead
+
+Boundary representation (Brep) support is realized in COMPAS using its `plugin` system.
+The expected interface for Brep related classes is defined in the `compas.geometry.brep` module
```
tomvanmele: :mod:`compas.geometry.brep`

tomvanmele (commented):
this is an old review i forgot to submit so it might be a bit outdated...

src/compas/geometry/brep/brep.py
chenkasirer: Right you are. It's actually already implemented in `Geometry` which is the super class of `Brep`.

src/compas_rhino/geometry/brep/brep.py
chenkasirer: yea I must say I was a bit surprised there was an additional curve needed for the trims, then I found an example where they are generated from the 3D curve of the edge and the face's surface. It's a bit magic for me but it's been working so far. Could you say if that's generally the way to go?
```python
    def _create_trim_curve(rhino_edge, rhino_surface):
        curve_2d = rhino_surface.Pullback(rhino_edge.EdgeCurve, TOLERANCE)
        curve_2d.Reverse()
        return curve_2d
```

src/compas/geometry/brep/brep.py
chenkasirer: How about `native_brep`? I just find `brep` alone a bit confusing given that in the same scope there's at times 3 different types that all represent a Brep in different worlds (`compas.brep`/`compas_rhino.brep`/`Rhino.Geometry.Brep`). I find `shape` doesn't really convey this either.

src/compas/geometry/brep/brep.py
chenkasirer: Question about docstrings approach @brgcode @gonzalocasas, whenever we have an interface like `compas.geometry.Brep` and then the implementation, in this case `compas_rhino.geometry.Brep`, do we put the docstrings on the interface, on the implementation or both?

chenkasirer: `+` I'm gonna need help understanding what some of these methods should even do.

src/compas/geometry/brep/brep.py
```diff
@@ -0,0 +1,1011 @@
...
+def from_polygons(*args, **kwargs):
+    raise PluginNotInstalledError()
+
+
+@pluggable(category="factories")
+def from_curves(*args, **kwargs):
+    raise PluginNotInstalledError()
+
+
+class BrepType:
```
tomvanmele: i think, in Rhino (py 2.7) this will result in an "old style" class.
should be `class BrepType(object)`.
however, not sure this matters here...

src/compas/geometry/brep/brep.py
```diff
@@ -0,0 +1,1011 @@
...
+        -------
+        list[:class:`~compas.datastructures.Mesh`]
+
+        """
+        raise NotImplementedError
+
+    def to_viewmesh(self, precision):
+        """
+        Convert this Brep to a view mesh
+        Parameters
```
tomvanmele, line 679: missing newline

src/compas/geometry/brep/brep.py
```diff
@@ -0,0 +1,1011 @@
...
+        ----------
+        precision:
+            float
```
tomvanmele: `precision : float`

src/compas/geometry/brep/brep.py
```diff
@@ -0,0 +1,1011 @@
...
+    # Other Methods
+    # ==============================================================================
+
+    def trim(self, trimming_plane, tolerance):
+        """Trim this Brep using the given trimming plane
+
+        Parameters
+        ----------
+        trimming_plane: defines the trimming plane
+            :class:`~compas.geometry.Frame
```
tomvanmele:
```
trimming_plane : :class:`~compas.geometry.Frame`
    The trimming plane.
```

src/compas/geometry/brep/brep.py
```diff
@@ -0,0 +1,1011 @@
...
+
+        tolerance: the tolerance to use when trimming
+            float
```
tomvanmele:
```
tolerance : float
    The tolerance to use for trimming.
```

src/compas/geometry/brep/brep.py
```diff
@@ -0,0 +1,1011 @@
...
+        """Generate contour lines by slicing the Brep shape with a series of planes.
+
+        Parameters
+        ----------
+        planes : list[:class:`~compas.geometry.Plane`]
+            The slicing planes.
+
+        Returns
+        -------
+        list[list[:class:`~compas.geometry.Polyline`]]
```
tomvanmele, line 838: there is some inconsistency in the docstrings between using `List` and `list`

src/compas/geometry/brep/face.py
```diff
@@ -0,0 +1,172 @@
+from compas.data import Data
+
+
+class BrepFace(Data):
+    """
+    An interface for a Brep Face
+    """
```
tomvanmele, line 38: currently, properties will not show up in the documentation if they are documented in their own docstrings. to include them, please list them here under `Attributes`

tomvanmele: this means you can omit the docstring in the property definition...

src/compas/geometry/brep/face.py
```diff
@@ -0,0 +1,172 @@
...
+        """
+        raise NotImplementedError
+
+    @property
+    def centroid(self):
+        raise NotImplementedError
+
+    @property
+    def is_valid(self):
+        """
```
tomvanmele: here this is listed as a property, but in the main Brep class as a method...

src/compas/geometry/brep/loop.py
```diff
@@ -0,0 +1,60 @@
+from compas.data import Data
+
+
+class BrepLoop(Data):
+    """
+    An interface for a Brep Loop
+    """
```
tomvanmele, line 16: same remark about the docs of the properties

src/compas/geometry/brep/face.py
```diff
@@ -0,0 +1,172 @@
+from compas.data import Data
+
+
+class BrepFace(Data):
```
tomvanmele, line 4: because of how the plugins work, using the default constructor will not trigger one of the backends to be selected.

1. we are okay with this.
2. we overwrite the default constructor, such that it always requires a native face to be provided and creating an empty face is not possible.

tomvanmele: same for edge, vertex, etc

src/compas/geometry/brep/vertex.py
```diff
@@ -0,0 +1,40 @@
...
+
+    # ==============================================================================
+    # Properties
+    # ==============================================================================
+
+    @property
+    def point(self):
+        """
+        Returns the point representing this vertex
+        Returns
```
tomvanmele: missing newline

src/compas/geometry/brep/vertex.py
```diff
@@ -0,0 +1,40 @@
...
+        -------
+
```
tomvanmele: missing return type

src/compas/geometry/brep/vertex.py
```diff
@@ -0,0 +1,40 @@
...
+    # ==============================================================================
+    # Constructors
+    # ==============================================================================
+
+    @classmethod
+    def from_point(cls, point):
+        """
+        Creates a Brep vertex from a Point
+        Parameters
```
tomvanmele, line 30: missing newline

src/compas/geometry/brep/vertex.py
```diff
@@ -0,0 +1,40 @@
...
+        ----------
+        point: :class:`compas.geometry.Point`
+            The point to create a vertex from
+        Returns
```
tomvanmele, line 35: missing newline

src/compas_rhino/geometry/brep/brep.py
```diff
@@ -0,0 +1,248 @@
...
+    # this makes de-serialization backend-agnostic.
+    # The deserializing type is determined by plugin availability when de-serializing
+    # regardless of the context available when serializing.
+    __class__ = Brep
+
+    def __new__(cls, *args, **kwargs):
+        # This breaks the endless recursion when calling `compas.geometry.Brep()` and allows
+        # having Brep here as the parent class. Otherwise RhinoBrep() calls Brep.__new__()
+        # which calls RhinoBrep() and so on...
+        return object.__new__(cls, *args, **kwargs)
```
tomvanmele, line 60: i don't think this is correct. cf. the Curve and Surface constructors...

src/compas_rhino/geometry/brep/brep.py
```diff
@@ -0,0 +1,248 @@
...
+    def data(self, data):
+        """
+        Parameters
+        ----------
+        data
+
+        Returns
+        -------
+
+        """
```
tomvanmele: setters are meant to be documented on the getter...

src/compas_rhino/geometry/brep/brep.py
```diff
@@ -0,0 +1,248 @@
...
+
+    @property
+    def native_brep(self):
+        """
+        Returns the native representation of this Brep.
+
+        Returns
+        -------
+        :class: `Rhino.Geometry.Brep`
+        """
```
tomvanmele: any reason some of the properties and others not? perhaps leave this one out as well?

src/compas/geometry/brep/face.py
chenkasirer: That's a good point. 
I think that `BrepFace`, `BrepVertex`, `BrepEdge` and `BrepLoop` are best described as pure interfaces. Unlike `Brep()` I don't thing they should trigger anything, yet I agree you shouldn't be able to create instances of them. 

Maybe all we need is to define them as abstract?
```python
class BrepFace(metaclass=ABCMeta, Data):
    ...
```

src/compas_rhino/geometry/brep/brep.py
chenkasirer: We discussed this and decided to leave it for the time being. 

For the sake of documentation, this is to allow:
1. having `Brep` as a super class of `RhinoBrep`
2. creating an empty e.g. `RhinoBrep` using just `Brep()`


The way it's done in e.g. `Surface` makes 1 possible but not 2.

src/compas/geometry/brep/face.py
tomvanmele: if i remember correctly this creates major problem in ironpython with memory leaks etc

tomvanmele (approved):

src/compas/geometry/brep/face.py
chenkasirer: I see. I've actually heard about this from @gonzalocasas before. Couldn't find any reports that seemed relevant online though. 

Mmh then I guess we can also just leave it for now. Initially, I thought this will cause an issue once we shift to the new COMPAS2.0 serialization, but I'm thinking it might still make sense to do all (de)serialization from `Brep` since it obviously doesn't make sense to mix backends.

src/compas_rhino/geometry/brep/brep.py
chenkasirer: You mean why there's public access to the native type in Brep but not in the others? 

I found this is useful at the moment due to the lack of a `BrepArtist` for example. That way you can plug-in `Brep.native_brep` to a GH geometry component to visualize.

Maybe once there's more extensive support for Brep in COMPAS it would make sense to remove this one as well.

tomvanmele: no sorry, there were some crucial words missing...

any reason some of the properties HAVE DOCSTRINGS and others not

chenkasirer: No particular reason ;)
Removed all docstrings from properties and added to Attributes instead.

PR #1058, opened 2022-07-13
Gltf update

romanarust:
Gltf update for allowing extensions, [see here](https://github.com/KhronosGroup/glTF/tree/main/extensions/2.0/Khronos). Check the changelog for details.

### Checklist


src/compas/files/gltf/gltf_content.py
```diff
@@ -403,30 +490,132 @@ def visit(node, key):
 # Main
 # ==============================================================================
 
-if __name__ == '__main__':
+if __name__ == "__main__":
```
beverlylytle, line 494: I see the value of the changes you've made to this sections pointing out caveats to adding images and textures.  But I also see the value of providing a bare minimum example of how to create a GLTF.  Do you think it would be worthwhile to keep both?  

The other changes in the PR look great!

src/compas/files/gltf/gltf_content.py
```diff
@@ -403,30 +490,132 @@ def visit(node, key):
...
+        name=image_uri,
+        mime_type=MineType.PNG,
+        uri=image_file,
+    )
+    image_idx = cnt.add_image(image_data)
+    # TextureData.source takes the key of the ImageData that it should use as 'source'
+    texture = TextureData(source=image_idx)
+    texture_idx = cnt.add_texture(texture)
+
+    texture = TextureData(source=image_idx)
```
beverlylytle, line 541: Is it really necessary to add the same texture twice?  Can it not just be reused?

src/compas/files/gltf/gltf_content.py
```diff
@@ -267,6 +270,90 @@ def get_node_vertices(self, node):
...
+        int
+        """
+        key = len(self.textures)
+        while key in self.textures:
+            key += 1
+        self.textures[key] = texture
+        return key
+
+    def add_image(self, image):
+        """Adds a image to the content.
```
beverlylytle: "an image"

src/compas/files/gltf/gltf_content.py
romanarust: yes, true. but i wanted to test if the texture gets detected also in deeper depths of the gltf (it is added to `pbr_specular_glossiness.diffuse_texture` further below) to not be removed by `remove_orphans`

src/compas/files/gltf/gltf_content.py
romanarust: thank you beverly! 
yes, for now, I would like to keep both, as there might be further improvements needed.

src/compas/files/gltf/extensions.py
```diff
@@ -0,0 +1,335 @@
...
+        transmission_factor=None,
+        transmission_texture=None,
+        extensions=None,
+        extras=None,
+    ):
+        super(KHR_materials_transmission, self).__init__(extras, extensions)
+        self.transmission_factor = transmission_factor
+        self.transmission_texture = transmission_texture
+
+    def to_data(self, texture_index_by_key, **kwargs):
```
chenkasirer, line 29: Out of curiosity, why is `BaseGLTFDataClass` not inheriting from `compas.data.Data` and then the usual implementation of the `data` interface? are we intentionally keeping this package independent?

src/compas/files/gltf/extensions.py
```diff
@@ -0,0 +1,335 @@
...
+    """glTF extension that defines the clearcoat material layer.
+
+    https://github.com/KhronosGroup/glTF/blob/master/extensions/2.0/Khronos/KHR_materials_clearcoat
+    """
+
+    key = "KHR_materials_clearcoat"
+
+    def __init__(
+        self,
+        clearcoat_factor=None,
```
chenkasirer, line 154: Are any of these maybe not optional? Since it seems we're not bound to the `data` interface here, you could pass on the `None` default and assume they're present (of course, the user can still send `None`s but then it's their fault ;)

This will also allow you to skip all these `is not None` checks when serializing.

src/compas/files/gltf/gltf_content.py
```diff
@@ -267,6 +270,90 @@ def get_node_vertices(self, node):
...
+        """Adds a texture to the content.
+
+        Parameters
+        ----------
+
+        Returns
+        -------
+        int
+        """
+        key = len(self.textures)
```
chenkasirer: mmh maybe I'm missing something but what happens here exactly? 
if it's about mapping an incrementing integer to textures why not just
```python3
key = len(self.textures) + 1 
self.textures[key] = texture
```

But if that's all we're doing, why not just make `selt.texture` a `List`. Then you'd have this mapping is implied and you could use `enumerate` if you need.

src/compas/files/gltf/gltf_content.py
```diff
@@ -267,6 +270,90 @@ def get_node_vertices(self, node):
...
+        while key in self.materials:
+            key += 1
+        self.materials[key] = material
+        return key
+
+    def add_texture(self, texture):
+        """Adds a texture to the content.
+
+        Parameters
+        ----------
```
chenkasirer, line 313: `texture` missing in docstring

src/compas/files/gltf/gltf_content.py
```diff
@@ -267,6 +270,90 @@ def get_node_vertices(self, node):
...
+        while key in self.textures:
+            key += 1
+        self.textures[key] = texture
+        return key
+
+    def add_image(self, image):
+        """Adds an image to the content.
+
+        Parameters
+        ----------
```
chenkasirer, line 329: `image` missing in doc string

src/compas/files/gltf/gltf_content.py
```diff
@@ -267,6 +270,90 @@ def get_node_vertices(self, node):
...
+
+        Returns
+        -------
+        int
+        """
+        key = len(self.images)
+        while key in self.images:
```
chenkasirer: same as textures..

src/compas/files/gltf/gltf_content.py
```diff
@@ -267,6 +270,90 @@ def get_node_vertices(self, node):
...
+        Parameters
+        ----------
+        name : str
+            The name of the material
+
+        Returns
+        -------
+        int or None
+        """
+        for key in self.materials:
```
chenkasirer: Doesn't make much difference, but `dict().items()` returns a `key:value` tuple for each item, so you could do instead:
```python3
for key, material in self.materials.items():
    if material.name == name:
        return key
```

src/compas/files/gltf/gltf_content.py
```diff
@@ -403,30 +490,132 @@ def visit(node, key):
...
 
     import os
     import compas
-
+    import urllib
+    from compas.geometry import Box, Frame
     from compas.datastructures import Mesh
+    from compas.utilities import download_file_from_remote
     from compas.files.gltf.gltf import GLTF
+    from compas.files.gltf.data_classes import (
```
chenkasirer: I think the COMPAS standard for imports is single-line imports. if you run `isort` it should **sort** ;) it out for you.

CHANGELOG.md
```diff
@@ -8,8 +8,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 
 ### Added
+* Added gltf extensions: `KHR_materials_transmission`, `KHR_materials_specular`, `KHR_materials_ior`, `KHR_materials_clearcoat`, `KHR_Texture_Transform`, `KHR_materials_pbrSpecularGlossiness` 
+* Added `GLTFContent.check_extensions_texture_recursively`
+* Added `GLTFContent.get_node_by_name`, `GLTFContent.get_material_index_by_name`
+* Added `GLTFContent.add_material`, `GLTFContent.add_texture`, `GLTFContent.add_image`
+* Added `GLTFExporter._add_extensions_recursively` and added the check in mostly all `GLTFExporter._add_*`
```
chenkasirer: I think the general rule is that changes to protected/private members should not typically appear CHANGELOG

chenkasirer (commented):
No idea about GLTF but left a few comments anyways :)

One general comment: I'm guessing (but unsure) that you used `black` with no parameters for the formatting. By default, it breaks lines which are longer than 88 characters whereas the COMPAS standard is, apparently, 180. So try running `black -l 180` (that's a small L).

src/compas/files/gltf/extensions.py
romanarust: hmm, can't remember why, but I think it is also not needed to inherit from `compas.data.Data`  as the point is to convert to gltf and not json

src/compas/files/gltf/extensions.py
romanarust: there is probably a better way to handle this -you are right-, but then I would need to change the `gltf.data_classes` as well, since the same checking is performed there...

src/compas/files/gltf/gltf_content.py
romanarust: i think we cannot be sure that users use the `add_texture` and set the key numbers in ascending order. the 
```
while key in self.textures:
     key += 1
```
ensures that we don't overwrite a key that has been wrongly set.

Having  `self.texture` as `dict` follows the  implementation for `self.materials, self.images, ...`

chenkasirer: Sure. I'd maybe put it in a static `_get_next_available_key()` function as I find it's not immediately clear what's happening. Also it's a recurring pattern in all these `add_x()` functions.

On a more nitpicky note, if user is expected to interact with `self.textures` directly, they might also put in keys that aren't at all integers. So maybe `textures` should be marked as `protected/private`. Also, if the keys have no meaning other than being unique, you could consider using `uuid`.

src/compas/files/gltf/extensions.py
beverlylytle: Yes, this is the reason:  glTF is it's own serialization standard separate from compas' serialization standard.

src/compas/files/gltf/extensions.py
beverlylytle: I can't speak to the optionality of these specific attributes, but in general, many attributes in glTF are optional.  I believe though that it's invalid to include `null` values in the JSON part of a glTF.  When the values in question are `None`, those keys should not be added to the data which is then dumped to the JSON even if it is the value the user passed.  I agree though that there probably is a better way to do this.

src/compas/files/gltf/gltf_content.py
beverlylytle: In the glTF format, things like textures are stored in a list and for a material to reference a texture, it references the texture by the texture's index in that list.  If a user imports an existing glTF into compas and wants to modify it by adding or removing a texture, it would be very hard to update all the references to those indices consistently.  So, as a work around, these dictionaries were added as a substitute, where materials reference a texture's key until export where the key is replaced by the finalized texture list index.  It's true though that the keys don't have to be integers, but it was a convenient choice when writing the importer.

chenkasirer: Cool :) Thank you for the explanation, @beverlylytle!

src/compas/files/gltf/gltf_content.py
```diff
@@ -267,6 +264,95 @@ def get_node_vertices(self, node):
...
+        Parameters
+        ----------
+        name : str
+            The name of the material
+
+        Returns
+        -------
+        int or None
+        """
+        for key, material in self.materials:
```
beverlylytle: I think this needs `.items()`

romanarust:
ipy keeps failing, but the tests with ipy work fine on my computer

chenkasirer:
> 

Since for me locally these tests fail as well, I had a look and to be honest I had difficulties following everything that's happening in this code (again, not familiar with GLTF) so I'll go ahead and put here what I could find and maybe it would have more of a meaning to you, knowing what's going on. 

Focusing on parsing and exporting of `SpecularTest.glb`: 
4 textures are loaded and parsed from the `.glb` file and AFAIK all is parsed properly.

When exporting though, 3 out of the four textures are removed here for being "unvisited"
```python
# in GLTFExporter.load()
self._content.remove_orphans()
```
```python
#before removing orphans: 
    {
        0: <TextureData object at 0x0000000000000A7A>, 
        1: <TextureData object at 0x0000000000000A7B>,  
        2: <TextureData object at 0x0000000000000A7C>, 
        3: <TextureData object at 0x0000000000000A7D>
    }

# removing unvisited item - key:1 item:<TextureData object at 0x0000000000000A7B>
# removing unvisited item - key:2 item:<TextureData object at 0x0000000000000A7C>
# removing unvisited item - key:3 item:<TextureData object at 0x0000000000000A7D>
```
which results in the following `self._content.textures`:
```python
{
    0: <TextureData object at 0x0000000000000A7A>
}
```
Later, in `_add_materials()`, => 
`material_data.to_data(self._texture_index_by_key)` => 
`self.extensions_to_data(texture_index_by_key=texture_index_by_key)` => 
`self.specular_texture.to_data(texture_index_by_key)`

The index of the `TextureInfoData` that's inside the extension is being looked up in `texture_index_by_key`, but it's not there since it has been removed earlier in `remove_orphans()`.

Whether or not these textures should have been removed, or if this should get somehow updated in the references to them inside the extensions, I cannot say. But it seems that the reason these tests don't fail in CPython is that there these textures don't get removed in the first place. 

@romanarust can you say if these textures should even be removed or not?

romanarust:
@chenkasirer : Thanks a lot for checking and the detailed explanation! 
I suspect that the new function `check_extensions_texture_recursively(material)`, which is called in `remove_orphans` is not working properly in ipy.
which ipy version are you using?

chenkasirer:
> 
```
>>> ipy
IronPython 2.7.12 (2.7.12.1000)
[.NETFramework,Version=v4.5 on .NET Framework 4.8.4510.0 (64-bit)]
```
```
>>> ipy -m compas
Yay! COMPAS is installed correctly!

COMPAS: 1.16.0-fea3e9f7
```

beverlylytle:
I was just thinking about this, and I agree with your suspicions @romanarust.  There are a few things in python 2 that can behave a bit differently than in python 3 that might be exposed in that function.  In particular, there is always the new vs old style class issue (which I think shouldn't be a problem because everything inherits from `object`), but still I wonder if `isinstance` is behaving the way it is intended.  I also wonder if `dir` and `callable` behave in exactly the same way between the two versions.  Another (perhaps far-flung) thought is scoping.  Python 2's scoping of variables is a little different, but only with list/dict comprehensions and not recursive functions as far as I know.

chenkasirer:
@romanarust  @beverlylytle So following your suspicions, I put some more time into this issue and sure enough, the culprit seems to be `isinstance` in `check_extensions_texture_recursively`.
Some calls to `isinstance(item, TextureInfoData)` return `False` even though type is actually a child of `TextureInfoData`.

Excerpt from my debugging extravaganza: 
```python
# GLTFExporter.check_extensions_texture_recursively
print("a: {} instance:{} is texture:{}".format(a, a_instance, isinstance(a_instance, TextureInfoData)))
...
a: specular_texture instance:<TextureInfoData object at 0x0000000000000352> is texture:False
```
This prevents the texture from being flagged as `visited` and it gets removed later, which causes all the trouble when exporting.

Annoyingly enough, this problem is intermittent:
1. It doesn't happen when running the code directly from an `ipy` session
2. Is doesn't happen when running just one of the tests, only when running both
3. the worst: it doesn't always happen for both tests, sometimes only one of them fails

To check for the types more reliably I'd suggest either comparing strings e.g. `a.__class__.__name__ == "compas.files.TextureDataInfo"`
Or the more elegant (but still hacky) solution, in my opinion, add some class attribute and check for it instead:
```python
# data_classes.py
class TextureInfoData(object):
    IS_TEXTURE_INFO_DATA = True  # value is not actually checked..
    ...

# gltf_content.py
...
if getattr(instance, "IS_TEXTURE_INFO_DATA", False):  # instead of isinstance(a, TextureInfoData)
    texture_visit_log[getattr(item, a).index] = True
...
```
I'd do the same for the check `isinstance(a, BaseGLTFDataClass)` although I haven't had issues with it with the current tests.

beverlylytle:
oh ironpython... what is dead may never die.

romanarust:
Thank you @chenkasirer for the detailed analysis and yet more proof of why IronPython should be buried.
I implemented your suggested changes. It fails again, but due to a different problem.

chenkasirer:
> 

@romanarust Not your fault actually. Seems there's a new version of `flake8` which enforces some stuff that hasn't been detected/enforced before. The code that causes it to fail has been around for at least the past 15 months..

I'll sort it out in a separate PR.

**Update: opened https://github.com/compas-dev/compas/pull/1061

chenkasirer (approved):
LGTM!

PR #1062, opened 2022-08-11
fix color getattr error

Licini:
Give a default value to `getattr`, to prevent it from raising `AttributeError`
image


tomvanmele (approved):
LGTM

PR #1065, opened 2022-09-02
add transformed to `compas_rhino.geometry.Curve`

Licini:
As titled


gonzalocasas (changes requested):
mmm something is not right here.. `Curve` inherits from `Geometry` and `Geometry` has a default implementation of `transformed()` which is the same that you added here, am I missing something?

tomvanmele:
@gonzalocasas is right, this should not be necessary...
@Licini ?

Licini:
> @gonzalocasas is right, this should not be necessary... @Licini ?

Right.... I somehow missed that, I made this PR because I bumped into a `NotImplemented` error when calling curve.transformed, and this one got it fixed. Let me double check again

Licini:
@brgcode @gonzalocasas Turned out to be false alarm. the error I met must be cause by some other reasons. Now I just tested it works properly. Closing this one now.

PR #1074, opened 2022-09-21
Brep artist

chenkasirer:
Added `BrepArtist`s for GH and Rhino.

### Checklist


CHANGELOG.md
```diff
@@ -20,6 +20,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Added `format` task using `black` formatter.
 * Added a `test_intersection_circle_circle_xy` in the `test_intersections`
 * Added split operation to `compas_rhino.geometry.Brep`.
+* Added a `RhinoArtist` in `compas_rhino`.
+* Added a `RhinoArtist` in `compas_ghpython`.
```
gonzalocasas, line 24:
```suggestion
* Added a `BrepArtist` in `compas_rhino`.
* Added a `BrepArtist` in `compas_ghpython`.
```

gonzalocasas (approved):
LGTM!

gonzalocasas:
Side comment: I still find it weird that I cannot do `Brep.from_shape(box)`. And I tried `Brep.from_brep(box)` which I understood from another PR was the current naming but that didn't work either.

chenkasirer:
> 

Ahh I think I get now what you're asking. 
Originally `shape` was referring to the native/underlying type. Like in OCC that would be `TopoDS_Shape` (which is a superclass of the Brep solid), and in Rhino this would be `Rhino.Geometry.Brep`. So `Brep.from_shape` would take in the native type and instantiate our wrapper.

In the context of COMPAS `shape` sounded quite ambiguous so we decided on `from_brep` instead. 
Originally discussed [here](https://github.com/compas-dev/compas/pull/1057#discussion_r920507239).

As for creating a `compas.geometry.Brep` from the different COMPAS shape primitives, each primitive has its own explicit factory function: `from_box`, `from_sphere` etc. Maybe in the future we could add a `to_brep()` to the `compas.geometry.Geometry` interface. That way you could do `Box().to_brep()` a la Rhino.

Put in code
```python
# Rhino context

from compas.geometry import Brep, Box
box = Box.from_width_heigh_depth(...)
b = Brep.from_box(box)
```
Or alternatively (and this is how `from_box` is implemented, of course)
```python
from compas.geometry import Brep, Box
from compas_rhino.conversions import box_to_rhino

import Rhino

box = Box.from_width_height_depth(...)
rhino_box = box_to_rhino(box)
box_brep = rhino_box.ToBrep()
b = Brep.from_brep(brep)
```

gonzalocasas:
Exactly, I meant COMPAS shape primitives. Have a `Brep.from_shape()` would make a lot of sense because we already have `Mesh.from_shape()`, so it's almost logical to assume there would be a matching method on Breps.

> ```python
> box = Box.from_width_height_depth(...)
> rhino_box = box_to_rhino(box)
> box_brep = rhino_box.ToBrep()
> b = Brep.from_brep(brep)
> ```

About this snippet, I think `Brep.from_brep()` should be called `Brep.from_native_brep()` or something less prominent, because now it is pretty confusing. In fact, now it allows to do `Brep.from_brep(compas_brep_instance)` without complaining but then the brep has zero everything.

chenkasirer:
> 

OK yea I see what you mean. I agree. 
Good idea to rename `from_brep` -> `from_native_brep`, I'll open another PR for that. `compas_occ` would ideally follow suit @brgcode.

The `from_shape` I'd implement the way I mentioned above, by adding a `to_brep` to either `Shape` or `Geometry`. It would be kinda analogous to `Shape.to_vertices_and_faces` which `Mesh.from_shape` uses. Also on it's own PR.

About `Brep.from_brep(compas_brep_instance)`, interesting.. would you add a type check there? we don't normally do, right?

tomvanmele:
will update `compas_occ` accordingly...

tomvanmele:
i would prefer `from_native` over `from_native_brep`, but either way is fine for me...

gonzalocasas:
> About `Brep.from_brep(compas_brep_instance)`, interesting.. would you add a type check there? we don't normally do, right?

agree, we don't normally do it, but it's weird that it doesn't fail...  Perhaps just changing to `to_native` is enough to signal that this should not be done.

PR #1078, opened 2022-10-03
Rhino surface curvature fix

nmaslarinos:
The RhinoSurface.curvature_at should return a Vector of the normal direction at the given u,v parameters of the surface. However, accessing rg.Surface.CurvatureAt returns an object of the class SurfaceCurvature, not a vector. 
image
In order to access the principal curvature vector, we need to call the method Direction() of the SurfaceCurvature class. 
Perhaps optional flag attributes could be introduced for the various attributes the SurfaceCurvature class holds like Mean and Gaussian curvature.
image

- [ ] Breaking change: bug fix or new feature that involves incompatible API changes.

### Checklist


src/compas_rhino/geometry/surfaces/surface.py
```diff
@@ -189,7 +189,8 @@ def curvature_at(self, u, v):
         :class:`~compas.geometry.Vector`
 
         """
-        vector = self.rhino_surface.CurvatureAt(u, v)
+        surface_curvature = self.rhino_surface.CurvatureAt(u, v)
+        vector = surface_curvature.Direction()
```
gonzalocasas: I get an error with this, because the `Direction` method needs an `int direction` parameter (0 or 1):
image

gonzalocasas (changes requested):

src/compas_rhino/geometry/surfaces/surface.py
nmaslarinos: Correct, I forgot about this. Apparently, with 0 it returns the maximum principal curvature direction and with 1 it returns the minimum principal curvature direction. My guess is that the maximum could be generally more useful than the minimum.
image

gonzalocasas: Yes, I think so too :)

nmaslarinos: I updated the code passing a 0 argument :)

gonzalocasas (approved):
LGTM! @brgcode agree?

CHANGELOG.md
```diff
@@ -21,6 +21,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 * Fixed `Color.__get___` AttributeError.
 
+* Fixed  `RhinoSurface.curvature_at` not returning a Vector, but a Rhino SurfaceCurvature class object
```
tomvanmele, line 30: never really understand why people do this :)
could you move to line 23 and leave 24 blank?

src/compas_rhino/geometry/surfaces/surface.py
```diff
@@ -189,7 +189,8 @@ def curvature_at(self, u, v):
         :class:`~compas.geometry.Vector`
 
         """
-        vector = self.rhino_surface.CurvatureAt(u, v)
+        surface_curvature = self.rhino_surface.CurvatureAt(u, v)
+        vector = surface_curvature.Direction(0)
```
tomvanmele: wouldn't we rather change it to return a tuple of the two principle curvature directions?
otherwise, to compute the perpendicular direction, the user would have to also get the normal vector and compute an additional cross product...

tomvanmele: or even return a tuple with the same info as contained in the surface curvature object?

CHANGELOG.md
nmaslarinos: I thought it was to separate different contributors, I am removing it!

src/compas_rhino/geometry/surfaces/surface.py
nmaslarinos: The tuple as a return would be an elegant solution, there are many values returned that could be of use

tomvanmele: would you mind updating the PR with the tuple as output?
@gonzalocasas no?

nmaslarinos: Absolutely! I will do it later today :)

src/compas_rhino/geometry/surfaces/surface.py
```diff
@@ -186,11 +186,23 @@ def curvature_at(self, u, v):
...
+        surface_curvature = self.rhino_surface.CurvatureAt(u, v)
+        if surface_curvature:
+            point, normal, kappa_u, direction_u, kappa_v, direction_v, gaussian, mean = surface_curvature
+            cpoint = point_to_compas(point)
+            cnormal = vector_to_compas(normal)
+            cdirection_u = vector_to_compas(direction_u)
+            cdirection_v = vector_to_compas(direction_v)
+            return (cpoint, cnormal, kappa_u, cdirection_u, kappa_v, cdirection_v, gaussian, mean)
+        else:
+            return None
```
tomvanmele: just fyi, this part is redundant...

tomvanmele (approved):
@gonzalocasas ok?

gonzalocasas:
> @gonzalocasas ok?

yep! 👍

src/compas_rhino/geometry/surfaces/surface.py
nmaslarinos: right, I didn't realise it!

gonzalocasas:
Thanks @nmaslarinos !

PR #1090, opened 2022-10-18
Use a junction instead of a symbolic link

tomvanmele:
Supposedly, this solves the installation problem for people with minimal rights on their computers...

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+* Changed `compas._os._polyfill_symlinks` to use junction (/J) instead of symbolic link (/D).
+
 ### Removed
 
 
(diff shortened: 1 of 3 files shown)


tomvanmele:
the difference with a symbolic link is minimal, and in fact it might be the better choice, reading this...
https://superuser.com/questions/343074/directory-junction-vs-directory-symbolic-link

gonzalocasas (changes requested):
I don't think this works for all our uses. In particular, we use symlinks to link ghuser components, and junctions only works with directories, so, it would fail in that one. If we want to change to junctions, we could change the install procedure of gh components to copy them instead of symlink them (there's no real reason to symlink those I think).

tomvanmele:
something like this?

tomvanmele:
@gonzalocasas ping

gonzalocasas (approved):
lgtm

gonzalocasas:
Three small comments: 
1) there's some commented out code that could be deleted
2) this change causes the privilege elevation pop up to appear multiple times
3) the `install_userobjects` function was added later, so, `compas_fab` has an older implementation in which the same code is in that repo, so this change doesn't affect it. I will need to update that

PR #1095, opened 2022-11-08
Workflow

Licini:
Upgrade workflows to `v2` set, which uses the finally supported nested workflows and if conditions, making outside much more cleaner.

https://github.com/compas-dev/compas-actions.build/blob/main/action.yml
https://github.com/compas-dev/compas-actions.docs/blob/main/action.yml
https://github.com/compas-dev/compas-actions.publish/blob/main/action.yml

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -18,7 +18,7 @@ jobs:
         python: ['3.8', '3.9', '3.10']
 
     steps:
-      - uses: compas-dev/compas-actions.build@v2
+      - uses: compas-dev/compas-actions.build@v3
         with:
           python: ${{ matrix.python }}
-          test_lint: true
+          invoke_lint: true
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -9,22 +9,16 @@ on:
       - main
 
 jobs:
-  build-packages:
+  build:
+    if: "!contains(github.event.pull_request.labels.*.name, 'docs-only')"
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        python-version: ['3.6', '3.7', '3.8', '3.9', '3.10']
+        python: ['3.8', '3.9', '3.10']
 
     steps:
-      - uses: actions/checkout@v2
-
-      - name: Set up Python ${{ matrix.python-version }}
-        uses: actions/setup-python@v2
+      - uses: compas-dev/compas-actions.build@v2
         with:
-          python-version: ${{ matrix.python-version }}
-
-      - uses: compas-dev/compas-actions.build@v1.0.0
-        with:
+          python: ${{ matrix.python }}
           test_lint: true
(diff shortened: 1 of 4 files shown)


.github/workflows/build.yml
```diff
@@ -9,22 +9,16 @@ on:
...
-        with:
-          python-version: ${{ matrix.python-version }}
-
-      - uses: compas-dev/compas-actions.build@v1.0.0
+      - uses: compas-dev/compas-actions.build@v2
         with:
-          test_lint: true
-          test_compas: true
+          python: ${{ matrix.python }}
+          test_lint: true
```
gonzalocasas: perhaps it would make sense to rename these inputs, currently `test`, `test_compas`, `test_lint`, could turn into: `test`, `lint`, `check_import`, or `invoke_lint`, `invoke_test`, `check_import`?

tomvanmele (approved):
LGTM

Licini:
@gonzalocasas all good?

.github/workflows/release.yml
```diff
@@ -1,74 +1,34 @@
...
-        if: success() && startsWith(github.ref, 'refs/tags')
-        run: |
-            twine check dist/*
-            twine upload dist/* --skip-existing
-        env:
-            TWINE_USERNAME: __token__
-            TWINE_PASSWORD: ${{ secrets.PYPI }}
+          pypi_token: ${{ secrets.PYPI }}
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          python: '3.10'
```
gonzalocasas: why is python specified again here?

tomvanmele:
@gonzalocasas @Licini can we try to wrap this up?

.github/workflows/build.yml
Licini: indeed, will rename it

.github/workflows/release.yml
Licini: The previous os-python matrix only applies to`build` job for one last overall testing before release. If passed, then the `publish` job is allowed to run. The python version here is used either for building gh component: https://github.com/compas-dev/compas-actions.publish/blob/1d0330a035f8030ce4d3f3f9d16c74c32d031e64/action.yml#L62
Or simple pypi publishing :https://github.com/compas-dev/compas-actions.publish/blob/1d0330a035f8030ce4d3f3f9d16c74c32d031e64/action.yml#L67

However the default value is `3.10` anyway and this line could be omitted.

Licini:
> @gonzalocasas @Licini can we try to wrap this up?

Sorry was interrupted by all the vacations stuff... @gonzalocasas Could you take one more look at the last changes? thx!

Licini:
> > @gonzalocasas @Licini can we try to wrap this up?
> 
> Sorry was interrupted by all the vacations stuff... @gonzalocasas Could you take one more look at the last changes? thx!

mmm wait there is some weird issue i need to resolve first..

gonzalocasas (approved):
LGTM

PR #1104, opened 2022-12-05
Add per-vertex color specification option to mesh drawing in Rhino

tomvanmele:

### Checklist

### DIFF
--- a/src/compas_rhino/utilities/drawing.py
+++ b/src/compas_rhino/utilities/drawing.py
@@ -782,6 +782,7 @@ def draw_mesh(vertices, faces, name=None, color=None, vertex_color=None, disjoin
     System.Guid
 
     """
+
     def average_color(colors):
         c = len(colors)
         r, g, b = zip(*colors)
--- a/src/compas_rhino/utilities/drawing.py
+++ b/src/compas_rhino/utilities/drawing.py
@@ -783,10 +783,11 @@ def draw_mesh(vertices, faces, name=None, color=None, vertex_color=None, disjoin
 
     """
     def average_color(colors):
+        c = len(colors)
         r, g, b = zip(*colors)
-        r = sum(r) / len(face)
-        g = sum(g) / len(face)
-        b = sum(b) / len(face)
+        r = sum(r) / c
+        g = sum(g) / c
+        b = sum(b) / c
         return int(r), int(g), int(b)
 
     vertex_color = vertex_color or {}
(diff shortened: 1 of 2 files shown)


src/compas_rhino/utilities/drawing.py
```diff
@@ -767,7 +767,13 @@ def draw_mesh(vertices, faces, name=None, color=None, disjoint=False, **kwargs):
     faces : list[list[int]]
         A list of faces as lists of indices into `vertices`.
     name : str, optional
+        The name of the mesh object in Rhino.
     color : tuple[[int, int, int]], optional
+        The base color of the mesh.
+    vertex_color : dict[int, tuple[int, int, int]], optional
```
chenkasirer, line 773: I find it not obvious what's the expected mapping here. If a class is an overkill then maybe just explicit example in the docstring. `vertex_index => (r, g, b)`?

src/compas_rhino/utilities/drawing.py
```diff
@@ -786,29 +795,48 @@ def draw_mesh(vertices, faces, name=None, color=None, disjoint=False, **kwargs):
...
                 c = mesh.Vertices.Add(*vertices[face[2]])
                 d = mesh.Vertices.Add(*vertices[face[3]])
+                vertexcolors.append(vertex_color.get(face[0], color))
+                vertexcolors.append(vertex_color.get(face[1], color))
+                vertexcolors.append(vertex_color.get(face[2], color))
+                vertexcolors.append(vertex_color.get(face[3], color))
                 mesh.Faces.AddFace(a, b, c, d)
             else:
                 if MeshNgon:
+                    cornercolors = [vertex_color.get(vertex) for vertex in face]
```
chenkasirer: This is the color interpolation, right? maybe it's worth moving it to some helper function to make it more readable.

src/compas_rhino/utilities/drawing.py
```diff
@@ -833,6 +869,12 @@ def draw_mesh(vertices, faces, name=None, color=None, disjoint=False, **kwargs):
     guid = add_mesh(mesh)
 
     if guid != System.Guid.Empty:
+        if vertexcolors:
+            try:
+                compas_rhino.set_mesh_vertex_colors(guid, vertexcolors)
```
chenkasirer, line 874: Not sure here, but what happens if user specifies `vertexcolors` but not `color`? is that even a valid combination?
From https://developer.rhino3d.com/api/RhinoScriptSyntax/#mesh-MeshVertexColors
```
Note, for each vertex, there must
  be a corresponding vertex color. If the value is None, then any
  existing vertex colors will be removed from the mesh
```

chenkasirer (approved):
LGTM!

PR #1106, opened 2022-12-06
Rhino brep

chenkasirer:
Updated `RhinoBrep` serialization format and reconstruction procedure to fix invalid reconstruction on more complex shapes. Highlights are:
- Changed [serialization format](https://salty-halloumi-27e.notion.site/Schema-72a82d96a1bb4617ae96142977ecb416) to eliminate duplicate elements.
- Serializing to/from primitives instead of NURBS where possible
- Moved brep reconstruction logic from `RhinoBrep` over to a the new `brep/builder.py` module.
- Added representation for trims with `BrepTrim`.
- Added `RhinoNurbsSurface.from_frame`

image

### Checklist


src/compas/geometry/brep/trim.py
```diff
@@ -0,0 +1,40 @@
...
+class BrepTrimIsoStatus(object):
+    """An enumeration of isoparametric curve direction on the surface."""
+
+    NONE = 0
+    X = 1
+    Y = 2
+    West = 3
+    South = 4
+    East = 5
+    North = 6
```
gonzalocasas: should these be uppercased?

src/compas_rhino/geometry/brep/face.py
```diff
@@ -107,23 +122,34 @@ def is_plane(self):
...
 
     @staticmethod
-    def _make_surface_from_loop(loop):
-        # order of corners determines the normal of the resulting surface
-        corners = [loop.edges[i].start_vertex.point for i in range(4)]
-        surface = RhinoNurbsSurface.from_corners(corners)
+    def _make_surface_from_data(surface_type, surface_data, uv_domain):
+        if surface_type == "plane":
+            frame = Frame.from_data(surface_data)
+            surface = RhinoNurbsSurface.from_frame(frame, uv_domain[0], uv_domain[1], (1, 1), (2, 2))
```
gonzalocasas: what's the meaning of `(1, 1), (2, 2)`? does it need a comment?

src/compas_rhino/geometry/brep/edge.py
```diff
@@ -116,12 +119,40 @@ def vertices(self):
...
 
     @property
     def is_ellipse(self):
-        return self._curve.IsEllipse()
+        return self._edge.EdgeCurve.IsEllipse()
+
+    def _get_curve_geometry(self):
+        curve = self._edge.EdgeCurve
+        if self.is_line:
+            type_ = "line"
```
gonzalocasas, line 135: what's the meaning of the trailing underscore?

CHANGELOG.md
```diff
@@ -9,9 +9,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+* Added conversion function `frame_to_rhino_plane` to `compas_rhino.conversions`.
+* Added `RhinoSuface.from_frame` to `compas_rhino.geometry`.
```
tomvanmele: not important but there is an "r" missing in Surface

gonzalocasas (approved):
I find it hard to judge if the brep parts are correct, but it LGTM!

tomvanmele:
would it not make more sense to add stuff to the builder instead of adding the builder to each of the `from_data` functions?

src/compas/geometry/brep/trim.py
chenkasirer: yup!

src/compas_rhino/geometry/brep/edge.py
chenkasirer: sometimes I insist on naming a var `type` so this is to avoid shadowing the builtin. It's apparently PEP8:
```
single_trailing_underscore_: used by convention to avoid conflicts with Python keyword, e.g.:
```

gonzalocasas: ahhh thanks!

src/compas_rhino/geometry/brep/brep.py
```diff
@@ -70,18 +71,22 @@ def __init__(self, brep=None):
...
-            faces.append(face)
-        self._create_native_brep(faces)
+        builder = RhinoBrepBuilder()
+        for v_data in data["vertices"]:
+            RhinoBrepVertex.from_data(v_data, builder)
+        for e_data in data["edges"]:
+            RhinoBrepEdge.from_data(e_data, builder)
+        for f_data in data["faces"]:
+            RhinoBrepFace.from_data(f_data, builder)
+        self._brep = builder.result
```
tomvanmele, line 89: would it not make more sense to add stuff to the builder rather than to modify the signature of  all the constructors to send the builder around?

src/compas_rhino/geometry/brep/builder.py
```diff
@@ -0,0 +1,164 @@
...
+        c_index = self.instance.AddTrimCurve(curve)
+        edge = self.instance.Edges[edge_index]
+        trim = self.instance.Trims.Add(edge, is_reversed, self.loop, c_index)
+        trim.IsoStatus = iso_status
+        trim.SetTolerances(TOLERANCE, TOLERANCE)
+        return trim
+
+    @property
+    def result(self):
+        return self.loop
```
tomvanmele: is this loop somehow updated in the `add_trim` method

src/compas_rhino/geometry/brep/builder.py
```diff
@@ -0,0 +1,164 @@
...
+    """Builds a Brep loop.
+
+    Attributes
+    ----------
+    result : :rhino: Rhino.Geometry.BrepTrim
+        The created loop.
+
+    """
+
+    def __init__(self, loop=None, instance=None):
```
tomvanmele: could you add the types of these parameters in the docstring above because it is not very clear to me what they are supposed to be

src/compas_rhino/geometry/brep/builder.py
tomvanmele: actually, how/when/where is this result attribute actually used (in any of the builder classes)?

tomvanmele (commented):
can we a short meeting about the code?
have the feeling it would be more meaningful than writing a thousand comments...

src/compas_rhino/geometry/brep/builder.py
chenkasirer: renamed and added to docstrings to maybe make some more sense

src/compas_rhino/geometry/brep/brep.py
chenkasirer: The builder serves:
- keeps the reconstruction logic in a single place
- keeps the reconstruction logic separate from the de-serialization
- makes the reconstruction context (i.e. `Brep`) available to each of the components while avoiding global state
- enables the de-serialized elements to hold the native `Rhino.Geometry` type immediately after de-serialization.

`BrepBuilder` is the main context, it essentially wraps around the newly created `Rhino.Geometry.Brep` and holds it until it's properly built.

Since Face, Loop and Trim are nested in each other:

`BrepFaceBuilder` wraps around `Rhino.Geometry.BrepFace` and handed over to the loops so that they can add themselves to the face.
`BrepLoopBuilder` wraps around `Rhino.Geometry.BrepLoop` and handed over to the trims so that they can add themselves to the loop.

The `result` property of the builders gives access to the newly built native element once it's ready. `BrepLoopBuilder.result` returns the `Rhino.Geometry.BrepLoop` after all the trims have been added to it. `BrepFaceBuilder.result` returns the `Rhino.Geometry.BrepFace` after all the loops have been added to it. `BrepBuilder.result` returns the `Rhino.Geometry.Brep` after all the vertices, edges and faces have been added to it.

src/compas_rhino/geometry/brep/face.py
chenkasirer: These are the required constant to create a simple planar NURBS surface with first degree and 2 control point in each direction (4 corners). Agreed with @brgcode these will be moved as default kwargs to the signature of `from_frame` since these will normally be used.

src/compas_rhino/geometry/surfaces/surface.py
```diff
@@ -189,7 +189,7 @@ def from_plane(cls, plane, box):
         return cls.from_rhino(rhino_surface)
 
     @classmethod
-    def from_frame(cls, frame, u_interval, v_interval, uv_degrees, uv_point_counts):
+    def from_frame(cls, frame, u_interval, v_interval, uv_degrees=(1, 1), uv_point_counts=(2, 2)):
```
tomvanmele: would it not make more sense to do `u_degree=1, v_degree=1, u_count=2, v_count=2`? since this is how it is done for the intervals as well...

chenkasirer: agree. done.

tomvanmele (approved):

PR #1131, opened 2023-03-23
fix in `area_polygon`

baehrjo:
Modified the `area_polygon` method to return abs(area) as it was sometimes returning a negative (but numerically correct) area. In the original code (below), the method is vectorially summing 'pieces' of area, with sign + or - relatively to the direction of n0 : if dot_vectors(n, n0) > 0 .... else .... 
n0 is computed starting with two vertices a and b of the polygon. If these vertices lie in a locally concave zone of the polygon (n0 direction is flipped regarding the normal of the entire polygon), the final area will return negative.
(See the test)

+ suggest : add a condition or raise an error if the polygon is self-intersecting ?
For instance a symmetric hourglass shaped polygon (e.g. two triangles joined at their summit) will have an area of 0 while the projected area will be twice the single triangle area.

``` original code
o = centroid_points(polygon)
    a = polygon[-1]
    b = polygon[0]
    oa = subtract_vectors(a, o)
    ob = subtract_vectors(b, o)
    n0 = cross_vectors(oa, ob)
    area = 0.5 * length_vector(n0)
    for i in range(0, len(polygon) - 1):
        oa = ob
        b = polygon[i + 1]
        ob = subtract_vectors(b, o)
        n = cross_vectors(oa, ob)
        if dot_vectors(n, n0) > 0:
            area += 0.5 * length_vector(n)
        else:
            area -= 0.5 * length_vector(n)
    return area
```

- [x ] Bug fix in a **backwards-compatible** manner.

### Checklist

- [x ] I added a line to the `CHANGELOG.md` file in the `Unreleased` section under the most fitting heading (e.g. `Added`, `Changed`, `Removed`).
- [x ] I ran lint on my computer and there are no errors (i.e. `invoke lint`).
- [x ] I have added tests that prove my fix is effective or that my feature works.

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -36,3 +36,4 @@
 - Chen Kasirer <<kasirer@arch.ethz.ch>> [@chenkasirer](https://github.com/chenkasirer)
 - Nickolas Maslarinos <<maslarinosnickolas@gmail.com>> [@nmaslarinos](https://github.com/nmaslarinos)
 - Katerina Toumpektsi <<e.toumpeksti@gmail.com>> [@katarametin](https://github.com/katarametin)
+- Joelle Baehr-Bruyere <<joelle.baehr-bruyere@epfl.ch>> [@baehrjo](https://github.com/baehrjo)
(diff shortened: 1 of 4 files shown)


tomvanmele:
hi,

switching to the absolute value of the area sounds good.

however, for self-intersecting polygons, i am not sure what the correct solution is. the definition of area is then a bit unclear. for polygon centroids we added a note to the docs about this (https://compas.dev/compas/latest/api/generated/compas.geometry.centroid_polygon_xy.html#compas.geometry.centroid_polygon_xy). perhaps we should do the same for the area?

checking for self-intersections is costly and would be a big penalty for all normal cases...

tomvanmele:
you have some lint in your code though

tomvanmele:
https://github.com/compas-dev/compas/actions/runs/4503258653/jobs/7940691230?pr=1131#step:2:963

tomvanmele (approved):

PR #1134, opened 2023-03-25
Fix release name

Licini:
Added a new parameter `release_name_prefix` in `compas-actions.publish` 
So that the releasees are created with name like `COMPAS 1.2.3` rather than `v1.2.3`


tomvanmele (approved):
LGTM

PR #1135, opened 2023-04-18
Simplified `Part`

chenkasirer:
Introducing a simplified, more generic, version of `Part` as well as base classes for features: `Feature`, `GeometricFeature` and `ParametricFeature`.

* Handling of geometry creation is moved to an extending class (see `compas_timber.parts.Beam`).
* Added base classes for features (see `compas_timber.parts.BeamExtendingFeature` and `compas_timber.parts.BeamTrimmingFeature`)
* Removed the transformation deque
* Removed `to_mesh`

This has been used during the development of `compas_timber.parts.Beam`. Moving it over from `compas_future` to allow for open-sourcing of `compas-timber`.

### Checklist

### DIFF
--- a/src/compas/datastructures/assembly/part.py
+++ b/src/compas/datastructures/assembly/part.py
@@ -32,6 +32,27 @@ class GeometricFeature(Feature):
     An implementation of this class may offer support for various geometry types by adding an entry to OPERATIONS
     mapping the geometry type to its corresponding operation.
 
+    Examples
+    --------
+    >>>    def trim_brep_plane(brep, plane):
+    >>>         # trim brep with plane, return trimmed brep
+    >>>
+    >>>    def trim_mesh_plane(mesh, plane):
+    >>>         # trim mesh with plane, return trimmed mesh
+    >>>
+    >>>    class TrimmingFeature(GeometricFeature):
+    >>>        OPERATIONS = {Brep: trim_brep_plane, Mesh: trim_mesh_plane}
+    >>>
+    >>>            def __init__(self, trimming_plane):
+    >>>                super(TrimmingFeature, self).__init__()
+    >>>                self._geometry = trimming_plane
+    >>>
+    >>>            def apply(self, part):
+    >>>                part_geometry = part.get_geometry(with_features=True)
+    >>>                type_ = Brep if isinstance(part_geometry, Brep) else Mesh
+    >>>                operation = OPERATIONS[type_]
+    >>>                return operation(part_geometry, self._geometry)
+
     """
 
     OPERATIONS = {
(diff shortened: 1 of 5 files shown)


src/compas/datastructures/assembly/part.py
```diff
@@ -2,23 +2,92 @@
...
+            The part onto which this feature should be applied.
+
+        """
+        raise NotImplementedError
+
+
+class GeometricFeature(Feature):
+    """Base class for geometric feature which may be applied to a :class:`~compas.datastructures.Part`.
+
+    Applies a binary operation on Part's current geometry and the feature's.
```
tomvanmele, line 30: if this operation is always a boolean, perhaps we should rename to `BooleanFeature`?

src/compas/datastructures/assembly/part.py
```diff
@@ -2,23 +2,92 @@
...
+
+    An implementation of this class may offer support for various geometry types by adding an entry to OPERATIONS
+    mapping the geometry type to its corresponding operation.
+
```
tomvanmele, line 34: Perhaps it would be good to add an example, so implementers know what is expected/intended...

src/compas/datastructures/assembly/part.py
```diff
@@ -2,23 +2,92 @@
...
+    @data.setter
+    def data(self, value):
+        self._geometry = value["geometry"]
+
+
+class ParametricFeature(Feature):
+    """Base class for Features that may be applied to the parametric definition
+    of a :class:`~compas.datastructures.Part`.
+
```
tomvanmele, line 79: Perhaps it would be good to add an example so implementers...

src/compas/datastructures/assembly/part.py
```diff
@@ -107,134 +143,49 @@ def data(self):
...
-        """Transform the part with respect to the local cooordinate system.
-
-        Parameters
-        ----------
-        T : :class:`~compas.geometry.Transformation`
-
-        Returns
-        -------
-        None
+        self.frame = data["frame"]
```
tomvanmele, line 191: :)

tomvanmele (approved):

src/compas/datastructures/assembly/part.py
chenkasirer: It's not though. Does binary mean boolean? I just meant that it is performed between two geometries, one from the part and the other from the feature.

This has currently two implementations in `compas_timber`: boolean subtraction and plane trimming

src/compas/datastructures/assembly/part.py
chenkasirer: :))

chenkasirer: :))

src/compas/datastructures/assembly/part.py
tomvanmele: no it doesn't. all good then...

PR #1139, opened 2023-04-24
Geometry cleanup and some additions

tomvanmele:

### Checklist


src/compas/geometry/bbox/bbox_numpy.py
```diff
@@ -264,6 +246,9 @@ def oabb_numpy(points):
         XYZ coordinates of 8 points defining a box.
 
     """
+    from compas.geometry import Transformation
```
gonzalocasas, line 249: Is this to prevent some circular reference loading problem?

src/compas/geometry/primitives/plane.py
```diff
@@ -288,6 +292,26 @@ def from_frame(cls, frame):
...
+        Returns
+        -------
+        :class:`~compas.geometry.Plane`
+
+        """
+        if compas.NUMPY:
+            from compas.geometry import bestfit_plane_numpy as bestfit_plane
+        else:
+            from compas.geometry import bestfit_plane
+        point, normal = bestfit_plane(points)
```
gonzalocasas, line 312: it would be cool to convert this to a pluggable/plugin thingy

CHANGELOG.md
```diff
@@ -10,6 +10,33 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 * Added `create_id` to `compas_ghpython.utilities`. (moved from `compas_fab`)
+* Added igl plugin for `compas.geometry.trimesh_gaussian_curvature`.
+* Added igl plugin for `compas.geometry.trimesh_principal_curvature`.
+* Added igl plugin for `compas.geometry.trimesh_geodistance`.
+* Added igl plugin for `compas.geometry.trimesh_isolines`.
+* Added igl plugin for `compas.geometry.trimesh_harmonic` and `compas.geometry.trimesh_lscm`.
```
gonzalocasas, line 17: Shouldn't all of these plugins be implemented in `compas_igl` and only the pluggable here?

src/compas/geometry/trimesh/parametrisation.py
```diff
@@ -5,17 +5,18 @@
 from compas.plugins import pluggable
 
 
-__all__ = ["trimesh_harmonic", "trimesh_lscm"]
-
-
 @pluggable(category="trimesh")
-def trimesh_harmonic(M):
+def trimesh_harmonic(M, boundary, values):
```
gonzalocasas, line 9: Please add this change of signature to the changelog.

src/compas/geometry/trimesh/parametrisation.py
```diff
@@ -31,13 +32,15 @@ def trimesh_harmonic(M):
 
 
 @pluggable(category="trimesh")
-def trimesh_lscm(M):
+def trimesh_lscm(M, boundary):
```
gonzalocasas, line 35: same here

gonzalocasas (commented):
Generally looking good, but the one important thing that is missing is unit tests for the new functions. Other than that, only some small comments here and there, the most relevant perhaps is the igl pluggables one.

src/compas/geometry/bbox/bbox_numpy.py
tomvanmele: yes, it makes more sense like this in the `__init__`. as a general rule, if you need any of the objects in the lower level functions, you need to import it in that scope.

src/compas/geometry/primitives/plane.py
tomvanmele: yes, wanted to ask about. should we make all the numpy implementations plugins?

CHANGELOG.md
tomvanmele: yes and no.
the current implementations are based on the official `igl` python bindings, but they are not great (some functions really don't work). if we can get `compas_libigl` to become available on `conda` we can switch to using those indeed.

src/compas/geometry/primitives/plane.py
gonzalocasas: ideally yes, it would simplify the calling code and always choose the best possible implementation

CHANGELOG.md
gonzalocasas: Is there anything preventing us from moving `compas_libigl` to `conda`? or just needs to be done?

tomvanmele: it is a bit trickier than CGAL but it should be possible. don't remember why exactly i gave up at the time...

tomvanmele: but feel free to have a go at it :)

PR #1144, opened 2023-05-20
General clean up of init files

tomvanmele:

### Checklist


tomvanmele:
@gonzalocasas @chenkasirer @licini ready for review

src/compas/geometry/_core/angles.py
Licini, line 1: Not a big deal, but I find these two folder names `_core` and `_algebra` that start with "_" feel a bit inconsistent with the rest

src/compas/geometry/icp/icp_numpy.py
```diff
@@ -84,6 +79,9 @@ def icp_numpy(source, target, tol=1e-3):
     >>>
 
     """
+    from compas.geometry import Transformation
+    from compas.geometry import Frame
+
```
Licini, line 84: Is this to avoid circular dependency? Could it be resolved by using relative import at the top?

src/compas/geometry/primitives/_primitive.py
Licini, line 1: mmmmm, still not a big fan of _ in front of file name

src/compas/utilities/coercing.py
Licini, line 1: Should we just officially let this go already in 2.0?

src/compas/utilities/encoders.py
Licini, line 1: Same above?

Licini (approved):
Looking good! Just some minor observations

src/compas/geometry/icp/icp_numpy.py
tomvanmele: no that doesn't change anything

src/compas/geometry/primitives/_primitive.py
tomvanmele: here it can be removed indeed

tomvanmele: will do this in the geometry PR though

PR #1146, opened 2023-05-25
Extendible artist

chenkasirer:
Makes the context discovery of `compas.artists.Artist` a bit more flexible so that that other artists can be added which are not `Rhino`, `Grasshopper` or `Blender`.

Rather than explicitly checking for either of the three CAD's, `Artist` now is chosen based on a priority list:
1. Viewer (to be implemented)
2. Plotter (deprecated)
4. Grasshopper/Rhino
7. Other CAD (3rd party context [first available context which is none of the above, based on plugin detected contexts])

The user can also override this choice by adding an explicit using `Artist(item, context="Blender")` 

### Checklist


chenkasirer:
@gonzalocasas @brgcode been wondering. What could be a strategy for choosing the default `Artist` when several contexts are available? 

I see Rhino and GH as a special case since both are generally available, you can use a `RhinoArtist` from GH (e.g. to bake) but it makes sense to check and set default to the GHPtyhon one. In Rhino however, using a `GHPythonArtist` does nothing.. so it should be checked.

Outside Rhino though it gets less obvious for me. Like `Plotter` would typically be available and maybe other ones in the future (e.g. `ViewArtist` ;) should we leave it random? should we let the user know somehow that they should choose?

src/compas/artists/artist.py
```diff
@@ -94,7 +117,8 @@ class Artist(object):
 
     def __new__(cls, item, **kwargs):
         if not Artist.__ARTISTS_REGISTERED:
-            register_artists()
+            detected_contexts = cls.register_artists()
```
tomvanmele: is this not fragile?
should we perhaps not add a function that detects available contexts explicitly?

tomvanmele:
> @gonzalocasas @brgcode been wondering. What could be a strategy for choosing the default `Artist` when several contexts are available?
> 
> I see Rhino and GH as a special case since both are generally available, you can use a `RhinoArtist` from GH (e.g. to bake) but it makes sense to check and set default to the GHPtyhon one. In Rhino however, using a `GHPythonArtist` does nothing.. so it should be checked.
> 
> Outside Rhino though it gets less obvious for me. Like `Plotter` would typically be available and maybe other ones in the future (e.g. `ViewArtist` ;) should we leave it random? should we let the user know somehow that they should choose?

in case of plotter and/or viewer, we could try to detect if an instance of the plotter or the viewer was created and then set the corresponding context. this does mean the user would have to make sure to create an instance of the viewer or plotter before attempting at using the artists if they want the artists to use the plotter or viewer as context...

src/compas/artists/artist.py
chenkasirer: On second thought, it would be more robust to take the detected contexts from `Artist.ITEM_ARTIST`.. that why no changes are needed on the plugin side.

src/compas/artists/artist.py
```diff
@@ -27,35 +27,73 @@ def redraw():
...
+    # TODO: implement
+    return False
+
+
+def _choose_context():
+    """Chooses an appropriate context depending on available contexts and open instances. with the following priority:
+    1. Viewer
+    2. Plotter
+    3. Rhino / GH
+    4. Other - randome choice from Artist.ITEM_ARTISTS.values()
```
tomvanmele: should this not just be combined into "CAD"? since you can only be in one CAD environment at the same time...

src/compas/artists/artist.py
```diff
@@ -27,35 +27,73 @@ def redraw():
...
     if compas.is_grasshopper():
         return "Grasshopper"
     if compas.is_rhino():
         return "Rhino"
     if compas.is_blender():
         return "Blender"
-    return None
+    other_contexts = [v for v in Artist.ITEM_ARTIST.keys() if v not in Artist.KNOWN_CONTEXTS]
+    if other_contexts:
+        return other_contexts[0]
```
tomvanmele, line 82: i assume this is to accommodate the context you are working on now? sounds like a new problem in the making as soon as there is more than 1. a bit like with the plugins...

src/compas/artists/artist.py
```diff
@@ -27,35 +27,73 @@ def redraw():
...
-    if "context" in kwargs:
-        Artist.CONTEXT = kwargs["context"]
-    else:
-        Artist.CONTEXT = identify_context()
-
-    if Artist.CONTEXT is None:
-        raise NoArtistContextError()
-
+    # in any case user gets to override the choice
+    context_name = kwargs.get("context") or _choose_context()
```
tomvanmele: ``_detect_current_context``? since there is no active choice possible...

src/compas/artists/artist.py
```diff
@@ -88,13 +132,12 @@ class Artist(object):
 
     __ARTISTS_REGISTERED = False
 
-    AVAILABLE_CONTEXTS = ["Rhino", "Grasshopper", "Blender", "Plotter"]
-    CONTEXT = None
+    KNOWN_CONTEXTS = ["Rhino", "Grasshopper", "Blender", "Plotter", "Viewer"]
```
tomvanmele: perhaps this can be populated during the plugin discovery? then automatically all registered artist plugins add their context to the "known contexts"...

src/compas/artists/artist.py
```diff
@@ -27,35 +27,73 @@ def redraw():
...
+    """Chooses an appropriate context depending on available contexts and open instances. with the following priority:
+    1. Viewer
+    2. Plotter
+    3. Rhino / GH
+    4. Other - randome choice from Artist.ITEM_ARTISTS.values()
+
+    Returns
+    -------
+    str
+        one-of: ["Viewer", "Plotter", "Grasshopper", "Rhino"] or another as detected by the register_artists plugin.
```
tomvanmele: Blender is missing here. also, a set of choices is ``{'a', 'b', 'c', 'd'}``

src/compas/artists/artist.py
```diff
@@ -73,12 +111,18 @@ def _get_artist_cls(data, **kwargs):
 class Artist(object):
     """Base class for all artists.
 
+    Parameters
+    ----------
+    item: Any
+        The item which should be visualized using the created Artist.
+    context: str, optional
+        Explicit context to pick the Artist from. One of :attr:`AVAILABLE_CONTEXTS`.
```
tomvanmele, line 123: :attr:`KNOWN_CONTEXTS`

tests/compas/artists/test_artists.py
```diff
@@ -41,3 +44,19 @@ def test_get_artist_cls_with_out_of_order_registration():
...
     artist = Artist(item, context="fake")
     assert isinstance(artist, FakeSubArtist)
+
+
+if not compas.IPY:
+
+    def test_artist_auto_context_discovery(mocker):
+        mocker.patch("compas.artists.Artist.register_artists")
+        Artist.register_artists.side_effect = register_fake_context
+        Artist._Artist__ARTISTS_REGISTERED = False
```
tomvanmele, line 70: `Artist.__ARTISTS_REGISTERED = False`

tomvanmele (commented):
a few more things...

src/compas/artists/artist.py
chenkasirer: As discussed, after removing Plotter we can assume only one context available at any given time (with the exception of GH/Rhino which is checked explicitly).

src/compas/artists/artist.py
chenkasirer: Removed this class attribute as `Artist.ITEM_ARTIST` is already populated with discovered contexts by the plugins.

src/compas/artists/artist.py
chenkasirer: rephrased the comment

src/compas/artists/artist.py
chenkasirer: removed `KNOWN_CONTEXTS`

src/compas/artists/artist.py
```diff
@@ -27,35 +28,76 @@ def redraw():
...
+def is_plotter_open():
+    """Returns True if an instance of the Plotter is available.
+
+    Returns
+    -------
+    bool
+
+    """
+    # TODO: implement
+    return False
```
tomvanmele, line 56: shall we not remove this?

src/compas/artists/artist.py
```diff
@@ -27,35 +28,76 @@ def redraw():
...
+    Returns
+    -------
+    str
+        Name of an available context, used as key in :attr:`Artist.ITEM_ARTIST`
+
+    """
+    if is_viewer_open():
+        return "Viewer"
+    if is_plotter_open():
+        return "Plotter"
```
tomvanmele, line 75: remove?

src/compas/artists/artist.py
```diff
@@ -27,35 +28,76 @@ def redraw():
...
-    else:
-        Artist.CONTEXT = identify_context()
+    # in any case user gets to override the choice
+    context_name = kwargs.get("context") or _detect_current_context()
 
-    if Artist.CONTEXT is None:
-        raise NoArtistContextError()
+    # TODO: remove when Plotter is removed
+    if context_name == "Plotter":
+        warn("The usage of Plotter with COMPAS Artist is deprecated!")
```
tomvanmele, line 92: ha ok, i see. should we not just immediately remove it?

tomvanmele (approved):

src/compas/artists/artist.py
chenkasirer: Will do in a separate PR. Maybe we can discuss briefly what we exactly wanna do there. I could remove the artist registration for the plotter artists but then we might as well just remove the entire artists package of compas_plotter.

tomvanmele: okay let's just get this one merged and then we will see...

PR #1150, opened 2023-06-06
fix for offset_polygon not working with a Polygon input

nmaslarinos:
Fix a bug in `offset_polygon` raising a TypeError when the input was a Polygon instead of a list of Points. Added an example in the docstring that correctly runs with a Polygon as an input. 
The fix is rather simple, in the line `polygon = polygon + polygon[:1]` we just need to add a `[:]` after the `polygon` in order to get the list in both input cases (Polygon or list of Points).

### Checklist


tomvanmele (approved):
thanks! have had this error many times myself...

nmaslarinos:
Just to make our lives simpler 😊

PR #1152, opened 2023-06-08
added split methods to polyline

nmaslarinos:
Added `split` and `split_by_length` methods on `compas.geometry.Polyline` to complement the existing divide methods. These methods return a list of Polylines, similar to the split option in Rhino when dividing a curve. Also, updated the docstrings of the divide methods with examples 😊

### Checklist


nmaslarinos:
I have also developed a `break_polyline_at_point` method that opens a polyline on a specified point if closed or splits it in two. I could add it to this pull request since it's all about Polyline methods, or create a new one

src/compas/geometry/primitives/polyline.py
```diff
@@ -352,14 +352,21 @@ def divide_polyline(self, num_segments):
         list
             list[:class:`~compas.geometry.Point`]
 
+        Examples
+        --------
+        >>> polyline = Polyline([(0, 0, 0), (1, 1, 0), (2, 3, 0), (4, 4, 0), (5, 2, 0)])
+        >>> divided_polylines = polyline.divide(3)
+        >>> divided_polyline
```
tomvanmele, line 351: this variable doesn't seem to be defined

src/compas/geometry/primitives/polyline.py
```diff
@@ -379,6 +386,18 @@ def divide_polyline_by_length(self, length, strict=True, tol=1e-06):
         The points of the new polyline are constrained to the segments of the old polyline.
         However, since the old points are not part of the new set of points, the geometry of the polyline will change.
 
+        Examples
+        --------
+        >>> polyline = Polyline([(0, 0, 0), (1, 1, 0), (2, 3, 0), (4, 4, 0), (5, 2, 0)])
+        >>> divided_polylines = polyline.divide_by_length(3)
+        >>> divided_polyline
```
tomvanmele, line 385: same here?

src/compas/geometry/primitives/polyline.py
```diff
@@ -379,6 +386,18 @@ def divide_polyline_by_length(self, length, strict=True, tol=1e-06):
...
+        [Point(0.000, 0.000, 0.000), Point(1.709, 2.418, 0.000), Point(4.051, 3.898, 0.000)]
+
+        >>> polyline = Polyline([(0, 0, 0), (1, 1, 0), (2, 3, 0), (4, 4, 0), (5, 2, 0)])
+        >>> divided_polylines = polyline.divide_by_length(3, strict=False)
+        >>> divided_polyline
```
tomvanmele, line 390: and here...

src/compas/geometry/primitives/polyline.py
```diff
@@ -412,6 +431,98 @@ def divide_polyline_by_length(self, length, strict=True, tol=1e-06):
 
     divide_by_length = divide_polyline_by_length
 
+    def split_polyline_by_length(self, length, strict=True):
```
tomvanmele, line 426: what is the difference between split and divide?

tests/compas/geometry/test_polyline.py
```diff
@@ -196,6 +196,140 @@ def test_polyline_split_at_corners(coords, input, expected):
     assert expected == Polyline(coords).split_at_corners(input)
 
 
+@pytest.mark.parametrize(
```
tomvanmele, line 186: could you add a few boundary cases, and tests that are expected to fail?

src/compas/geometry/primitives/polyline.py
nmaslarinos: Sorry about those, I only now realise that for some reason `invoke testdocs` does not perform a check of the files in src, but of the ones in tests

src/compas/geometry/primitives/polyline.py
nmaslarinos: The difference is that divide returns a list of the points that the polyline is divided into. Split returns a list of Polylines. It breaks the polyline into different parts, without disregarding it's geometry.

tomvanmele:
if i understand correctly, the division points are placed based on the length of the original polyline. does this mean the length of the segments of the new polyline is somewhat random?

perhaps this should be clearly stated in the docs, because users might be expecting to get a polyline with segments of the specified length as the result...

tomvanmele:
> if i understand correctly, the division points are placed based on the length of the original polyline. does this mean the length of the segments of the new polyline is somewhat random?
> 
> perhaps this should be clearly stated in the docs, because users might be expecting to get a polyline with segments of the specified length as the result...

ha no sorry, i see now that i indeed misunderstood

tomvanmele:
LGTM
if you resolve the merge conflicts and fix the docstrings this is good to go...

nmaslarinos:
I added some boundary cases tests in the `test_polyline`, fixed the docstrings which should now work, and added some additional test to the input length (to accept only values larger than zero and smaller than the length of the polyline)

tomvanmele:
@nmaslarinos almost there :) your test file has some formatting issues (https://github.com/compas-dev/compas/actions/runs/5297432096/jobs/9589200519?pr=1152). other than that it is good to go...

nmaslarinos:
Seems like after formating with black, there is still an [error on the mac-os 3.7 build](https://github.com/compas-dev/compas/actions/runs/5309441021/jobs/9610099589) __FAILED tests/compas/datastructures/test_graph.py::test_graph_networkx_conversion - ModuleNotFoundError: No module named '_bz2'__ . Any idea what this could be?

tomvanmele:
no worries this is a github hickup

tomvanmele:
i will rerun and the problem probably goes away (had the same a few days ago)

nmaslarinos:
I tried to do that, but for the love of me I couldn't find the rerun button 🙄

tomvanmele (approved):

PR #1130, opened 2023-03-23
bug fix on `is_polygon_in_polygon_xy`

baehrjo:
The is_polygon_in_polygon method was sometimes mistaking returning polygon2 was inside polygon1 while it wasn't. Issue occurs when edge intersection check is required inside is_polygon_in_polygon_xy to 'help decide'. Lines are first created, both for the contour polygon and the polygon to check on. Here is part of the code :
for j in range(len(polygon2)):
        line = [polygon2[-j], polygon2[j - 1]]
should be  [... , polygon2[-j - 1]] to accurately loop through all the edges of polygon2

### What type of change is this?

- [x] Bug fix in a **backwards-compatible** manner.
- [ ] New feature in a **backwards-compatible** manner.
- [ ] Breaking change: bug fix or new feature that involve incompatible API changes.
- [ ] Other (e.g. doc update, configuration, etc)

### Checklist

- [x] I added a line to the `CHANGELOG.md` file in the `Unreleased` section under the most fitting heading (e.g. `Added`, `Changed`, `Removed`).
- [x] I ran all tests on my computer and it's all green (i.e. `invoke test`).
- [x] I ran lint on my computer and there are no errors (i.e. `invoke lint`).
- [ ] I added new functions/classes and made them available on a second-level import, e.g. `compas.datastructures.Mesh`.
- [x] I have added tests that prove my fix is effective or that my feature works.
- [x] I have added necessary documentation (if appropriate)

PR #1153, opened 2023-06-08
Predicates2 new methods addition

baehrjo:
I added new predicates in `predicates_2` for both point/polygon(XY) and polygon/polygon(XY): `polygon_to_polygon_relationship_xy`, `point_to_polygon_relationship_xy`, `is_intersection_polygon_polygon_xy`
Full documentation provided in the methods comments. There are on purpose overlaps in what the methods are doing, for instance 'polygon_to_polygon_relationship_xy' first checks for intersection, then for containment, and if not deducts that the polygon is outside instead of calling separate methods. Separate methods ('is_polygon_in_polygon_xy', 'is_intersection_polygon_polygon_xy', ...) are still useful when looking for a True/False answer only.

- [x ] New feature in a **backwards-compatible** manner.

### Checklist

- [x ] I added a line to the `CHANGELOG.md` file in the `Unreleased` section under the most fitting heading (e.g. `Added`, `Changed`, `Removed`).
- [x ] I have added tests that prove my fix is effective or that my feature works.


src/compas/geometry/predicates/predicates_2.py
```diff
@@ -383,7 +386,7 @@ def is_polygon_in_polygon_xy(polygon1, polygon2):
         for i in range(len(polygon1)):
             line = [polygon1[-i], polygon1[-i - 1]]
             for j in range(len(polygon2)):
-                line_ = [polygon2[-j], polygon2[j - 1]]
+                line_ = [polygon2[-j], polygon2[-j - 1]]
```
tomvanmele, line 369: looking at this code now, would it not be much simpler (and more readable) to do the following:

```python
p1 = len(polygon1)
p2 = len(polygon2)
for i in range(-1, p1 - 1):
    line1 = polygon[i], polygon[i + 1]
    for j in range(-1, p2 - 1):
        line2 = polygon[j], polygon[j + 1]
```

src/compas/geometry/predicates/predicates_2.py
```diff
@@ -392,6 +395,118 @@ def is_polygon_in_polygon_xy(polygon1, polygon2):
...
+        0 if polygon1 intersects polygon2.
+        -1 if polygon1 is outside polygon2.
+        -2 if polygon1 contains polygon2 (polygon2 is inside polygon1).
+    """
+    for i in range(len(polygon1)):
+        line1 = [polygon1[i], polygon1[i - 1]]
+        for j in range(len(polygon2)):
+            line2 = [polygon2[j], polygon2[j - 1]]
+            if is_intersection_segment_segment_xy(line1, line2):
+                return 0
```
tomvanmele, line 405: should we not replace this by a call to `is_intersection_polygon_polygon_xy`?

src/compas/geometry/predicates/predicates_2.py
```diff
@@ -392,6 +395,118 @@ def is_polygon_in_polygon_xy(polygon1, polygon2):
         return False
 
 
+def polygon_to_polygon_relationship_xy(polygon1, polygon2):
+    """Determine if a polygon (polygon1) is inside, intersects, is outside or contains another polygon (polygon2) on the XY-plane.
```
tomvanmele, line 379: is it required that the polygons are convex?

src/compas/geometry/predicates/predicates_2.py
```diff
@@ -392,6 +395,118 @@ def is_polygon_in_polygon_xy(polygon1, polygon2):
...
+        line1 = [polygon1[i], polygon1[i - 1]]
+        for j in range(len(polygon2)):
+            line2 = [polygon2[j], polygon2[j - 1]]
+            if is_intersection_segment_segment_xy(line1, line2):
+                return 0
+    for pt in polygon1:
+        if is_point_in_polygon_xy(pt, polygon2):
+            return 1
+    if is_point_in_polygon_xy(polygon2.centroid, polygon1):
+        return -2
```
tomvanmele, line 410: why the different approach here?

src/compas/geometry/predicates/predicates_2.py
```diff
@@ -392,6 +395,118 @@ def is_polygon_in_polygon_xy(polygon1, polygon2):
...
+        if is_point_on_segment_xy(point, segment, tolerance):
+            return 0
+        elif y > min(y1, y2):
+            if y <= max(y1, y2):
+                if x <= max(x1, x2):
+                    if y1 != y2:
+                        xinters = (y - y1) * (x2 - x1) / (y2 - y1) + x1
+                    if x1 == x2 or x <= xinters:
+                        flag = -flag
+    return flag
```
tomvanmele, line 449: this seems to be a combination of `is_point_on_polyline_xy` and `is_point_in_polygon_xy`? why not use those?

src/compas/geometry/predicates/predicates_2.py
baehrjo: (see #1130 )

src/compas/geometry/predicates/predicates_2.py
baehrjo: It's not required, the methods works for any (XY) polygon. 

In the is_polygon_in_polygon method, there's a convexity check at the beginning because if both polygons are convex then checking that all points of one polygon are contained in the other is enough and faster, while without that convexity, we have to check that both all points are inside and no edges intersect.

In polygon_to_polygon_relationship, we need the intersection check anyway as we deduct information from it for other relationship than containment, that's why I don't check convexity at the begenning. Polygon_to_polygon_relationship first checks for intersection, then for containment, and if not deducts that the polygon is outside instead of calling separate methods. Separate methods ('is_polygon_in_polygon_xy', 'is_intersection_polygon_polygon_xy', ...) are still useful when looking for a True/False answer only but calling them inside polygon_to_polygon_relationship would run several times the same pieces of code.

src/compas/geometry/predicates/predicates_2.py
baehrjo: You mean different because of calling other existing methods inside the loop here ?

tomvanmele:
@baehrjo #1130 is merged. please resolve the conflicts by merging main into your PR branch so we can get this one done as well...

tomvanmele:
@baehrjo i think you still need to merge main back into your branch, or rebase it onto main...

baehrjo:
Hello,

Sorry for the delay, done!

Best regards,

Joelle

On 24 Oct 2023, at 20:57, Tom Van Mele ***@***.***> wrote:



@baehrjo<https://github.com/baehrjo> i think you still need to merge main back into your branch, or rebase it onto main...

PR #1160, opened 2023-06-19
lock python 3.7

Licini:
https://github.com/actions/setup-python/issues/682

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -15,7 +15,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        python: ['3.7.16', '3.8', '3.9', '3.10']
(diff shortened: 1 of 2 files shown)


gonzalocasas:
interesting.. why is that?

tomvanmele:
> interesting.. why is that?

there is a problem with the recent builds of Python 3.7 on OSX and it makes the tests fail because `_bz2` cannot be found. seems to be a known problem and a fix is in the works but until then we will have to skip the tests for it...

Licini:
I recommend we skip testing for 3.7 for now. Locking it to specific version does not really work if we want to use the version matrix setup. The latest sub-version of 3.7 for windows is 3.7.9, however this one only supports ubuntu 20, while the latest 22 is used. Very likely they will just do a 3.7.18 patching this problem soon, we can add it back by then.

tomvanmele (approved):

PR #1163, opened 2023-07-04
Geometry Refactoring

tomvanmele:

### Checklist


tomvanmele:
I assume some explanations are required :)

tomvanmele:
Maybe short disclaimer... Many of the points below are how (i think) it should be. Not all of this is entirely implemented yet.

**General**

* Removed ``Primitive`` as a type.
* Two base classes are ``Geometry`` and ``Transformation``
* ``Point``, ``Vector``, ``Plane``, ``Frame``, ``Quaternion``, ``Polygon``, ``Polyhedron`` are now "basic geometry objects".
* Basic geometry objects don't have a frame and are always explicitly defined.

**Curves**

*  ``Line``, ``Polyline``, ``Arc``, ``NurbsCurve`` are ``Curve`` objects
* ``Circle``, ``Ellipse``, ``Hyperbola``, ``Parabola`` are ``Conic`` curve objects.
* All curves are parametric and have a local coordinate system. They support ``point_at``, ``tangent_at``, ``normal_at``, ``frame_at``.
* The coordinate system of ``Line`` and ``Polyline`` cannot be changed and is always the world coordinate system, because their parametrisation is based on a base point and direction vector(s).
* The domain of all curves is ``0, 1`` with ``point_at(0)`` and ``point_at(1)`` being the start and end point.
* Closed curves will be convertible to ``Polygon``.

**Surfaces**

Haven't started with those yet...

**Shapes**

* Shapes are parametric solids.
* Like curves, they are implicitly defined wrt to a local coordinate system.
* They have an explicit discretised representation though the generation of vertices and faces with a specific resolution.
* They support conversion to ``Mesh``, ``Brep`` (soon), ``Frep`` (less soon), ``Polyhedron``.
* Boolean operations are supported via those formats, but no longer directly, since this required a lot of implied and opaque behaviour in the background and since the results of those operations were in fact no longer part of the shape family.

**Polygon and Polyhedron**

* Can be discretised representations of curves and shapes.
* Can be objects in their own right.
* Both planar and skewed geometries are supported.
* Support boolean operations.
* 2D boolean operations are pluggables and some plugins are already added using shapely (directly in the core package since shapely is as standard as numpy. to be discussed how to formalise this)

tomvanmele:
@gonzalocasas @chenkasirer @Licini i would (review and) merge as is and add more docs and tests later. otherwise this PR will be never-ending...

src/compas/geometry/bbox/bbox_numpy.py
```diff
@@ -155,41 +139,123 @@ def oriented_bounding_box_xy_numpy(points):
...
+    # frame = mean, vectors[0], vectors[1]
+    # points = world_to_local_coordinates_numpy(frame, points)
+
+    rect1 = minimum_area_rectangle_xy(points)
+    # area1 = (rect1[1][0] - rect1[0][0]) * (rect1[3][1] - rect1[0][1])
+    # rect2 = bounding_box(points)[:4]
+    # area2 = (rect2[1][0] - rect2[0][0]) * (rect2[3][1] - rect2[0][1])
+    # rect = rect1 if area1 < area2 else rect2
+    # rect = [[x, y, 0.0] for x, y in rect1]
+    # bbox = local_to_world_coordinates_numpy(frame, rect)
```
chenkasirer: Dead code?

src/compas/geometry/curves/nurbs.py
```diff
@@ -13,7 +13,9 @@
 
 @pluggable(category="factories")
 def new_nurbscurve(cls, *args, **kwargs):
-    raise NotImplementedError
+    curve = object.__new__(NurbsCurve)
+    curve.__init__(*args, **kwargs)
+    return curve
```
chenkasirer, line 18: wait can we do that? I thought that didn't work..

tomvanmele:
?

src/compas/geometry/shapes/box.py
```diff
@@ -99,35 +99,65 @@ class Box(Shape):
...
+
+    def __repr__(self):
+        return "Box(xsize={0!r}, ysize={1!r}, zsize={2!r}, frame={3!r})".format(
+            self.xsize, self.ysize, self.zsize, self.frame
+        )
+
+    def __len__(self):
+        return 4
+
+    def __getitem__(self, key):
```
chenkasirer, line 128: maybe it's worth somehow mentioning this in the main docstring, otherwise I wouldn't expect Box to be subscriptable.

src/compas/geometry/shapes/box.py
```diff
@@ -519,10 +486,77 @@ def from_diagonal(cls, diagonal):
...
+    #     """
+    #     self.frame.transform(transformation)
+    #     # Always local scaling, non-uniform scaling based on frame not yet considered.
+    #     Sc, _, _, _, _ = transformation.decomposed()
+    #     scalex = Sc[0, 0]
+    #     scaley = Sc[1, 1]
+    #     scalez = Sc[2, 2]
+    #     self.xsize *= scalex
+    #     self.ysize *= scaley
+    #     self.zsize *= scalez
```
chenkasirer, line 539: dead code?

chenkasirer:
> @gonzalocasas @chenkasirer @Licini i would (review and) merge as is and add more docs and tests later. otherwise this PR will be never-ending...

Agree. I will review today

src/compas/geometry/shapes/shape.py
```diff
@@ -2,147 +2,312 @@
...
+        data : dict
+            The data dictionary.
+
+        Returns
+        -------
+        :class:`~compas.geometry.Shape`
+            The constructed shape.
+
+        """
+        return cls(**data)
```
chenkasirer: Isn't this a redefinition of the `Data` interface?

src/compas/geometry/shapes/sphere.py
```diff
@@ -70,10 +69,12 @@ class Sphere(Shape):
         "required": ["frame", "radius"],
     }
 
-    def __init__(self, frame=None, radius=None, **kwargs):
+    def __init__(self, frame=None, point=None, radius=None, **kwargs):
         super(Sphere, self).__init__(frame=frame, **kwargs)
         self._radius = 1.0
         self.radius = radius
+        if point:
+            self.frame.point = point
```
chenkasirer, line 77: So if I get it right, a sphere is defined by a point and a radius. the point is defined in a certain coordinate system given by the frame. So isn't setting the given point as the origin of the coordinate-system messes with this distinction?

Like this if transforming the Sphere means transforming its frame, then the center point which should be defined in local coordinates would also change. right?

src/compas/geometry/surfaces/surface.py
```diff
@@ -5,13 +5,17 @@
...
 from compas.plugins import pluggable
 from compas.utilities import linspace
 
 
 @pluggable(category="factories")
 def new_surface(cls, *args, **kwargs):
-    raise NotImplementedError
+    surface = object.__new__(cls)
+    surface.__init__(*args, **kwargs)
+    return surface
```
chenkasirer, line 19: I don't really get how this works. If there's no plugins discovered wouldn't it just raise a `NotImplementedError`?

src/compas/geometry/polygon.py
```diff
@@ -166,25 +178,26 @@ def centroid(self):
...
+        #     p1 = points[i]
+        #     p2 = points[i + 1]
+        #     u = [p1[_] - o[_] for _ in range(3)]  # type: ignore
+        #     v = [p2[_] - o[_] for _ in range(3)]  # type: ignore
+        #     w = cross_vectors(u, v)
+        #     a2 += sum(w[_] ** 2 for _ in range(3)) ** 0.5
+        #     normals.append(w)
+        # n = [sum(axis) / a2 for axis in zip(*normals)]
+        # n = Vector(*n)
+        # return n
```
chenkasirer, line 195: remove?

src/compas/geometry/pointcloud.py
```diff
@@ -143,7 +149,15 @@ def from_ply(cls, filepath):
...
 
         """
-        pass
+        points = []
+        # normals = []
+        ply = PLY(filepath)
+        for vertex in ply.reader.vertices:  # type: ignore
+            points.append([vertex["x"], vertex["y"], vertex["z"]])
+            # if "nx" in vertex and "ny" in vertex and "nz" in vertex:
+            #     normals.append([vertex["nx"], vertex["ny"], vertex["nz"]])
```
chenkasirer: remove?

src/compas/geometry/surfaces/surface.py
```diff
@@ -505,3 +581,34 @@ def intersections_with_curve(self, curve):
...
+    #     v : float
+    #     du : int, optional
+    #     dv : int, optional
+
+    #     Returns
+    #     -------
+    #     :class:`~compas.geometry.NurbsSurface`
+
+    #     """
+    #     raise NotImplementedError
```
chenkasirer, line 713: remove?

src/compas/geometry/surfaces/spherical.py
```diff
@@ -0,0 +1,378 @@
...
+            raise ValueError("At least three points are required to construct a sphere.")
+
+        if len(points) == 3:
+            return cls.from_three_points(*points)
+
+        from compas.geometry import bestfit_sphere_numpy
+
+        center, radius = bestfit_sphere_numpy(points)
+        return cls(radius, frame=Frame(center, [1, 0, 0], [0, 1, 0]))
+
```
chenkasirer, line 174: maybe add a `from_sphere()`?

src/compas/geometry/surfaces/surface.py
```diff
@@ -355,18 +476,18 @@ def transform(self, T):
...
+        # T[2, 2] = 1
+        # T[3, 3] = 1
+
+        # T[0, 3] = 0
+        # T[1, 3] = 0
+        # T[2, 3] = 0
+
+        # T[3, 0] = 0
+        # T[3, 1] = 0
+        # T[3, 2] = 0
```
chenkasirer: remove?

src/compas/geometry/surfaces/cylindrical.py
```diff
@@ -0,0 +1,220 @@
...
+
+        Returns
+        -------
+        :class:`compas.geometry.CylindricalSurface`
+            A cylindrical surface.
+
+        """
+        circle = Circle.from_three_points(a, b, c)
+        return cls(circle.radius, frame=circle.frame)
+
```
chenkasirer, line 147: maybe add `from_cylinder`?

src/compas/geometry/surfaces/conical.py
```diff
@@ -0,0 +1,190 @@
...
+    #     :class:`compas.geometry.Frame`
+    #         The frame at the given parameters.
+
+    #     """
+    #     u = u * 2 * pi
+    #     point = self.point_at(u, v)
+    #     zaxis = self.normal_at(u)
+    #     yaxis = self.frame.zaxis
+    #     xaxis = yaxis.cross(zaxis)
+    #     return Frame(point, xaxis, yaxis)
```
chenkasirer: ?

chenkasirer (approved):
LGTM

Note: I went though the commits themselves and made my comments so maybe some of comments are not relevant cause the code changed in a later commit.

src/compas/geometry/shapes/box.py
tomvanmele: was rather thinking of removing these accessors from most objects, only keeping them for fundamental geometries like point, vector, plane, frame, etc.

chenkasirer: even better ;)

src/compas/geometry/shapes/sphere.py
tomvanmele: the idea is that the sphere is always centered at the origin of its local coordinate system. so changing the center of the sphere and the origin of the coordinate system are equivalent...

src/compas/geometry/polygon.py
tomvanmele: yes will remove all commented sections

src/compas/geometry/surfaces/spherical.py
tomvanmele: it is rather the other way around. the surfaces form the underlying geometry of shapes. so the idea would be to redefine the shapes such that they use an underlying surface geometry or geometries to define the shape. a bit like how breps work...

src/compas/geometry/shapes/sphere.py
chenkasirer: gotcha. then I guess what confuses me is that you can provide both a point and a frame, and the frame's origin gets implicitly overridden when you do. But maybe it's sufficient to just make that clear in the docstring (missing `point` there btw).

PR #1165, opened 2023-07-10
Offset plugin

nmaslarinos:
Added support for Shapely offset using a plugin implementation. 
There are a few things worth discussing about it, mainly concerning the different arguments that compas and shapely functions require. If we want to keep the existing compas functions, for backward compatibility, then we could add in the now-plugin functions and the in-class calls a few lines indication the different args a user can input depending on the backend they are using. 
Additionally, I rephrased the original shapely corner style options from 'miter, bevel, round' to 'sharp, chamfer, sharp' to match the ones in Rhino. Seeing as most users of compas will be more familiar with Rhino's terminology and the potential addition of rhino plugins for them, I thought it's more fitting! 
Let me know your thoughts :)

### Checklist


jf---:
@tomvanmele || @gonzalocasas any thoughts on this PR? 
Since its so tied to `shapely`, perhaps a `pluggable` would be fitting?

tomvanmele:
@nmaslarinos should move it out of "Draft" when it is ready so we can run the CI workflows on it...

tomvanmele:
is this still relevant?

PR #1169, opened 2023-07-21
Tree

Licini:
Adding the tree class. @tomvanmele for serialization strategy, I ended up opting for directed nested dictionaries. This way it uses least amount of memory and not needing any types of uuid referencing. I'm aware the docstrings is not formatted entirely correctly, but besides that let me know how does the class structure itself look like. thx!

### Checklist


src/compas/datastructures/tree/tree.py
```diff
@@ -0,0 +1,259 @@
...
+    is_branch: bool
+        True if the node is a branch node of the tree.
+    acestors: generator[:class:`~compas.datastructures.TreeNode`]
+        A generator of the acestors of the tree node.
+    descendants: generator[:class:`~compas.datastructures.TreeNode`]
+        A generator of the descendants of the tree node.
+
+    """
+
+    def __init__(self, name: str = None, attributes: dict = None):
```
tomvanmele: you can't use typing in the core compas lib...

src/compas/datastructures/tree/tree.py
```diff
@@ -0,0 +1,259 @@
...
+    def remove(self, node: TreeNode):
+        """Remove a child node from this node."""
+        self._children.remove(node)
+        node._parent = None
+        node._tree = None
+        if self.tree:
+            self.tree.nodes.remove(node)
+
+    @property
+    def acestors(self):
```
gonzalocasas: Typo

src/compas/datastructures/tree/tree.py
```diff
@@ -0,0 +1,259 @@
...
+
+    @property
+    def descendants(self):
+        for child in self.children:
+            yield child
+            for descendant in child.descendants:
+                yield descendant
+
+    def traverse(self):
+        """Traverse the tree from this node."""
```
gonzalocasas: Traverse can be done many ways, is it ok to have a single top level traverse method and no alternative for choice which traversal strategy?

src/compas/datastructures/tree/tree.py
```diff
@@ -0,0 +1,259 @@
...
+        if self.root is None:
+            raise ValueError("The tree has no root node, use add_root() first.")
+        else:
+            parent.add(node)
+
+    @property
+    def nodes(self):
+        return self._nodes
+
+    def remove_root(self):
```
gonzalocasas: I would make this method private (underscored), otherwise the API surface is confusing.

src/compas/datastructures/tree/tree.py
```diff
@@ -0,0 +1,259 @@
...
+
+    @property
+    def leaves(self):
+        for node in self.nodes:
+            if node.is_leaf:
+                yield node
+
+    def __repr__(self):
+        nodes = len(self.nodes)
+        branches = sum(1 for node in self.nodes if node.is_branch)
```
gonzalocasas: This feels like an extremely expensive operation, for a statement that will be triggered inadvertently without considering its cost. I would remove all the summing of branches and leaves.

gonzalocasas (changes requested):
I quickly glanced over it, I think it's a great idea to add trees in general! Here's a first pass of comments!
Other random comments would include considering whether adding a more sophisticated type of tree would make sense (red-black, for example?)

src/compas/datastructures/tree/tree.py
Licini: removed them

src/compas/datastructures/tree/tree.py
Licini: Added pre-order and post-order, things like in-order are for binary tree so not included here

src/compas/datastructures/tree/tree.py
Licini: makes sense

src/compas/datastructures/tree/tree.py
Licini: removed : )

Licini:
> I quickly glanced over it, I think it's a great idea to add trees in general! Here's a first pass of comments! Other random comments would include considering whether adding a more sophisticated type of tree would make sense (red-black, for example?)

Thx for the feedbacks! For more specific type of trees like binary tree, red black tree etc we can add them in the same folder inheriting from this class. For instance for binary tree we'd add left/right property, in-order traverse, for red black tree we add color property and corresponding algorithms, but it is very important that the base tree class does not create any limitation on later more specific implementations, as well as not creating any significant computational overhead due to the way we structure it. The primary use cases we have in mind for now are things like scene objects graph (like spatial hierarchy in ifc files), classes inheritance tree etc, rather high-level stuff.

btw I'm curious to know your opinion on the serialization strategy, is it good the way it is now as nested dicts or should we do flat node list + uuid identifiers?

src/compas/datastructures/tree/tree.py
tomvanmele: please add parameter to the docs and document, return, raise, ...

src/compas/datastructures/tree/tree.py
```diff
@@ -0,0 +1,264 @@
...
+    Parameters
+    ----------
+    name: str, optional
+        The name of the tree ndoe.
+    attributes: dict[str, Any], optional
+        User-defined attributes of the datastructure.
+
+    Attributes
+    ----------
+    name: str
```
tomvanmele: there needs to be a space after the colon: `name : str`

tomvanmele: see other items as well

src/compas/datastructures/tree/tree.py
```diff
@@ -0,0 +1,264 @@
...
+
+from compas.datastructures import Datastructure
+
+
+class TreeNode(object):
+    """A node of a tree data structure.
+
+    Parameters
+    ----------
+    name: str, optional
```
tomvanmele: space after colon

src/compas/datastructures/tree/tree.py
```diff
@@ -0,0 +1,264 @@
...
+
+    @property
+    def children(self):
+        return self._children
+
+    @property
+    def tree(self):
+        return self._tree
+
+    def add(self, node):
```
tomvanmele, line 107: document parameters and raise

src/compas/datastructures/tree/tree.py
```diff
@@ -0,0 +1,264 @@
...
+        """Add a child node to this node."""
+        if not isinstance(node, TreeNode):
+            raise TypeError("The node is not a TreeNode object.")
+        self._children.add(node)
+        node._parent = self
+        node._tree = self.tree
+        if self.tree:
+            self.tree.nodes.add(node)
+
+    def remove(self, node):
```
tomvanmele, line 132: document parameters

src/compas/datastructures/tree/tree.py
```diff
@@ -0,0 +1,264 @@
...
+            for child in self.children:
+                for node in child.traverse(strategy):
+                    yield node
+            yield self
+        else:
+            raise ValueError("Unknown traversal strategy: {}".format(strategy))
+
+
+class Tree(Datastructure):
+    """A tree data structure.
```
tomvanmele: perhaps a short overview of the most important characteristics of the data structure would be useful.

src/compas/datastructures/tree/tree.py
```diff
@@ -0,0 +1,264 @@
...
+
+    Parameters
+    ----------
+    name: str, optional
```
tomvanmele: space

src/compas/datastructures/tree/tree.py
```diff
@@ -0,0 +1,264 @@
...
+    <Tree with 4 nodes, 1 branches, and 2 leaves>
+    >>> tree.print()
+    <TreeNode root>
+        <TreeNode branch>
+            <TreeNode leaf2>
+            <TreeNode leaf1>
+
+    """
+
+    JSONSCHEMA = {}
```
tomvanmele: please add

src/compas/datastructures/tree/tree.py
```diff
@@ -0,0 +1,264 @@
...
+        self.name = data["name"]
+        self.attributes = data["attributes"]
+
+        def node_from_data(data):
+            node = TreeNode(data["name"], data["attributes"])
+            for child in data["children"]:
+                node.add(node_from_data(child))
+            return node
+
+        self.add_root(node_from_data(data["root"]))
```
tomvanmele: why not just make this the responsibility of the nodes?

src/compas/datastructures/tree/tree.py
```diff
@@ -0,0 +1,264 @@
...
+
+    @property
+    def root(self):
+        return self._root
+
+    def add_root(self, node):
```
tomvanmele: why is this a separate method? and not just `root=None` in `add` with an error raised if a root already exists? or something like that...

src/compas/datastructures/tree/tree.py
Licini: Indeed, in that case should TreeNode also inherent from `Data` class ??

tomvanmele: yes

jf---:
Greenwashing compas with trees ;)

Licini:
@tomvanmele How does it look now?
Besides the requested changes, one more thing:
- I changed `Tree.nodes` attribute to be dynamically inferred. Because maintaining this list separately is a bit error-prone especially when you start to trim branches or merge trees (you would need always traverse the given branch and update the node list every-time you add or remove). The decision comes with a bit computational penalty when you print the tree. But since we are not likely to operate on super huge trees yet, I'd vote for keeping the base class simple, and later have more advanced implementations when performance issue arises.

src/compas/datastructures/tree/tree.py
```diff
@@ -0,0 +1,373 @@
...
+            yield this
+            this = this.parent
+
+    @property
+    def descendants(self):
+        for child in self.children:
+            yield child
+            for descendant in child.descendants:
+                yield descendant
+
```
tomvanmele, line 162: this implies the children are returned in "depth-first" order, no? either way, could you add a comment about this in the attributes section of the class docstring?

src/compas/datastructures/tree/tree.py
```diff
@@ -0,0 +1,373 @@
...
+    def traverse(self, strategy="preorder"):
+        """
+        Traverse the tree from this node.
+
+        Parameters
+        ----------
+        strategy : str, optional
```
tomvanmele: `strategy : {"preorder", "postorder"}, optional`

src/compas/datastructures/tree/tree.py
```diff
@@ -0,0 +1,373 @@
...
+            The traversal strategy. Options are ``"preorder"`` and ``"postorder"``.
+            Default is ``"preorder"``.
```
tomvanmele: this can then just be "The traversal strategy."

src/compas/datastructures/tree/tree.py
```diff
@@ -0,0 +1,373 @@
...
+            for child in self.children:
+                for node in child.traverse(strategy):
+                    yield node
+        elif strategy == "postorder":
+            for child in self.children:
+                for node in child.traverse(strategy):
+                    yield node
+            yield self
+        else:
+            raise ValueError("Unknown traversal strategy: {}".format(strategy))
```
tomvanmele: preorder and postorder is the same as depth-first, breadth-first?

src/compas/datastructures/tree/tree.py
```diff
@@ -0,0 +1,373 @@
...
+
+    Attributes
+    ----------
+    name : str
+        The name of the datastructure.
+    attributes : dict[str, Any]
+        User-defined attributes of the datastructure.
+    root : :class:`~compas.datastructures.TreeNode`
+        The root node of the tree.
+    nodes : list[:class:`~compas.datastructures.TreeNode`]
```
tomvanmele: this is a generator, not a list

src/compas/datastructures/tree/tree.py
```diff
@@ -0,0 +1,373 @@
...
+            If the node is already part of another tree.
+            If the supplied parent node is not part of this tree.
+            If the tree already has a root node, when trying to add a root node.
+
+        """
+        if not isinstance(node, TreeNode):
+            raise TypeError("The node is not a TreeNode object.")
+
+        if node.parent:
+            raise ValueError("The node is already part of another tree, remove it from that tree first.")
```
tomvanmele: can also be THIS tree, no?

tomvanmele (commented):
maybe you should also add a few tests :)

src/compas/datastructures/tree/tree.py
Licini: I thought they are, but just realized I got it confused too... Turned out pre and post order are both subcategory of depth-first, the difference is whether you visit current node first or not. None of them are breath-first. Will make some adjustments here

src/compas/datastructures/tree/tree.py
```diff
@@ -159,14 +160,18 @@ def descendants(self):
...
         """
         Traverse the tree from this node.
 
         Parameters
         ----------
-        strategy : str, optional
-            The traversal strategy. Options are ``"preorder"`` and ``"postorder"``.
+        strategy : {"depthfirst", "breadthfirst"}, optional
+            The traversal strategy.
+            Default is ``"depthfirst"``.
```
tomvanmele: in my opinion you don't have to mention the default. this is already clear from line 163

src/compas/datastructures/tree/tree.py
Licini: @tomvanmele ok now there are two arguments `strategy` and `order`, strategy refers to depth first of breath first, and order is pre or post order (only for depth first). This should better aligns the terminologies.

src/compas/datastructures/tree/tree.py
Licini: fixed : )

src/compas/datastructures/tree/tree.py
```diff
@@ -359,6 +371,76 @@ def leaves(self):
...
 
+    def traverse(self, strategy="depthfirst", order="preorder"):
+        """
+        Traverse the tree from the root node.
+
+        Parameters
+        ----------
+        strategy : {"depthfirst", "breadthfirst"}, optional
+            The traversal strategy.
+            Default is ``"depthfirst"``.
```
tomvanmele: same here. no mention of default needed in my opinion

tomvanmele (approved):

src/compas/datastructures/tree/tree.py
Licini: Sorry for the confusion, I adjusted a bit the sentence here, the idea is that if a node is already added somewhere (a parent link already exists, no matter it is this tree or another), we can ask user to explicitly remove that link first. Another option would be to not care about it and directly overwrite this link, which one you think is better?

Licini:
hey @gonzalocasas could you take another look at the current state of this PR, it needs your approval

src/compas/datastructures/tree/tree.py
```diff
@@ -379,11 +377,9 @@ def traverse(self, strategy="depthfirst", order="preorder"):
         ----------
         strategy : {"depthfirst", "breadthfirst"}, optional
             The traversal strategy.
-            Default is ``"depthfirst"``.
```
tomvanmele: there can't be a blank line in parameter lists

src/compas/datastructures/tree/tree.py
```diff
@@ -168,11 +168,9 @@ def traverse(self, strategy="depthfirst", order="preorder"):
...
```
tomvanmele: same here

src/compas/datastructures/tree/tree.py
Licini: oops.... fixed

tomvanmele (approved):

Licini:
@gonzalocasas Pin.... : ) Could you approve this so it can be merged?

PR #1176, opened 2023-08-16
Introduce the compas.tolerance module

tomvanmele:
Add general precision handling, differentiating between distance precision, angular precision, and parametric precision.

Reference: https://dev.opencascade.org/doc/refman/html/class_precision.html

### Checklist


tomvanmele:
@jf--- how about something like this?

tomvanmele:
added an example of how it could be used in the pointcloud class. would of course have to be propagated all over the code base...

tomvanmele:
https://www.electronicdesign.com/technologies/test-measurement/article/21802799/the-difference-between-accuracy-and-precision

should we use accuracy, precision or tolerance?

tomvanmele:
i propose we replace ``tol=...`` (or similar) in function parameters with ``precision=None`` and add in the function body:
```python
if precision is None:
    precision = Precision().confusion

```
and similar for the other precision values...

tomvanmele:
not sure if precision is the right term though

https://www.electronicdesign.com/technologies/test-measurement/article/21802799/the-difference-between-accuracy-and-precision

src/compas/precision.py
```diff
@@ -0,0 +1,182 @@
...
+        The approximation precision.
+        The default is ``precision.confusion * 10``.
+    confusion : float
+        The confusion precision.
+        The default is 1e-7 for unit "M", and 1e-4 for unit "MM".
+    intersection : float
+        The intersection precision.
+        The default is ``precision.confusion / 100``.
+    squareconfusion : float
+        The square confusion precision.
```
tomvanmele: these names are based OCC terminology

tomvanmele:
tolerance seems to be more common than precision (perhaps because more appropriate for the context?)

src/compas/precision.py
tomvanmele: confusion could also be location or distance

yck011522:
Hello Tom, can you describe briefly what is the intended use of this module? Is it a wrapper for the OCC Precision class? From its docs, I can see that their main use case is for "... offers a set of functions defining precision criteria for use in conventional situations when comparing two numbers."

Since you questioned about the change of `tolerance` or ` tol` in functions to `precision`, I wonder if you mean by functions in the geometry module and the like, (such as https://github.com/compas-dev/compas/blob/49bfb6587bc585c21c15a5f8a21525a4bd1c78b5/src/compas/geometry/point.py#L405-L430)   

If you mean by these functions, I want to point out that in the context of mechanical engineering, and to my limited understanding, the terms accuracy and precision are both related to a **measurement**. Both are descriptions of a real-world property, mostly used to describe the quality of something, after some observations are made. On the other hand, tolerance is more of a description about the design intent, a number that determines what is acceptable. 

With this colloquial definition, the term `tolerance` fits better in many of the geometrical functions. Especially where the functions is asking the user how much error is acceptable. Furthermore, many of these geometrical functions are numerical and have no consideration for the input values's method of measurement, its random error, or statistical error. It is better to refrain from introducing the term precision into those functions because the term `precision` implies statistical measurement and the discussion about random error.

yck011522:
The term Precision according to ISO 5725-1 definition is the following:
https://en.wikipedia.org/wiki/Accuracy_and_precision#ISO_definition_(ISO_5725)

> According to ISO 5725-1, accuracy consists of trueness (proximity of measurement results to the true value) and precision (repeatability or reproducibility of the measurement).

The definition also changes slightly when the term is used in the discussion about measuring manufactured parts, vs the measurement method. However, in many of these context, it is a statistical discussion. So perhaps it is good to discuss what is the intent of this Precision module before going into a semantics debate. The way I see it is that the name `compas.Precision` appears to certain users as a class for modeling mechanical systems but turns out to be completely unrelated. If this module is only about wrapping OCC.Precision, as you have written [here](https://github.com/compas-dev/compas/pull/1176/files#diff-2ad18d84b332f2c81ac6f1677960958325bad629d4f09347bc14be447bda53ecR35), then it probably should not be so high up in the `compas dot` namespace.

tomvanmele:
hi @yck011522 

yes the purpose of the module is to centralise (and make more transparent) the use of precision/tolerance values throughout the code base. currently, these values are localised in the functions and are rather inconsistent. we have `compas.PRECISION` but this originates from a different time and purpose...

i agree that tolerance probably better fits. the name precision came partly from replacing `compas.PRECISION` and partly because of the name in OCC. the module is meant to be its own thing btw, not a wrapper for the precision class of OCC.

about its location in the compas namespace. i think it should be as high up as it can be, since it is a "system-wide setting" that potentially has an effect on all of the subpackages, and it should be clearly visible to all extension packages as well.

tomvanmele:
some interesting points of view
https://web.mit.edu/10.001/Web/Tips/Converge.htm
https://drake.mit.edu/doxygen_cxx/group__accuracy__and__tolerance.html
https://stackoverflow.com/questions/65909842/what-is-rtol-for-in-numpys-allclose-function

tomvanmele:
as discussed in those references, i guess we need both an absolute and relative tolerance, with the latter related to accuracy (aka the number of relevant digits), and the former related to "what is zero". based on the formula ``|true_value - computed_value| < rtol * true_value + atol``, there are two cases:
1. the true value is (almost) zero
2. the true value is not zero

consider for example the "point on line" check. the point is considered on the line if the distance from the point to the line is "close to zero": ``|0 - distance| < rtol * 0 + atol``

in this case, ``atol`` will be the decisive criterion since ``rtol * 0``. ``atol`` therefore means, "what is considered small enough to be zero".

when comparing for example the length of something to a target value (that is not zero) the situation is different: 
```
target = 1
length = 1.1
rtol = 0.001
atol = 0.0000001

|target - length| < rtol * target + atol
```
in this case, the length is considered close enough to the target if it deviates less than a certain percentage from the target, determined by ``rtol``

yck011522:
> some interesting points of view https://web.mit.edu/10.001/Web/Tips/Converge.htm https://drake.mit.edu/doxygen_cxx/group__accuracy__and__tolerance.html https://stackoverflow.com/questions/65909842/what-is-rtol-for-in-numpys-allclose-function

Cool. I agree that if the module is supposed to be a system-wide setting, it should be higher up. I like the separation of rtol vs atol. I think the reply in the StackOverflow reference with the two equations is rather straight forward to understand:

> Usecase 1: absolute tolerance (atol): absolute(a - b) <= atol
> Usecase 2: relative tolerance (rtol): absolute(a - b) <= rtol * absolute(b)

tomvanmele:
i based `Tolerance` on `Data` to allow for serialisation. i can imagine that work session files include a tolerance setting such that comparisons are handled on the same basis even after deserialising the file later on...

tomvanmele:
i also added a default instance of `Tolerance` at the end of the module, such that no overhead is created when doing many comparisons...

tomvanmele:
@gonzalocasas @chenkasirer @Licini how about this one?

PR #1180, opened 2023-09-15
Polyline new methods

nmaslarinos:
Added a couple of new methods to polyline. 
`polyline.parameter_at` takes as input a Point on the polyline and returns the equivalent parameter. 
`polyline.divide_at_corners` behaves similarly to `split_at_corners`, but instead of returning a list of Polylines, it returns a list of Points that justify the condition. In a way, it completes the duo of functions like divide/split and divide/split_by_length. See the image for the before and after calculations.

Also fixed a bug in `polyline.split_at_corners` where the angle between two consecutive segments was sometimes wrongly calculated.  See the image for the before and after calculations.
Artboard 1-100

### Checklist


nmaslarinos:
As a side-note, I was thinking that a lot of the polyline functions can also be used in the polygon, especially since now there is a method to convert between the two!

src/compas/geometry/curves/polyline.py
```diff
@@ -6,11 +6,11 @@
 
 from compas.geometry import allclose
 from compas.geometry import transform_points
-from compas.geometry import is_point_on_line
+from compas.geometry import is_point_on_line, is_point_on_polyline
 from compas.geometry import Point
 from compas.geometry import Line
 from compas.geometry import Frame
-from .curve import Curve
+from compas.geometry.curves.curve import Curve
```
tomvanmele: why change this?

tests/compas/geometry/test_curves_polyline.py
```diff
@@ -431,6 +449,16 @@ def test_polyline_tangent_at_point(coords, input, expected):
     assert expected == Polyline(coords).tangent_at_point(input)
 
 
+@pytest.mark.parametrize("input, expected", [((0, 0, 0), 0.0), ((1, 0, 0), 0.5), ((1, 1, 0), 1.0), ((2, 0, 0), None)])
```
tomvanmele: i think the space behind the comma is something the ipy will not like...

nmaslarinos: Oops that's my bad, I'm sorry

src/compas/geometry/curves/polyline.py
nmaslarinos: I think that at the time I was getting an import error, now that I tried again everything works well. I probably forgot to change it to the original when I was done

tomvanmele (approved):

PR #1181, opened 2023-09-22
fixed bug in polyline shorten and extend not reseting the lines

nmaslarinos:
Fixed a bug where calling polyline shorten or extend would alter the polyline points, but not reset the lines attribute. This creates problems where the polylines lines are wrongly displayed and the length attribute returns the previous, now wrong, value. I updated the test documents to also include a check for the length of the resulting polyline.

### Checklist


src/compas/geometry/curves/polyline.py
```diff
@@ -578,10 +578,10 @@ def extend(self, length):
         """
         try:
             start, end = length
-            self.points[0] = self.points[0] + self.lines[0].vector.unitized().scaled(-start)
+            self[0] = self[0] + self.lines[0].vector.unitized().scaled(-start)
```
tomvanmele: this necessary for the fix?

src/compas/geometry/curves/polyline.py
```diff
@@ -578,10 +578,10 @@ def extend(self, length):
...
         except TypeError:
             start = end = length
-        self.points[-1] = self.points[-1] + self.lines[-1].vector.unitized().scaled(end)
+        self[-1] = self[-1] + self.lines[-1].vector.unitized().scaled(end)
```
tomvanmele: same here. if it is not necessary, i find the code more readable with the explicit reference to points...

nmaslarinos:
You are right for both, this is one way of emptying the `polyline._lines` without explicitly doing it. Of course, it can remain as is and simply add a `self._lines = None` before the end of the function

tomvanmele (approved):

gonzalocasas:
@tomvanmele two questions: 1) ready to merge, right? 2) should I backport this to LTS?

tomvanmele:
yes and yes :)

PR #1194, opened 2023-10-04
Small feature: adding `axes` to Frame

romanarust:
Small feature: adding `axes` to Frame to iterate over x-,y-, z- axes

### Checklist


Licini (approved):

src/compas/geometry/frame.py
```diff
@@ -190,6 +190,10 @@ def zaxis(self):
             self._zaxis = self.xaxis.cross(self.yaxis)
         return self._zaxis
 
+    def axes(self):
```
Licini, line 195: Hey @romanarust Just realized, could you also add this one in the docstring on top of class?

romanarust: added `axes` to the `Box` as well for convenience

src/compas/geometry/shapes/box.py
```diff
@@ -263,6 +265,10 @@ def diagonal(self):
     def dimensions(self):
         return [self.xsize, self.ysize, self.zsize]
 
+    def axes(self):
```
tomvanmele: i would prefer we don't do this. first of all, we would have to do it many other places for consistency, since many other objects now have a frame. secondly, we already have too many of these aliases/shortcuts as it is. they create unnecessary bloated APIs in my opinion...

tomvanmele (approved):

src/compas/geometry/shapes/box.py
romanarust: ok, makes sense.

PR #1206, opened 2023-11-02
More flexible services for RPC

tomvanmele:

### Checklist


Licini:
nice!

tomvanmele:
```python
# special.py

from compas.rpc import Server
from compas.rpc import Dispatcher


class SpecialService(Dispatcher):
    def special(self):
        return "special"


def start_service():
    server = Server(("0.0.0.0", 1753))
    service = SpecialService()
    server.register_instance(service)
    server.serve_forever()


if __name__ == "__main__":
    import argparse
    parser = argparse.ArgumentParser()
    args, unknown = parser.parse_known_args()
    start_service()

```

```python
# client.py

import os
from compas.rpc import Proxy

proxy = Proxy(service="special", cwd=os.path.dirname("__file__"))

proxy.special()

```

src/compas/rpc/services/default.py
```diff
@@ -4,67 +4,32 @@
...
-    server = Server(("0.0.0.0", port))
+    host = "0.0.0.0"
+    address = host, port
+    server = Server(address)
 
     # register a few utility functions
-    server.register_function(server.ping)
-    server.register_function(server.remote_shutdown)
+    # server.register_function(server.ping)
+    # server.register_function(server.remote_shutdown)
```
gonzalocasas: delete

src/compas/rpc/proxy.py
```diff
@@ -56,6 +56,12 @@ class Proxy(object):
     capture_output : bool, optional
         If True, capture the stdout/stderr output of the remote process.
         In general, `capture_output` should be True when using a `pythonw` as executable (default).
+    path : str, optional
+        Path to the folder containing the module to be proxied.
+        This is useful for cases where the module to be proxied is not on the PYTHONPATH.
+    cwd : str, optional
+        Current working directory for the process that will be started to run the server.
+        This is useful for cases where a custom service is used and the service is not on the PYTHONPATH.
```
gonzalocasas: Isn't `cwd` too cryptic for most people? Shouldn't we just err on the side of verbosity and call it `working_directory`?

src/compas/rpc/services/special.py
```diff
@@ -0,0 +1,42 @@
+"""
+Minimal example of a "special" service.
```
gonzalocasas: I'm not very convinced about the naming, "special" but why? What's special about it? :) Do you mean "Custom" or "UserDefined"?

gonzalocasas (approved):
lgtm + feedback

src/compas/rpc/services/special.py
tomvanmele: not it is just an example. perhaps it should be removed and the example added to the docs

gonzalocasas: 👍

PR #1207, opened 2023-11-07
Release for Alpha.2

Licini:
Releasing alpha.2 so that new classes like `compas.brep` is included. We need them to be present for packages like compas_occ to correctly build their latest documentation in github actions. Once this is merged I will create the tag to trigger the release

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 2.0.0-alpha.1
+current_version = 2.0.0-alpha.2
 message = Bump version to {new_version}
 commit = True
 tag = True
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+### Changed
+
+### Removed
+
+
+## [2.0.0-alpha.2] 2023-11-07
+
+### Added
+
 * Added `Frame.axes`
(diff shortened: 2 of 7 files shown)


gonzalocasas (approved):
LGTM! 
The comment led me to think: have we documented the rationale behind moving brep out of geometry and into its own package? Because from the distance, brep feels like it should be either geometry or it should be datastructures, but a new top-level package feels odd

chenkasirer:
> LGTM! The comment led me to think: have we documented the rationale behind moving brep out of geometry and into its own package? Because from the distance, brep feels like it should be either geometry or it should be datastructures, but a new top-level package feels odd

What I remember from this discussion is that Brep is kind of both geometry and datastructure (topology?). Then in `compas_occ` (which is the only other place that has `brep`) it has its own top level package..

tomvanmele:
yes indeed. a Brep is both geometry and topology, but it is not like our data structures in the sense that it is not used to structure arbitrary other data.

gonzalocasas:
> What I remember from this discussion is that Brep is kind of both geometry and datastructure (topology?).

I can imagine the discussion, but now it feels off. Mesh is also both geometry and datastructure and we don't have it on its own top-level. I am not suggesting to change now, but if there's a strong argument for why this is like this, it should be documented to avoid having circling back in to the same discussion in the future, and/or to clarify for externals why it is the way it is.

tomvanmele:
so what is the verdict?

chenkasirer:
Edit - comment not constructive to this discussion.

PR #1218, opened 2023-12-04
Add to_meshes() method for compas_rhino brep.py

jckenny59:
Added method for conversion of RhinoBrep faces to compas Meshes.

image

The primary question is the input signature. 

The initial input signature is `u=16, v=16`, however, this is unused other possible input options could be `System.Double density (refine intensity of the mesh), and System.Double minium_edge_length (allow user definition of the minimum edge length of the mesh). `

This would allow more user control over the resulting mesh faces, however, the current signature is the same as before.

### Checklist


src/compas_rhino/geometry/brep/brep.py
```diff
@@ -247,6 +248,26 @@ def from_mesh(cls, mesh):
...
+    def to_meshes(self, density, minimum_edge_length):         
+    
+        """Convert the faces of this Brep shape to meshes.
+
+        Parameters
+        ----------
+        u : int, optional
+            The number of mesh faces in the U direction of the underlying surface geometry of every face of the Brep.
+        v : int, optional
+            The number of mesh faces in the V direction of the underlying surface geometry of every face of the Brep.
```
tomvanmele, line 311: this doesn't match the parameters of the function

jckenny59: My apologies, I was debating if I should leave it with the initial `u , v` parameters the the other parameters and I forgot to change that one! should be different now!

tomvanmele: no worries, that's what the review is for :)

src/compas_rhino/geometry/brep/brep.py
```diff
@@ -247,6 +248,25 @@ def from_mesh(cls, mesh):
     # Methods
     # ==============================================================================
 
+    def to_meshes(self, u=16, v=16):
```
chenkasirer, line 303: I know it's `to_meshes` and not `to_mesh` but would it maybe make sense to return a single joint mesh? If remember right, `Mesh.CreateFromBrep` returns a mesh for each face of the Brep, is that something a caller should always expect?

src/compas_rhino/geometry/brep/brep.py
```diff
@@ -298,6 +299,25 @@ def contains(self, object):
...
+            The number of mesh faces in the U direction of the underlying surface geometry of every face of the Brep.
+        v : int, optional
+            The number of mesh faces in the V direction of the underlying surface geometry of every face of the Brep.
+
+        Returns
+        -------
+        list[:class:`~compas.datastructures.Mesh`]
+
+        """
+        rg_meshes = Rhino.Geometry.Mesh.CreateFromBrep(self._brep.native_brep, Rhino.Geometry.MeshingParamaters.Default)
```
chenkasirer: Notice `self._brep` is already the `Rhino.Geometry.Brep` instance so here you should use either `self._brep` directly or `self.native_brep`.

chenkasirer (approved):
LGTM!
Don't forget do add yourself to `AUTHORS.md` :)

PR #1221, opened 2023-12-07
Fall back to the default implementation if possible, raise otherwise

tomvanmele:

### Checklist

### DIFF
--- a/src/compas/plugins.py
+++ b/src/compas/plugins.py
@@ -284,15 +284,15 @@ def pluggable(
                 if plugin_impl is None:
                     try:
                         return func(*args, **kwargs)
-                    except PluginNotInstalledError as e:
-                        raise e
                     except NotImplementedError:
                         raise PluginNotInstalledError(
-                            "Plugin not found for extension point URL: {}".format(extension_point_url)
+                            "Plugin not found and no default implementation for extension point URL: {}".format(
+                                extension_point_url
+                            )
                         )
                     except ImportError:
                         raise PluginDefaultNotAvailableError(
-                            "The default implementation for extension point URL {} is not available in the current environment.".format(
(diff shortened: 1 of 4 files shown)


src/compas/plugins.py
```diff
@@ -264,6 +269,13 @@ def wrapper(*args, **kwargs):
             # Select first matching plugin
             if selector == "first_match":
                 plugin_impl = _select_plugin(extension_point_url)
+                if plugin_impl is None:
+                    try:
+                        return func(*args, **kwargs)
+                    except Exception:
```
Licini: Shouldn't this Exception be more specific? Otherwise it will catch other type of errors and falsely indicates it was because of Plugin not found

Licini: maybe use `ImportError`?

tomvanmele: yes that would be better indeed

tomvanmele:
how about something like this?

src/compas/plugins.py
```diff
@@ -264,6 +281,21 @@ def wrapper(*args, **kwargs):
             # Select first matching plugin
             if selector == "first_match":
                 plugin_impl = _select_plugin(extension_point_url)
+                if plugin_impl is None:
+                    try:
+                        return func(*args, **kwargs)
+                    except PluginNotInstalledError as e:
+                        raise e
```
Licini: Seems this one is now redundant, no? Because from now on there will always be a function to fallback to, which is the one decorated as pluggable. If this original default fails, it's either one of the following two errors (or something else internal of the function itself).

tomvanmele: so the pluggable either has an implementation (which might fail for some reason), or it raises `NotImplementedError`?

Licini: Yes I think so

Licini (approved):
LGTM now!

PR #1222, opened 2023-12-07
Moving global scene functions

Licini:
Moving global Scene operations like `register`, `build`, `redraw`, `clear` etc. to an independent file at `compas.scene.context`. This way the sceneobject is much more simplified, and we no longer need to use `SceneObject.__new__()` , since the build is now the responsibility of `Scene`


tomvanmele (commented):
generally i think this is okay, but how and where does the user have the possibility to manually specify a context? i imagine this to be part of the scene constructor. something like
```python
scene = Scene(context='Rhino')
```

Licini:
> generally i think this is okay, but how and where does the user have the possibility to manually specify a context? i imagine this to be part of the scene constructor. something like
> 
> ```python
> scene = Scene(context='Rhino')
> ```

Maybe like this?
https://github.com/compas-dev/compas/pull/1222/commits/c78aeb5971a275ad533aea05228d32d5fb51e1e6

tomvanmele (approved):
LGTM, provided that it works :)

PR #1223, opened 2023-12-12
Track guids in drawn objects

Licini:
Still in the work. This is to add the mechanism to track guids in scene objects. So the `draw` function should now always returns a list of GUIDs. Question: I currently wrote `item_guids`, but not sure it is the best name, any other suggestions @tomvanmele ? For instance in GH the draw function will be returning the instances of rhino geometry like Box, Sphere instead of the doc objects. And in blender it will be a list of `bpy.types.Object`.


src/compas/scene/sceneobject.py
```diff
@@ -13,14 +13,12 @@ class SceneObject(object):
...
-    context : str, optional
-        Explicit context to pick the SceneObject from.
-        If not specified, an attempt will be made to automatically detect the appropriate context.
+
 
     Attributes
     ----------
-    ITEM_SCENEOBJECT : dict[str, dict[Type[:class:`~compas.data.Data`], Type[:class:`~compas.scene.SceneObject`]]]
-        Dictionary mapping data types to the corresponding scene objects types per visualization context.
+    item_guids : list[object]
```
Licini: Or perhaps this attribute goes to `SceneObject` implementation of each contexts, then the name varies depends on what it is actually tracking. In compas_rhino we call it `item_guids`, in GH `item_gemetries` and in blender `item_objects`?

tomvanmele: why not just guids

Licini: That was my initial attention too. But won't `guids` creates the confusion whether it means guids for itself or the guids for drawn entities of underlying item? I understand a possible distinction could be `guid` singular refers to self, `guids` plural refers to drawn entities. Another discussion I didn't want to go there yet is whether we should have guid for SceneObject as well. Anyway, maybe easier to discuss all this in tomorrow's meeting

Licini:
@tomvanmele Ok this is ready for review. The tracking for `guids` is now added to all scene objects. There are also some fixes / implementations added now all types of objects works in all 3 platform

src/compas/scene/scene.py
```diff
@@ -52,17 +52,8 @@ def clear(self):
...
-            # TODO: unify output of draw(), so we don't have to do this
-            if isinstance(drawn_object, (list, tuple)):
-                for item in drawn_object:
-                    if isinstance(item, (list, tuple)):
-                        drawn_objects.extend(item)
-                    else:
-                        drawn_objects.append(item)
-            else:
-                drawn_objects.append(drawn_object)
+            sceneobject.clear()
```
tomvanmele: would it not be better to collect the objects that have to be cleared and do this all at once? at least in rhino it is a costly operation...

src/compas/scene/sceneobject.py
```diff
@@ -13,14 +13,12 @@ class SceneObject(object):
...
-        Explicit context to pick the SceneObject from.
-        If not specified, an attempt will be made to automatically detect the appropriate context.
+
 
     Attributes
     ----------
-    ITEM_SCENEOBJECT : dict[str, dict[Type[:class:`~compas.data.Data`], Type[:class:`~compas.scene.SceneObject`]]]
-        Dictionary mapping data types to the corresponding scene objects types per visualization context.
+    guids : list[object]
+        The GUIDs of the drawn item in the visualization context.
```
tomvanmele: items

tomvanmele: items drawn in the visualization context

src/compas/scene/sceneobject.py
```diff
@@ -52,7 +55,7 @@ def draw(self):
         """The main drawing method."""
         raise NotImplementedError
 
-    @staticmethod
-    def draw_collection(collection):
-        """Drawing method for drawing an entire collection of objects."""
+    @abstractmethod
```
tomvanmele: should the default not be to call the pluggable and have the pluggable throw the error?

src/compas_rhino/scene/pointobject.py
```diff
@@ -37,13 +37,15 @@ def draw(self, color=None):
 
         Returns
         -------
-        System.Guid
-            The GUID of the created Rhino object.
+        list[System.Guid]
+            List of GUIDs of the object created in Rhino.
```
tomvanmele: objects

src/compas_rhino/scene/polygonobject.py
```diff
@@ -39,8 +39,8 @@ def draw(self, color=None):
 
         Returns
         -------
-        System.Guid
-            The GUID of the created Rhino object.
+        list[System.Guid]
+            List of GUIDs of the object created in Rhino.
```
tomvanmele: probably everywhere

src/compas_rhino/scene/sceneobject.py
```diff
@@ -73,3 +74,14 @@ def clear_layer(self):
...
+
+    def clear(self, purge=None, redraw=False):
+        """Clear the scene object.
+
+        Returns
+        -------
+        None
+
+        """
+        delete_objects(self.guids, purge=purge, redraw=redraw)
```
tomvanmele: doing this one by one seems costly

src/compas/scene/scene.py
```diff
@@ -47,22 +47,18 @@ def _get_node(self, sceneobject):
...
-        clear()
+        guids = []
+        for sceneobject in self.sceneobjects:
+            guids += sceneobject.guids
+            sceneobject._guids = None
+        clear(guids=guids)
 
     def redraw(self):
+        self.clear()
+
```
Licini, line 58: @tomvanmele How about now? we collect the all guids of scene then clear all together. Also we add guids as argument for clear, so it does not delete the other stuff in context that are not managed by the scene.

src/compas/scene/sceneobject.py
```diff
@@ -52,7 +56,7 @@ def draw(self):
...
         raise NotImplementedError
 
-    @staticmethod
-    def draw_collection(collection):
-        """Drawing method for drawing an entire collection of objects."""
-        raise NotImplementedError
+    def clear(self):
+        """The main clearing method."""
+        clear(guids=self.guids)
+        self._guids = None
```
Licini, line 62: Then we also use same same `clear` function for individual object. In this way the clear function of the object doesn't need to be re-implemented anymore

tomvanmele (approved):

PR #1227, opened 2023-12-17
handle colors in SceneObject

Licini:
@tomvanmele 

1. Adding generic vis attributes to scene object:
- `color`, `pointcolor`, `linecolor`, `facecolor` as descriptors.
- `linewidth`, `pointsize`, `opacity` as simple attribute.

2. For `MeshObject` and `Networkobject`:
- Use `vertexcolor=SceneObject.pointcolor` and `nodecolor=SceneObject.pointcolor` to read from and write to shared descriptors.

3. Naming convention, should it be `pointcolor` or `point_color`??

### DIFF
--- a/src/compas/scene/descriptors/colordict.py
+++ b/src/compas/scene/descriptors/colordict.py
@@ -54,9 +54,6 @@ class ColorDictAttribute(object):
             A defaultdict with the value stored in the default attribute corresponding to the descriptor as a default value.
 
         """
-        if obj is None:
-            # return the descriptor itself if accessed from the class
-            return self
 
         if not hasattr(obj, self.private_name):
             setattr(obj, self.private_name, None)
--- a/src/compas/scene/descriptors/colordict.py
+++ b/src/compas/scene/descriptors/colordict.py
@@ -54,6 +54,10 @@ class ColorDictAttribute(object):
             A defaultdict with the value stored in the default attribute corresponding to the descriptor as a default value.
 
         """
+        if obj is None:
+            # return the descriptor itself if accessed from the class
+            return self
+
         if not hasattr(obj, self.private_name):
             setattr(obj, self.private_name, None)
 
--- a/src/compas/scene/geometryobject.py
+++ b/src/compas/scene/geometryobject.py
@@ -22,8 +22,6 @@ class GeometryObject(SceneObject):
         The color of the points.
     linecolor : :class:`compas.colors.Color`
         The color of the lines or curves.
-    curvecolor : :class:`compas.colors.Color`
-        The color of the lines or curves.
     surfacecolor : :class:`compas.colors.Color`
         The color of the surfaces.
     pointsize : float
@@ -34,7 +32,7 @@ class GeometryObject(SceneObject):
     """
 
     pointcolor = ColorAttribute()
-    curvecolor = linecolor = ColorAttribute()
+    linecolor = ColorAttribute()
     surfacecolor = ColorAttribute()
 
     def __init__(self, geometry, **kwargs):
--- a/src/compas/scene/geometryobject.py
+++ b/src/compas/scene/geometryobject.py
@@ -3,7 +3,6 @@ from __future__ import absolute_import
 from __future__ import division
 
 from .sceneobject import SceneObject
-from .descriptors.color import ColorAttribute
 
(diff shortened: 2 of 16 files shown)


src/compas/scene/descriptors/colordict.py
```diff
@@ -54,6 +54,10 @@ def __get__(self, obj, otype=None):
             A defaultdict with the value stored in the default attribute corresponding to the descriptor as a default value.
 
         """
+        if obj is None:
+            # return the descriptor itself if accessed from the class
+            return self
+
```
tomvanmele: why is this?

src/compas/scene/meshobject.py
```diff
@@ -27,11 +25,11 @@ class MeshObject(SceneObject):
         Defaults to the real coordinates.
     color : :class:`compas.colors.Color`
         The base RGB color of the mesh.
-    vertex_color : :class:`compas.colors.ColorDict`]
+    vertexcolor : :class:`compas.colors.ColorDict`]
```
tomvanmele: ] at the end

src/compas/scene/sceneobject.py
```diff
@@ -26,11 +27,24 @@ class SceneObject(object):
     # add this to support the descriptor protocol vor Python versions below 3.6
     __metaclass__ = DescriptorProtocol
 
+    color = ColorDictAttribute()
+    pointcolor = ColorDictAttribute()
+    linecolor = ColorDictAttribute()
+    facecolor = ColorDictAttribute()
```
tomvanmele, line 33: i guess a geometry object has points, curves (or lines), and surfaces.
each of these have one default color (not a color dict).

data structures have vertices, edges, faces.
there we want to provide the option to set different colors for different components of the same type.
so vertexcolor, edgecolor, facecolor are color dicts instead.

tomvanmele:
i guess a geometry object has points, curves (or lines), and surfaces.
each of these have one default color (not a color dict).

data structures have vertices, edges, faces.
there we want to provide the option to set different colors for different components of the same type.
so vertexcolor, edgecolor, facecolor are color dicts instead.

tomvanmele:
so

`SceneObject` defines `color` as `ColorAttribute`
`GeometryObject` adds `pointcolor`, `linecolor` (`curvecolor`), `surfacecolor` as `ColorAttribute`
`MeshObject` adds `vertexcolor`, `edgecolor`, `facecolor` as `ColorDictAttribute`
...

src/compas/scene/descriptors/colordict.py
Licini: This was to enable the usage like `nodecolor=SceneObject.pointcolor` in a descendent class. When the `pointcolor` is accessed from class itself, we need `___get___` to return the descriptor itself instead of the data. 

But seems this mechanism is not needed for now after the adjustment. So I'm removing it.

Licini:
@tomvanmele How about now?

src/compas/scene/geometryobject.py
```diff
@@ -18,15 +18,30 @@ class GeometryObject(SceneObject):
...
+    pointsize : float
+        The size of the points.
+    linewidth : float
+        The width of the lines or curves.
 
     """
 
-    color = ColorAttribute(default=None)
+    pointcolor = ColorAttribute()
+    curvecolor = linecolor = ColorAttribute()
```
Licini: let me know what do you think of this

src/compas/scene/geometryobject.py
```diff
@@ -17,11 +18,30 @@ class GeometryObject(SceneObject):
...
+        The color of the surfaces.
+    pointsize : float
+        The size of the points.
+    linewidth : float
+        The width of the lines or curves.
 
     """
 
+    pointcolor = ColorAttribute()
+    curvecolor = linecolor = ColorAttribute()
```
tomvanmele: let's decide on one of the two

tomvanmele: maybe stick with line for now

Licini:
@tomvanmele Ready for review now. This is just to setup these attributes in base classes. The actual implementation for each context gonna have to come later as a bigger PR.

tomvanmele (approved):

PR #1233, opened 2023-12-20
Rhino brep slice

chenkasirer:
Added `RhinoBrep.slice()` and `RhinoBrep.trimmed()`.

Changed signature of `Brep.slice()` to return `compas.geometry.Curve` instead of `BrepFace`.

@tomvanmele this isn't quite what we discussed, but in Rhino a Brep with no surfaces is not valid. `BrepFace` and other Brep components all don't have public constructors so I don't see a way to return a face either :/ 
Let me know what you think.

### Checklist


src/compas/geometry/brep/brep.py
```diff
@@ -883,6 +883,23 @@ def trim(self, trimming_plane, tolerance):
         """
         raise NotImplementedError
 
+    def trimmed(self, trimming_plane, tolerance):
+        """Returns a trimmed copy of this Brep using the given trimming plane.
+
+        Parameters
+        ----------
+        trimming_plane : :class:`~compas.geometry.Frame`
```
tomvanmele: why not just a plane?

src/compas_rhino/geometry/brep/brep.py
```diff
@@ -440,6 +466,26 @@ def from_boolean_intersection(cls, breps_a, breps_b):
...
+            The plane to slice through the brep.
+
+        Returns
+        -------
+        list(:class:`~compas.geometry.Curve`)
+            Zero or more curves which represent the intersection(s) between the brep and the plane.
+
+        """
+        if isinstance(plane, Frame):
+            plane = Plane.from_frame(plane)
```
tomvanmele, line 482: is there a reason why trimming converts a plane to a frame, and slicing converts a frame to a plane?

chenkasirer: totally random. but changed in `trimmed` as makes more sense to eventually use `Plane`..

tomvanmele (approved):

PR #1234, opened 2023-12-20
Add __str__ to Point and Vector for pretty prints

gonzalocasas:
A while ago, the `__repr__` of `Point` and `Vector` was changed to use unlimited precision, which is good for `eval()`ing them, but it's kind of ugly for printing and it makes some doctests hard to write. This PR adds a `__str__` method to these two classes, so that the `__repr__` still behaves properly outputting every digit, but when printing these as strings, it will use limited precision instead.

### Checklist


src/compas/geometry/point.py
```diff
@@ -130,7 +130,7 @@ def __str__(self):
             self.x,
             self.y,
             self.z,
-            TOL.PRECISION,
+            TOL.precision,
```
tomvanmele: technically, precision can actually also be on the other side of the decimal point. so, technically, better would be
```python
"{0}(x={1}, y={2}, z={3})".format(
    type(self).__name__,
    TOL.format_number(self.x),
    TOL.format_number(self.y),
    TOL.format_number(self.z)
)
```
but this also works

tomvanmele: :)

gonzalocasas: Ah! True! Fixed at https://github.com/compas-dev/compas/pull/1234/commits/dc19b7d42707e24b44ac0b773fa3020f5d6ca8ab

tomvanmele (approved):

PR #1235, opened 2023-12-21
Frame stack, transformation, adding SceneTree and SceneObjectNode

Licini:
- Handling scene object `frame stack` and `transformation`. 
- Added specialized classes for `SceneTree` and `SceneObjectNode`

Plz see notes down in comment session:


src/compas/scene/scene.py
```diff
@@ -6,58 +6,95 @@
...
+        sceneobject._node = node
+
+    def get_node_from_object(self, sceneobject):
+        for node in self.nodes:
+            if isinstance(node, SceneObjectNode):
+                if node.object == sceneobject:
+                    return node
+        raise ValueError("Scene object not in scene")
+
+
```
Licini: Adding these two classes to abstract tree operations away a bit from the `Scene`

src/compas/scene/scene.py
```diff
@@ -6,58 +6,95 @@
...
 
     @property
     def tree(self):
         return self._tree
 
     @property
-    def sceneobjects(self):
-        return [node.attributes["sceneobject"] for node in self.tree.nodes if "sceneobject" in node.attributes]
+    def objects(self):
+        return self.tree.objects
```
Licini, line 192: How do we feel about just `Scene.objects` instead of `Scene.sceneobjects` ?

src/compas/scene/scene.py
```diff
@@ -6,58 +6,95 @@
...
-        name = item.name or item.__class__.__name__
-        node = TreeNode(name, attributes={"sceneobject": sceneobject})
-
-        if parent is None:
-            self.tree.add(node, parent=self.tree.root)
-        else:
-            parent_node = self._get_node(parent)
-            self.tree.add(node, parent=parent_node)
-
+        self.tree.add_object(sceneobject, parent=parent)
```
Licini: Now all these code goes into `SceneTree`

src/compas/scene/scene.py
```diff
@@ -6,58 +6,95 @@
...
     def clear(self):
+        clear()
+
+    def clear_objects(self):
         guids = []
-        for sceneobject in self.sceneobjects:
+        for sceneobject in self.objects:
             guids += sceneobject.guids
             sceneobject._guids = None
         clear(guids=guids)
```
Licini: Added a dedicated function `clear_objects` to clear objects managed by Scene, and  set `clear` to clear the entire context

src/compas/scene/sceneobject.py
```diff
@@ -42,15 +48,46 @@ def __new__(cls, item, **kwargs):
...
+    def parent(self):
+        if self.node:
+            return self.node.parent_object
+
+    @property
+    def children(self):
+        if self.node:
+            return self.node.children_objects
+        else:
+            return []
```
Licini: Adding `parent` and `children` property to `SceneObject`. They are fetched through associated `SceneObjectNode`.

src/compas/scene/scene.py
```diff
@@ -6,58 +6,95 @@
...
+        self.object = sceneobject
+
+    @property
+    def parent_object(self):
+        if self.parent and isinstance(self.parent, SceneObjectNode):
+            return self.parent.object
+        return None
+
+    @property
+    def children_objects(self):
```
tomvanmele: child_objects

src/compas/scene/scene.py
```diff
@@ -6,58 +6,95 @@
...
+        else:
+            parent_node = self.get_node_from_object(parent)
+            self.add(node, parent=parent_node)
+
+        sceneobject._node = node
+
+    def get_node_from_object(self, sceneobject):
+        for node in self.nodes:
+            if isinstance(node, SceneObjectNode):
+                if node.object == sceneobject:
```
tomvanmele: == => is?

src/compas/scene/sceneobject.py
```diff
@@ -59,6 +96,31 @@ def transformation(self):
...
+        matrices = [Transformation.from_frame(f) for f in frame_stack]
+        if matrices:
+            transformation_world = reduce(mul, matrices[::-1])
+        else:
+            transformation_world = Transformation()
+
+        if self.transformation:
+            transformation_world *= self.transformation
+
+        return transformation_world
```
Licini: Added a `transformation_world` property. This is the calculated final transformation from the frame stack and local transformation.

src/compas/scene/sceneobject.py
```diff
@@ -59,6 +96,31 @@ def transformation(self):
...
+
+    def add(self, item, **kwargs):
+        if self.node:
+            return self.node.add_item(item, **kwargs)
+        else:
+            raise ValueError("Cannot add items to a scene object without a node.")
```
Licini: Expose `add` at `SceneObject`, so we can do stuff like
```python
obj1 = scene.add(item1)
obj2 = obj1.add(item2)
obj3 = obj2.add(item3)
```
or even:
```python
scene.add(item1).add(item2).add(item3)...
```

src/compas_rhino/scene/boxobject.py
```diff
@@ -46,10 +46,8 @@ def draw(self, color=None):
         attr = attributes(name=self.geometry.name, color=color, layer=self.layer)
 
         geometry = box_to_rhino(self.geometry)
-
-        if self.transformation:
-            transformation = transformation_to_rhino(self.transformation)
-            geometry.Transform(transformation)
+        transformation = transformation_to_rhino(self.transformation_world)
+        geometry.Transform(transformation)
```
Licini: Now in implementation of each objects, we should always use `transformation_world`.

Licini:
Let me know what you think @tomvanmele , if all good, I will update implementation of all objects. And add corresponding docs. Also here is a script I use for testing:
```python
from compas.geometry import Capsule
from compas.geometry import Circle

box = Box.from_width_height_depth(1, 1, 1)
capsule = Capsule(0.5, 1, Frame.worldXY())
circle = Circle(1, Frame.worldXY())


scene = Scene()
scene.clear()


f = Frame.from_transformation(Translation.from_vector([5, 0, 0]))
t1 = Translation.from_vector([0, 1, 0])
t2 = Translation.from_vector([0, 2, 0])
t3 = Translation.from_vector([0, 3, 0])


scene.add(box, frame=f, transformation=t1).add(capsule, frame=f, transformation=t2).add(circle, frame=f, transformation=t3)


scene.print_hierarchy()
scene.redraw()

```

And it will draw this:
image

src/compas/scene/scene.py
gonzalocasas: or just `def children()` and `def parent()`?

gonzalocasas:
> Let me know what you think @tomvanmele , if all good, I will update implementation of all objects. And add corresponding docs. Also here is a script I use for testing:
> 
> ```python
> from compas.scene import Scene
> from compas.geometry import Box
> from compas.geometry import Capsule
> from compas.geometry import Circle
> from compas.geometry import Frame
> from compas.geometry import Translation
> 
> box = Box.from_width_height_depth(1, 1, 1)
> capsule = Capsule(0.5, 1, Frame.worldXY())
> circle = Circle(1, Frame.worldXY())
> 
> 
> scene = Scene()
> scene.clear()
> 
> 
> f = Frame.from_transformation(Translation.from_vector([5, 0, 0]))
> t1 = Translation.from_vector([0, 1, 0])
> t2 = Translation.from_vector([0, 2, 0])
> t3 = Translation.from_vector([0, 3, 0])
> 
> 
> scene.add(box, frame=f, transformation=t1).add(capsule, frame=f, transformation=t2).add(circle, frame=f, transformation=t3)
> 
> 
> scene.print_hierarchy()
> scene.redraw()
> ```
> 
> And it will draw this: image


Thanks for the example! I'm a bit confused about what we see here, when you use the chaining API (`.add().add()`..) is there any parenting being done implicitly right? So the capsule is a child of box, and the circle is a child of capsule, right? If that's the case, why do they seem to be all equally distant from each other? I would have assumed that since the transformations are `(0, 1, 0)`, `(0, 2, 0)` and `(0, 3, 0)`, then box would be at `(0, 1, 0)` of the world coordinate system, then capsule would be at `(0, 3, 0)`, and circle would be at `(0, 6, 0)`.

src/compas/scene/scene.py
tomvanmele: those names already exist and refer to the parnet/child nodes. this is about the objects contained in the nodes

Licini:
@gonzalocasas Yes, this is equivalent to 
```python
boxObj = scene.add(box, frame=f, transformation=t1)
capsuleObj = boxObj.add(capsule, frame=f, transformation=t2)
circleObj = capsuleObj.add(circle, frame=f, transformation=t3)
```
or 
```python
boxObj = scene.add(box, frame=f, transformation=t1)
capsuleObj = scene.add(capsule, frame=f, transformation=t2, parent=boxObj)
circleObj = scene.add(circle, frame=f, transformation=t3, parent=capsuleObj)
```
So the hierarchy here is:
boxObj
--capsuleObj 
----circleObj

The final transformation is calculate from the stack of frames plus applying the transformation. For each object we set the their local frames to be at [5, 0, 0], that means after considering their parent frames, the origin of objects are at:

boxObj [5, 0, 0]
--capsuleObj [10, 0, 0]
----circleObj [15, 0, 0]

Then we apply transformation on each:
boxObj [5, 1, 0]
--capsuleObj [10, 2, 0]
----circleObj [15, 3, 0]

src/compas/scene/sceneobject.py
Licini: @gonzalocasas The world transformation is calculated here

gonzalocasas:
> The final transformation is calculate from the stack of frames plus applying the transformation. For each object we set the their local frames to be at [5, 0, 0], that means after considering their parent frames, the origin of objects are at:
> 
> boxObj [5, 0, 0] --capsuleObj [10, 0, 0] ----circleObj [15, 0, 0]
> 
> Then we apply transformation on each: boxObj [5, 1, 0] --capsuleObj [10, 2, 0] ----circleObj [15, 3, 0]

The parenting is now clear, thanks!
And I understand now the visual output, I had missed the fact that there's a local frame. However, this thing about a transformation and frame are still confusing to me. Why do we have both a local frame AND also a local transformation to that local frame? (on a first read, I assumed the transformation was the one transforming world to local frame, because I missed the existence of the local frame. IMO having both is a bit redundant)

tomvanmele:
> > The final transformation is calculate from the stack of frames plus applying the transformation. For each object we set the their local frames to be at [5, 0, 0], that means after considering their parent frames, the origin of objects are at:
> > boxObj [5, 0, 0] --capsuleObj [10, 0, 0] ----circleObj [15, 0, 0]
> > Then we apply transformation on each: boxObj [5, 1, 0] --capsuleObj [10, 2, 0] ----circleObj [15, 3, 0]
> 
> The parenting is now clear, thanks! And I understand now the visual output, I had missed the fact that there's a local frame. However, this thing about a transformation and frame are still confusing to me. Why do we have both a local frame AND also a local transformation to that local frame? (on a first read, I assumed the transformation was the one transforming world to local frame, because I missed the existence of the local frame. IMO having both is a bit redundant)

what if you have a rotation around local axes?

gonzalocasas:
> > > The final transformation is calculate from the stack of frames plus applying the transformation. For each object we set the their local frames to be at [5, 0, 0], that means after considering their parent frames, the origin of objects are at:
> > > boxObj [5, 0, 0] --capsuleObj [10, 0, 0] ----circleObj [15, 0, 0]
> > > Then we apply transformation on each: boxObj [5, 1, 0] --capsuleObj [10, 2, 0] ----circleObj [15, 3, 0]
> > 
> > 
> > The parenting is now clear, thanks! And I understand now the visual output, I had missed the fact that there's a local frame. However, this thing about a transformation and frame are still confusing to me. Why do we have both a local frame AND also a local transformation to that local frame? (on a first read, I assumed the transformation was the one transforming world to local frame, because I missed the existence of the local frame. IMO having both is a bit redundant)
> 
> what if you have a rotation around local axes?

I guess I'm not considering the mutability of the local frame. So, local frame in this case would basically be immutable and transformation is the primary way to transform the object while still retaining its original local frame. Is that so?

tomvanmele:
you could rotate the frame, but since it your local coordinate system, keeping it fixed would make sense to me, and apply the local transformation to the object instead...

gonzalocasas:
> you could rotate the frame, but since it your local coordinate system, keeping it fixed would make sense to me, and apply the local transformation to the object instead...

Got it 👍  Makes sense to me too. Perhaps it would make sense to apply this to `compas_robots` at some point, currently the robotmodelartists (now robotmodelobjects) do the same thing but keep track of the initial frame when transforming it.

src/compas/scene/scene.py
```diff
@@ -6,58 +6,241 @@
...
+        The scene objects associated with the child nodes.
+
+    """
+
+    def __init__(self, sceneobject):
+        super(SceneObjectNode, self).__init__(name=sceneobject.name)
+        self.object = sceneobject
+
+    @property
+    def parent_object(self):
```
tomvanmele: in ling with my comments on the viewer PRs, i would write this in one word.
``parent_object`` could be interpreted as "action_subject"

src/compas/scene/scene.py
```diff
@@ -6,58 +6,241 @@
...
+        if self.parent and isinstance(self.parent, SceneObjectNode):
+            return self.parent.object
+        return None
+
+    @property
+    def child_objects(self):
```
tomvanmele: same here, even though "child" is not a verb. for consistency i would name this ``childobjects``

src/compas/scene/sceneobject.py
```diff
@@ -59,6 +106,50 @@ def transformation(self):
     def transformation(self, transformation):
         self._transformation = transformation
 
+    @property
+    def transformation_world(self):
```
tomvanmele: ``worldtransformation``?

tomvanmele: ``to_world``?

Licini:
@tomvanmele done

src/compas/scene/sceneobject.py
Licini: worldtransformation for me is better

tomvanmele (approved):

PR #1239, opened 2023-12-22
only trigger docs when PR approved

Licini:
This should make the docs workflow only triggered by the approval of PR

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -6,12 +6,12 @@ on:
       - main
     tags:
       - 'v*'
-  pull_request:
-    branches:
-      - main
+  pull_request_review:
+    types: [submitted]
 
 jobs:
   docs:
+    if: github.event.review.state == 'approved'
     runs-on: ubuntu-latest
     steps:
       - uses: compas-dev/compas-actions.docs@v3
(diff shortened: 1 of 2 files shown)


chenkasirer (approved):
LGTM!

Licini:
@chenkasirer Just realised i got it wrong, it would not run on push to main and version releases, could you check and approve again to see if it works properly this time? thx!

.github/workflows/docs.yml
```diff
@@ -6,12 +6,12 @@ on:
...
       - 'v*'
-  pull_request:
-    branches:
-      - main
+  pull_request_review:
+    types: [submitted]
 
 jobs:
   docs:
+    if: github.ref == 'refs/heads/main' || startsWith(github.ref, 'refs/tags/v') || (github.event_name == 'pull_request_review' && github.event.review.state == 'approved')
```
chenkasirer, line 14: so when merging to `main` does it still count as a `pull_request_review` event?

Licini: nop, this condition now means
`push to main` or `pushed a tag with v*` or `a pull_request_review event and it is an approval`

Licini: so it should trigger when satisfy any of the three

chenkasirer (approved):
LGTM

.github/workflows/docs.yml
chenkasirer: ah right! nice, seems to have worked :)

PR #1245, opened 2024-01-03
show_* flags to scene objects

Licini:
As titled


tomvanmele (approved):
eventually we should consider adding the possibility of visualising only a subset of the elements of data structures, but perhaps that is for another PR

PR #1247, opened 2024-01-04
Add storage and serialisation of general attributes

tomvanmele:

### Checklist

### DIFF
--- a/src/compas/data/data.py
+++ b/src/compas/data/data.py
@@ -74,7 +74,7 @@ class Data(object):
 
     def __init__(self, name=None):
         self._guid = None
-        self._name = None
+        self.attributes = {}
         if name:
             self.name = name
 
@@ -95,13 +95,15 @@ class Data(object):
             "dtype": self.dtype,
             "data": self.data,
         }
+        if self.attributes:
+            state["attrs"] = self.attributes
         if minimal:
             return state
         state["guid"] = str(self.guid)
         return state
 
     @classmethod
-    def __jsonload__(cls, data, guid=None):
+    def __jsonload__(cls, data, guid=None, attrs=None):
         """Construct an object of this type from the provided data to support COMPAS JSON serialization.
 
         Parameters
@@ -110,6 +112,8 @@ class Data(object):
             The raw Python data representing the object.
         guid : str, optional
             The GUID of the object.
+        attrs : dict, optional
+            The additional attributes of the object.
 
         Returns
         -------
@@ -119,6 +123,8 @@ class Data(object):
         obj = cls.from_data(data)
         if guid is not None:
             obj._guid = UUID(guid)
+        if attrs is not None:
+            obj.attributes.update(attrs)
         return obj
 
     def __getstate__(self):
@@ -130,6 +136,9 @@ class Data(object):
         self.__dict__.update(state["__dict__"])
         if "guid" in state:
             self._guid = UUID(state["guid"])
+        # could be that this is already taken care of by the first line
+        if "attrs" in state:
+            self.attributes.update(state["attrs"])
 
     @property
     def dtype(self):
@@ -161,13 +170,11 @@ class Data(object):
 
     @property
     def name(self):
-        if not self._name:
-            self._name = self.__class__.__name__
-        return self._name
+        return self.attributes.get("name") or self.__class__.__name__
 
     @name.setter
     def name(self, name):
-        self._name = name
+        self.attributes["name"] = name
 
     @classmethod
     def from_data(cls, data):  # type: (dict) -> Data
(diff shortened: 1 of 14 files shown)


tomvanmele:
@Licini tree tests fail with this change. :class:`compas.datastructures.Tree` stores name and attributes in the data dict and probably overwrites the name implementation already present in `compas.data.Data`

Licini:
I see, will make some modification on this branch in a bit

Licini:
@tomvanmele updated the tree, now the tests pass

src/compas/data/data.py
```diff
@@ -74,7 +74,7 @@ class Data(object):
 
     def __init__(self, name=None):
         self._guid = None
-        self._name = None
+        self.attributes = {}
```
Licini: Since now `attributes` comes to `Data`, the `__init__` and `name` getter and setter at `Datastructure`  seems to be duplicated, should they be removed there?

tomvanmele: i have already removed them but not pushed yet

src/compas/datastructures/datastructure.py
```diff
@@ -8,14 +8,5 @@
 class Datastructure(Data):
     """Base class for all data structures."""
 
-    def __init__(self, name=None, **kwargs):
+    def __init__(self, **kwargs):
         super(Datastructure, self).__init__(**kwargs)
```
Licini: In this case, is this function even need to be redefined?

src/compas/datastructures/tree/tree.py
```diff
@@ -47,16 +43,16 @@ class TreeNode(Data):
...
             "children": {"type": "array", "items": {"$recursiveRef": "#"}},
         },
-        "required": ["name", "attributes", "children"],
+        "required": ["attributes", "children"],
     }
 
     def __init__(self, name=None, attributes=None):
         super(TreeNode, self).__init__(name=name)
-        self.attributes = attributes or {}
+        if attributes:
```
tomvanmele: should it not be
```python
def __init__(self, **kwargs):
    super(TreeNode, self).__init__(**kwargs)
    self._parent = None
    ...
```

tomvanmele: and the same for the tree itself?
and should we not also remove the attributes from the data dict?

Licini: > should it not be
> 
> ```python
> def __init__(self, **kwargs):
>     super(TreeNode, self).__init__(**kwargs)
>     self._parent = None
>     ...
> ```

right, indeed! updated now

Licini: > and the same for the tree itself? and should we not also remove the attributes from the data dict?

You mean like this?
```python
@property
    def data(self):
        return {
            # "attributes": self.attributes,
            "children": [child.data for child in self.children],
        }
```

We would lost track of name then, unless we put it back explicitly.
At least for TreeNode isn't it nice to keep the whole attributes dict so there is a bit more flexibility?

tomvanmele: for the tree node there are two options
1. keep the attributes dict in the data dict and convert the node explicitly to/from data
2. don't include the attributes dict in the data dict and let the encoder/decoder handle to/from (then the attributes are included implicitly)

the file size of 1. is much smaller and is similar to what happens to nodes in a graph and vertices in a mesh. however, there the behaviour is not so explicit since the nodes and vertices are not separate objects.

if a tree node is not meant to exist on its own (like graph nodes and mesh vertices), perhaps we should make sure that they cannot be converted to/from data independently?

Licini: I'm more in favor of option.1. imo, the node can be create independently but there is no need to serialise them outside context of tree. They also don't need guid, because their structural location in the tree makes them naturally unique. Maybe it should even not inherant from `Data` at all... Well this is distracting the purpose of this PR, shall we keep it for now and open another PR for it?

Licini: mmm....  actually... option 2 is also not bad........ well let's have this discussion in another PR...

tomvanmele: currently everything works, so we can merge this and continue the conversation later. but we need to make sure the behaviour is consistent everywhere and 100% transparent.

@gonzalocasas @chenkasirer opinions?

tomvanmele: > mmm.... actually... option 2 is also not bad........ well let's have this discussion in another PR...

i also prefer option 1 btw

Licini: > currently everything works, so we can merge this and continue the conversation later. but we need to make sure the behaviour is consistent everywhere and 100% transparent.
> 
> @gonzalocasas @chenkasirer opinions?

yes agree

Licini (approved):
making the tree thing a Issue, everything else lgtm!

Licini:
Another thing, would it be possible to have something like `to_dict`, outputting same content of `to_jsonstring` but as dict instead of string? Again, doesn't have to be in this PR

PR #1259, opened 2024-01-16
Box collision

petrasvestartas:
Box collision method using a separation plane":
`my_box.has_collision(another_box)`

It helps to speed up searches like rtree, kdtree, since boundingbox collision often surrounds more space.
I often used this first before checking mesh collisions, which are slow.

image

### Checklist


src/compas/geometry/shapes/box.py
```diff
@@ -744,3 +744,52 @@ def contains_points(self, points, tol=1e-6):
...
+
+        Returns
+        -------
+        bool
+            True if the two boxes collide.
+            False otherwise.
+
+        """
+        # get separation plane
+        def GetSeparatingPlane(RPos, axis, self, other):
```
gonzalocasas: `get_separating_plane`

src/compas/geometry/shapes/box.py
```diff
@@ -744,3 +744,52 @@ def contains_points(self, points, tol=1e-6):
...
+                abs((self.frame.xaxis * self.width * 0.5).dot(axis))
+                + abs((self.frame.yaxis * self.depth * 0.5).dot(axis))
+                + abs((self.frame.zaxis * self.height * 0.5).dot(axis))
+                + abs((other.frame.xaxis * other.width * 0.5).dot(axis))
+                + abs((other.frame.yaxis * other.depth * 0.5).dot(axis))
+                + abs((other.frame.zaxis * other.height * 0.5).dot(axis))
+            )
+
+        # compute the obb collision
+        RPos = other.frame.point - self.frame.point
```
gonzalocasas: `RPos` is a local variable, should be lower-cased, with underscores if needed. it's also not clear what `rpos` stands for, i would go explicit, is it `relative_position`?

src/compas/geometry/shapes/box.py
```diff
@@ -744,3 +744,52 @@ def contains_points(self, points, tol=1e-6):
             results[index] = True
 
         return results
+
+    def has_collision(self, other):
```
gonzalocasas: I think this method should be called something with `intersection_...()`, maybe `intersection_with_box()`, to match other existing functions/methods in the `compas.geometry` module and in the surfaces stuff.

gonzalocasas (changes requested):
I think a single collision checking method for a single type is a bit incomplete, and while covering all existing solids will probably be too much, at least `Sphere` also should have an implementation of `intersection_with_sphere` (i believe the implementation is already available in the `intersection_sphere_sphere` method).

Also, unit tests are needed, functionality that is pure code and has no UI dependencies should always be unit tested. Especially if it has potentially plenty of edge cases.

src/compas/geometry/shapes/box.py
petrasvestartas: Thank you for the review.

I changed the style and currently writing tests.

It is only giving output True or False, I thought it belongs to class members that often starts with "is" like "is_parallel", "is_coplanar". Currently I renamed it to "intersection_with_box"

I also have a snippet for `boundingbox `collision, but there is no class, only a method to compute points.

___

It would be great to have a class like `Intersections.py` but for collisions like `Collisions.py` when no geometry is needed just True or False.

gonzalocasas: ahh ok! that's a good point, I'm not sure what could be a good name then.. `is_intersected_by_box(other)`? that's weird and too verbose, isn't it? `is_colliding_with_box(other)`?

/cc @tomvanmele @chenkasirer @Licini @romanarust suggestions?

PR #1267, opened 2024-01-21
Support for rhino8

tomvanmele:

### Checklist


src/compas_rhino8/scene/__init__.py
```diff
@@ -0,0 +1,128 @@
...
+
+
+@plugin(category="drawing-utils", pluggable_name="redraw", requires=["Rhino"])
+def redraw_rhino8():
+    compas_rhino8.redraw()
+
+
+@plugin(category="factories", requires=["Rhino"])
+def register_scene_objects():
+    register(Circle, CircleObject, context="Rhino")
```
Licini: Do we need to say Rhino8 here in context?

Licini: If the objects are completely the same we can perhaps also directly import from `compas_rhino` ?

Licini:
Hell Yeah 😆

src/compas_rhino8/__init__.py
```diff
@@ -0,0 +1,229 @@
+# type: ignore
+import io
+import os
+import compas
+
+__version__ = "2.0.0-beta.2"
```
gonzalocasas: I think the `.bumpversion.cfg` file will need to be updated to reflect this extra instance of version

src/compas_rhino8/scene/__init__.py
gonzalocasas: I would also vote for reducing code duplication wherever possible, by importing from `compas_rhino` or some other way of sharing code.

gonzalocasas (commented):
Having support for CPython in Rhino8 is awesome!

There's a couple of things that I believe need to be considered in more detail:
- The name `compas_rhino8` is a complicated choice: it will be outdated once `rhino9` comes out and the whole namespace will have to change when it probably isn't required. I can't think of a good alternative though: `compas_rhino_cpython` is too long, `compas_rhino_py3` is a bit odd but maybe could work, `compas_rhino_next` could be a temporary namespace if we decided that this will become the main API and we deprecate `compas_rhino` entirely (by adding `_legacy` or `_ipy` suffix to it) at which point we could rename `_next` to be the new `compas_rhino`. This is a big breaking change, and we should do this rather soon, so timing is not the best.
- There's a too-high degree of code duplication, I cannot judge if the current files are copies of `compas_rhino` or re-implementations. If they are copies, we need to find a method to share code or minimize duplication in some way.

tomvanmele:
i understand your points. reasons for the current choice are the following (but that doesn't mean we can't do things differently)

* if we decide to add type annotations (which we can), we cannot reuse any code from the existing package
* the way i understand the internals of Rhino8, we should install COMPAS separately for ipy support and cpy support. so there at least needs to be a separate installer.
* was using this package as a way to clean up the original one (but those changes are actually compatible with older Rhino versions)
* i started a cookiecutter template for CAD integrations and this was just a test of that template :)

tomvanmele:
maybe we can have a short meeting about it at 10? would like to finish this today...

tomvanmele:
About the legacy installation in Rhino 8...

There are now 2 "site-packages" folders for IronPython:
1. `path/to/ManagedPlugins/RhinoDLR_Python.rhp/Lib/site-packages`
2. `path/to/.rhinocode/py27-rh8/Lib/site-packages`

Not sure which one we should use.
I also have the feeling we didn't really use (1) before and just installed directly into `Lib`. So maybe we should change that in any case...

docs/userguide/cad.rhino8.rst
```diff
@@ -2,4 +2,65 @@
...
+
+You need to know the path to the Python executable that comes with Rhino 8.
+On macOS, this is typically something like ``~/.rhinocode/py39-rh8/python3.9``.
+
+.. code-block:: bash
+
+    $ ~/.rhinocode/py39-rh8/python3.9 -m pip install compas
+
+
+Using ``compas_rhino.install_with_pip``
```
Licini, line 53: I would recommend we say this option first, since finding the python executable of Rhino is still rather nuanced.

src/compas_rhino/__init__.py
```diff
@@ -298,9 +382,164 @@ def _get_rhino_grasshopperplugin_path(version):
...
+    return path
+
+
+def _get_default_rhino_cpython_path_windows(version):
+    raise NotImplementedError
+
+
+def _get_default_rhino_cpython_path_mac(version):
+    if version == "8.0":
+        return "{}/.rhinocode/py39-rh8/python3.9".format(os.path.expanduser("~"))
```
Licini: Should we also send a warning when people attempt to do symlink install on Rhino 8, say that we recommend pip install?

src/compas_rhino/utilities/__init__.py
```diff
@@ -7,67 +7,6 @@
...
-    select_lines,
-    select_polyline,
-    select_polylines,
-    select_polygon,
-    select_polygons,
-    select_surface,
-    select_surfaces,
-    select_mesh,
-    select_meshes,
-)
```
Licini, line 70: This feels great 🤣

Licini (approved):
All look good to me! just minor suggestions

src/compas_rhino/__init__.py
tomvanmele: the symlink install is not available for CPython, but it is still the default (and only) procedure for IronPython installations...

tomvanmele:
@Licini could you check the corresponding paths for windows?
@gonzalocasas will also do some testing, but then we already now if you guys get the same result...

tomvanmele:
@gonzalocasas forgot i have parallels :)

paths seems to be the same on Windows...

tomvanmele:
to summarize my rant on slack

* GH CPython scripts work as long as you don't use pluggables. so no scene, but geometry conversions are fine
* the ipy script component in GH works fine, also with the scene
* the paths on Mac and Windows are more or less the same (only the name of the python executable is different)
* pip install works well, but my convenience installer creates issues with numpy
* GH + ipy and Rhino + cpy work well together

tomvanmele:
@gonzalocasas this PR might not be perfect yet, but it doesn't break anything and it does add some specific support for Rhino 8, so how about we just merge this so people can start testing?

PR #1272, opened 2024-01-24
__eq__ on data

Licini:
Just an idea, adding `__eq__` on the level of `Data`, by just comparing the `__data__` dict If not implemented otherwise, what do you guys think?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Added `compas_rhino.objects`.
 * Added `compas_rhino.layers`.
 * Added `compas_rhino.install_with_pip`.
+* Added `__eq__` to `compas.data.Data`.
 
 ### Changed
 
--- a/src/compas/geometry/geometry.py
+++ b/src/compas/geometry/geometry.py
@@ -18,9 +18,6 @@ class Geometry(Data):
     def __init__(self, name=None):
         super(Geometry, self).__init__(name=name)
 
-    def __eq__(self, other):
-        raise NotImplementedError
-
     def __ne__(self, other):
         # this is not obvious to ironpython
         return not self.__eq__(other)
--- a/tests/compas/data/test_data.py
+++ b/tests/compas/data/test_data.py
@@ -14,7 +14,7 @@ def test_string_casting():
 
 
 def test___eq__():
-    class SerilailableClass(Data):
+    class DataClass(Data):
         def __init__(self, int_attr, str_attr, bool_attr, float_attr, list_attr, dict_attr, data_attr):
             self.int_attr = int_attr
             self.str_attr = str_attr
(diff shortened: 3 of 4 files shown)


tomvanmele:
would be good to add some tests for this...

tests/compas/data/test_data.py
```diff
@@ -11,3 +11,42 @@ def __str__(self):
 
     test = TestClass(42)
     assert str(test) == "TestClass 42"
+
+
+def test___eq__():
```
Licini, line 16: @tomvanmele added test now

src/compas/geometry/geometry.py
```diff
@@ -18,9 +18,6 @@ class Geometry(Data):
     def __init__(self, name=None):
         super(Geometry, self).__init__(name=name)
 
-    def __eq__(self, other):
-        raise NotImplementedError
-
```
Licini, line 23: removed to default to comparing the __data__ dict

tests/compas/data/test_data.py
tomvanmele: inequality seems to be the easy way out. does it also work when they should be equal? and what about when only the order of the keys of the dict is different, ...

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Added `compas_rhino.objects`.
 * Added `compas_rhino.layers`.
 * Added `compas_rhino.install_with_pip`.
+* Added `__eq__` to `compas.data.Data`.
```
gonzalocasas, line 15: If this gets added, I would suggest the changelog entry to be more explicit about what is happening, because this was rather extensive consequences, which are not evident by reading this sentence.

gonzalocasas (changes requested):
It's an interesting change, but I'm more leaning towards not adding this to the `compas.data.Data`. I don't think a full data compare is the default expected case for types across the board. This would be equivalent to making all types value-types by default in a language like C#. So, I think this has a has pretty large set of side-effects.

I would personally prefer to put this functionality as opt-in, eg. add a `DataComparable` class that implements the `__eq__` only and if anyway wants a class that is both a child of `Data` and can be compared this way, they can add `DataComparable` to the base class list. BTW, I suspect you'll need to implement `__ne__` for IronPython, because it tends to not figure that one out on its own.

Licini:
> It's an interesting change, but I'm more leaning towards not adding this to the `compas.data.Data`. I don't think a full data compare is the default expected case for types across the board. This would be equivalent to making all types value-types by default in a language like C#. So, I think this has a has pretty large set of side-effects.
> 
> I would personally prefer to put this functionality as opt-in, eg. add a `DataComparable` class that implements the `__eq__` only and if anyway wants a class that is both a child of `Data` and can be compared this way, they can add `DataComparable` to the base class list. BTW, I suspect you'll need to implement `__ne__` for IronPython, because it tends to not figure that one out on its own.

Indeed, I see your concern.

Just to make sure I understand what you mean. For example, would you consider descendants of `Geometry` classes, like `Box`, `Sphere` sort of `DataComparable`? And would you say same for `Mesh`, `Graph` and `Tree`?

Maybe to avoid too much side-effect from the bottom level as well as too many inheritance layers, I can instead just add this to `Geometry` or `Datastructure`. 

I was suggesting this PR simply because I wanted to add some tests for scene serialisation like: `assert scene == scene_deserialised`, and that boils down to comparison on each `Data` item that is in the scene

gonzalocasas:
I totally understand the need, because my code is also full of data comparisons for tests, but I think it's not a good option to do this for all objects only because it makes testing more convenient.

A mesh data compare would be quite a big change, I don't think it even makes much sense that meshes will data compare by default, one always assumes identity comparisons for large data types. So, from that point of view, neither `Geometry` not `Datastructure` should be data comparable by default. 

I'm also concerned about the effect of this on downstream code, libraries outside core that already depend on this.

If the main goal is to ease writing tests, then, I would suggest we write a unit test helper `assert_is_data_equal` and not implement this change at the core at all.

Licini:
> I totally understand the need, because my code is also full of data comparisons for tests, but I think it's not a good option to do this for all objects only because it makes testing more convenient.
> 
> A mesh data compare would be quite a big change, I don't think it even makes much sense that meshes will data compare by default, one always assumes identity comparisons for large data types. So, from that point of view, neither `Geometry` not `Datastructure` should be data comparable by default.
> 
> I'm also concerned about the effect of this on downstream code, libraries outside core that already depend on this.
> 
> If the main goal is to ease writing tests, then, I would suggest we write a unit test helper `assert_is_data_equal` and not implement this change at the core at all.

yep no worries, `assert_is_data_equal` is a good idea too, will do that for now

Licini:
Closing this one now, since we will do a custom test for now.

Issue #1273, opened 2024-01-24
`MissingMemberException` when trying to extend `SceneObject`


chenkasirer:
Been trying to create a new `SceneObject` implementation for something in COMPAS Timber and got this issue where it expects whatever item it wraps to have a `name` attribute. 

I think the culprit is this line: 
https://github.com/compas-dev/compas/blob/5bcf73733262230cc543505daf78270688afd5fd/src/compas/scene/sceneobject.py#L71

It seems it's supposed to take it from `kwargs`, from the instance or use the class name if the other two don't work. But really if the instance doesn't have `name` it always breaks.

Admittedly, I had to ask ChatGPT why this doesn't work cause for the life of me I couldn't figure it out, apparently the stuff around the `or` always gets evaluated..

@Licini maybe `getattr(item, "name", item.__class__.__name__)` instead?

**To Reproduce**
1. Inherit from `SceneObject`
2. Pass something that doesn't have attribute `name` as `item`.

```
Runtime error (MissingMemberException): 'FeatureApplicationError' object has no attribute 'name'

Traceback:
  line 25, in __init__, "C:\Users\ckasirer\AppData\Roaming\McNeel\Rhinoceros\7.0\scripts\compas_rhino\scene\sceneobject.py"
  line 73, in __init__, "C:\Users\ckasirer\AppData\Roaming\McNeel\Rhinoceros\7.0\scripts\compas\scene\sceneobject.py"
  line 15, in __init__, "C:\Users\ckasirer\AppData\Roaming\McNeel\Rhinoceros\7.0\scripts\compas_timber\rhino\scene\featureerrorobject.py"
  line 262, in add, "C:\Users\ckasirer\AppData\Roaming\McNeel\Rhinoceros\7.0\scripts\compas\scene\scene.py"
  line 11, in script
```

**Expected behavior**
Not assume `name` from `item`

**Desktop (please complete the following information):**
 - Rhino / GH

tomvanmele:
@chenkasirer if the item is a data class, how come it doesn't have a name attribute?

chenkasirer:
Does it have to be a `Data`? In my case I'm just trying to implement visualization of some custom errors for visual debugging in GH.

tomvanmele:
the entire setup is based on pairing `compas.data.Data` objects with `compas.scene.SceneObject` objects, so kind of :)

tomvanmele:
[...]

PR #1274, opened 2024-01-25
allow non-Data item in SceneObject

Licini:
Dealing with #1273 . 
`SceneObject` was expected to wrap on items that inherent from `Data`, which will always have a `name`.

This should avoid the error, but not sure yet if it will have other unexpected effects

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Changed `compas.tolerance.Tolerance` into singleton.
 * Changed `compas_rhino.geometry.curves.nursb.RhinoNurbsCurve` to use private data API.
 * Changed `compas_rhino.geometry.surfaces.nursb.RhinoNurbsSurface` to use private data API.
+* Changed `compas.scene.SceneObject` to allow usage of `Data` items.
 
 ### Removed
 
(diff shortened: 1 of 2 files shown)


Licini:
If we don't want to allow this, I can change the PR instead to raise a proper TypeError at construction of SceneObject. Let me know which you guys prefer.

chenkasirer:
I reported this cause it seemed like something that may have gotten overlooked. I can see though why you'd require whatever you're wrapping to be `Data`. I'd just make it clear in a `SceneObject` developer guide, then I'd say there's not even a need for raising an error, we don't normally enforce typing.. Your call.

Licini:
@tomvanmele any opinion on this?

tomvanmele:
it would keep the current Data/SceneObject pairing, at least for now, but indeed properly document it...

gonzalocasas:
If there's a conceptual reason to keep them paired, it should definitely be documented. I think also failing with a more descriptive error would help, if this is the way we're going for.

tomvanmele:
currently there is some duplication between the behaviour of the scene itself and its tree attribute. when this behaviour is merged, i assume the guid of the data object will play a role in keeping track of things. in that context, the name of the item is not important. it might even be better to not implicitly reuse the name of the item, but instead make the user assign a name to the scene object explicitly.

in any case, a scene should be serialisable. so the default behaviour in the core library should be that the items implement the data interface.

what downstream users do with this is of course up to them. in that regard, rather than saying that an item should be a data item, perhaps we should define the required behaviour of an item instead, and leave everything up to the downstream implementations...

how does that sound?

gonzalocasas:
> in any case, a scene should be serialisable. so the default behaviour in the core library should be that the items implement the data interface.

That's the key point. I'm sold. 😄👍

Licini:
I think we have an agreement. Will make later a PR for documenting the behavior, closing this now.

PR #1275, opened 2024-01-25
Scene.draw etc.

Licini:
@tomvanmele Changes you requested:

- Moved registration to `get_scene_object_cls` so it will always be called.
- Renamed `redraw` to `draw`, and kept it as alias.
- Added pluggables `predraw` and `postdraw`.
- Also raise TypeError if item is not an `Data`, as alternative solution to #1274 .

### DIFF
--- a/src/compas/scene/context.py
+++ b/src/compas/scene/context.py
@@ -31,7 +31,7 @@ clear.__pluggable__ = True
 
 @pluggable(category="drawing-utils")
 def before_draw():
-    """Pluggable to perform operations before drawing the scene.
+    """Pluggable to perform operations before drawing the scene. This function is automatically called in the beginning of `compas.scene.Scene.draw()`.
 
     Returns
     -------
@@ -46,8 +46,7 @@ before_draw.__pluggable__ = True
 
 @pluggable(category="drawing-utils")
 def after_draw(drawn_objects):
-    """Pluggable to perform operations after drawing the scene.
-
+    """Pluggable to perform operations after drawing the scene. This function is automatically called at the end of `compas.scene.Scene.draw()`.
     Parameters
     ----------
     drawn_objects : list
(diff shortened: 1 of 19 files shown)


src/compas/scene/scene.py
```diff
@@ -302,8 +306,12 @@ def redraw(self):
         if drawn_objects:
             redraw()
 
+        postdraw()
+
         return drawn_objects
 
+    redraw = draw
```
Licini: But to be honest, I would even vote for not keep this alias, to distinct it from the redraw pluggable

tomvanmele:
do we even need the redraw pluggable? rhino's redraw can be triggered by the postdraw hook...

src/compas/scene/scene.py
gonzalocasas: Agree

src/compas/scene/context.py
```diff
@@ -26,6 +26,22 @@ def redraw():
 redraw.__pluggable__ = True
 
 
+@pluggable(category="drawing-utils")
+def predraw():
```
gonzalocasas: Docstrings to explain what these do?

src/compas/scene/context.py
```diff
@@ -26,6 +26,22 @@ def redraw():
...
+    pass
+
+
+predraw.__pluggable__ = True
+
+
+@pluggable(category="drawing-utils")
+def postdraw():
```
gonzalocasas: If these are hooks, I would call them `before_draw` and `after_draw`

gonzalocasas (approved):
LGTM

src/compas/scene/context.py
tomvanmele: @Licini and please add them to `__init__` and to the api docs.

Licini:
@tomvanmele All adjusted, take a final look?

tomvanmele (approved):

PR #1278, opened 2024-01-26
allow list of keys for showing vertices/edges/faces

Licini:
Allow drawing selection of vertices/edges/faces on `MeshObject`
```python
from compas.datastructures import Mesh
from compas.scene import Scene

mesh = Mesh.from_polyhedron(8)

vertices = mesh.face_vertices(0)
edges = mesh.face_halfedges(0)

scene = Scene()

scene.add(mesh, show_vertices=vertices, show_edges=edges)

scene.draw()
```

image

PR #1283, opened 2024-01-30
Cellnetworkobject

Licini:
Adding `CellNetworkObject`s currently based on #1278 , will rebase after it's merged.


tomvanmele:
@romanarust since you are the expert on cell networks, it would be good if you could get involved here...

src/compas_rhino/scene/volmeshobject.py
```diff
@@ -123,33 +128,35 @@ def clear_facelabels(self):
...
-            The default color is :attr:`VolMeshObject.default_cellcolor`.
-
         Returns
         -------
         list[System.Guid]
             The GUIDs of the created Rhino objects.
             Every cell is drawn as an individual mesh.
 
         """
-        self._guids = self.draw_cells(cells=cells, color=color)
```
romanarust, line 152: IMO the default `draw` function should draw the faces and not the cells. This would cover faces without cells and avoid creating duplicate faces.

romanarust (approved):
I have only one comment on the general draw function.

The rest LGTM!

tomvanmele:
@Licini can we finish this up and move it out of "Draft"?

PR #1289, opened 2024-01-31
Add some more __str__ function to geometry classes for pretty print

### LINKED ISSUE #1287: Add __str__ to all geometry classes for pretty print
# Feature Request

I found the pretty-print `__str__` function added by @gonzalocasas in (https://github.com/compas-dev/compas/pull/1234) helps a lot when making example files. Should we add this to other geometry classes?

I can help make a PR if we are okay with this. @tomvanmele

yck011522:
Partially Fixes #1287 

### Checklist


src/compas/geometry/frame.py
```diff
@@ -109,6 +109,14 @@ def __repr__(self):
             self.yaxis,
         )
 
+    def __str__(self):
+        return "{0}(point={1}, xaxis={2}, yaxis={3})".format(
+            type(self).__name__,
+            self.point.__str__(),
+            self.xaxis.__str__(),
+            self.yaxis.__str__(),
```
gonzalocasas: I would call `str()` instead of invoking the dunder method directly:
```suggestion
            str(self.point),
            str(self.xaxis),
            str(self.yaxis),
```

src/compas/geometry/plane.py
```diff
@@ -76,6 +76,13 @@ def __repr__(self):
             self.normal,
         )
 
+    def __str__(self):
+        return "{0}(point={1}, normal={2})".format(
+            type(self).__name__,
+            self.point.__str__(),
```
gonzalocasas: same as above

src/compas/geometry/pointcloud.py
```diff
@@ -57,6 +57,9 @@ def __init__(self, points, name=None):
     def __repr__(self):
         return "{0}(points={1!r})".format(type(self).__name__, self.points)
 
+    def __str__(self):
+        return "{0}(len(points)={1})".format(type(self).__name__, [point.__str__() for point in self.points])
```
gonzalocasas: shouldn't this be having the `len(self.points)` instead? i.e.
```suggestion
        return "{0}(len(points)={1})".format(type(self).__name__, len(self.points))
```

src/compas/geometry/pointcloud.py
```diff
@@ -245,11 +248,9 @@ def from_box(cls, box, n):
...
 
         """
-        points = box.points
-        x, y, z = zip(*points)
-        xmin, xmax = min(x), max(x)
-        ymin, ymax = min(y), max(y)
-        zmin, zmax = min(z), max(z)
+        xmin, xmax = box.xmin, box.xmax
+        ymin, ymax = box.ymin, box.ymax
+        zmin, zmax = box.zmin, box.zmax
```
gonzalocasas, line 252: This change is not related to the str stuff, is it?

src/compas/geometry/polygon.py
```diff
@@ -103,6 +103,9 @@ def __init__(self, points, name=None):
     def __repr__(self):
         return "{0}(points={1!r})".format(type(self).__name__, self.points)
 
+    def __str__(self):
+        return "{0}(points={1})".format(type(self).__name__, [point.__str__() for point in self.points])
```
gonzalocasas: switch to `str()` as above

src/compas/geometry/polyhedron.py
```diff
@@ -9,6 +9,7 @@
 from compas.geometry import Polygon
 from compas.geometry import Point
 from compas.geometry import Line
+from compas.tolerance import TOL
 from .geometry import Geometry
```
gonzalocasas, line 14: since you're very close, could you make this import absolute instead? (unless there's any reason for the relative one, but I don't see any

src/compas/geometry/polyhedron.py
```diff
@@ -270,7 +278,7 @@ def __or__(self, other):
 
     @property
     def vertices(self):
-        if not self._vertices:
+        if self._vertices is None:
```
gonzalocasas: In this case, I think the falsey-ness check (`if not self._vertices`) is a better option

src/compas/geometry/polyhedron.py
```diff
@@ -280,7 +288,7 @@ def vertices(self, vertices):
 
     @property
     def faces(self):
-        if not self._faces:
+        if self._faces is None:
```
gonzalocasas: same

gonzalocasas (commented):
I added a few comments but generally looking good!

tomvanmele:
@yck011522 please keep the PR focused on the implementation of `__str__`. adding other stuff will just drag out the review process...

src/compas/geometry/pointcloud.py
yck011522: The change is there because I couldn't actually run tests without fixing the broken functions.

yck011522:
> @yck011522 please keep the PR focused on the implementation of `__str__`. adding other stuff will just drag out the review process...

@tomvanmele The associated changes are there only because those functions were broken to the point that prevented me to test out the new `__str__` functios.

src/compas/geometry/polyhedron.py
yck011522: This caused an error message on a quick test. The error message is 
```
The truth value of an array with more than one element is ambiguous.
```

src/compas/geometry/pointcloud.py
yck011522: Thanks. I had half the intention there.

src/compas/geometry/plane.py
yck011522: yep

gonzalocasas (approved):
sorry, this slipped through the cracks... it lgtm!

gonzalocasas:
@yck011522 could you fix the conflict so that we can merge?

CHANGELOG.md
```diff
@@ -268,6 +268,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Changed `compas.scene.VolMesh`'s `show_vertices`, `show_edges`, `show_faces`, `show_cells` to optionally accept a sequence of keys.
 * Fixed missing implementation of `Sphere.base`.
 * Fixed bug in `intersection_sphere_sphere`.
+* Changed the `__str__` of `compas.geometry.Frame`, `compas.geometry.Plane`, `compas.geometry.Polygon`, `compas.geometry.Polyhedron`, `compas.geometry.Quaternion` to use a limited number of decimals (determined by `Tolerance.PRECISION`). Note: `__repr__` will instead maintain full precision.
```
gonzalocasas: Could you update this? Because after the merge it's way down in the file, changing an already released version

gonzalocasas (approved):
Besides the changelog fix it lgtm

yck011522:
@tomvanmele @gonzalocasas 
I notice that the new check called codecov is not passing. The checker seems to want every line of code being included in test. How should I handle this? Should I make the compliant? It seems like some of the tests will be excessive if every line needs to be in CI.

yck011522:
I was trying to add some more tests to cover the __str__ and __repr__ functions. However I realized that the underlying mechanism of formatting the numbers in Point is based on `TOL.format_number` and its behavior is unstable across different environments. 

So I have decided to open another PR to investigate the problem and this PR will not have those tests included.

gonzalocasas:
> @tomvanmele @gonzalocasas 
> I notice that the new check called codecov is not passing. The checker seems to want every line of code being included in test. How should I handle this? Should I make the compliant? It seems like some of the tests will be excessive if every line needs to be in CI.

Having coverage metrics is good, but we don't intend to be obsessive about it

tomvanmele:
@yck011522 the test is just feedback about coverage. it will not block the PR from being merged...

PR #1304, opened 2024-02-26
Fix bug related to value of degree in Nurbs Curves

tomvanmele:
Maximum degree should be "number_of_control_points - 1".

### Checklist

### DIFF
--- a/src/compas_rhino/geometry/curves/__init__.py
+++ b/src/compas_rhino/geometry/curves/__init__.py
@@ -1,21 +1,19 @@
-from .curve import RhinoCurve  # noqa: F401
+from .curve import RhinoCurve
 from .nurbs import RhinoNurbsCurve
 
-from compas.geometry import Curve
-from compas.geometry import NurbsCurve
 from compas.plugins import plugin
 
 
 @plugin(category="factories", requires=["Rhino"])
 def new_curve(cls, *args, **kwargs):
-    curve = super(Curve, cls).__new__(cls)
+    curve = object.__new__(RhinoCurve)
     curve.__init__(*args, **kwargs)
     return curve
 
 
 @plugin(category="factories", requires=["Rhino"])
 def new_nurbscurve(cls, *args, **kwargs):
-    curve = super(NurbsCurve, cls).__new__(cls)
+    curve = object.__new__(RhinoNurbsCurve)
     curve.__init__(*args, **kwargs)
     return curve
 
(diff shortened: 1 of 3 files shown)


src/compas_rhino/geometry/curves/__init__.py
```diff
@@ -1,19 +1,21 @@
...
 
+from compas.geometry import Curve
+from compas.geometry import NurbsCurve
 from compas.plugins import plugin
 
 
 @plugin(category="factories", requires=["Rhino"])
 def new_curve(cls, *args, **kwargs):
-    curve = object.__new__(RhinoCurve)
+    curve = super(Curve, cls).__new__(cls)
```
tomvanmele: in my opinion this handles the inheritance chain better...

src/compas_rhino/geometry/curves/__init__.py
```diff
@@ -1,19 +1,21 @@
...
     curve.__init__(*args, **kwargs)
     return curve
 
 
 @plugin(category="factories", requires=["Rhino"])
 def new_nurbscurve(cls, *args, **kwargs):
-    curve = object.__new__(RhinoNurbsCurve)
+    curve = super(NurbsCurve, cls).__new__(cls)
```
tomvanmele: same here

tomvanmele:
ping

src/compas_rhino/geometry/curves/__init__.py
chenkasirer: when I do this in `Brep` I get an instance of `compas.geometry.Brep` even in Rhino, which is unexpected in my opinion.

tomvanmele: i think it would be better if it returns `compas.geometry.NurbsCurve` and not the Rhino version, because then the returned object matches the object you tried to create (and its documentation), but this requires that the Rhino implementation completely aligns with the pluggable API. and since that is currently not the case, it would create unexpected results.

was also included more or less by accident so we can ignore the issue for now...

chenkasirer (approved):
LGTM!

src/compas_rhino/geometry/curves/__init__.py
chenkasirer: If it returns a `NurbsCurve` then every single method has to be a pluggable/plugin, that would be really cumbersome, no?

PR #1307, opened 2024-03-02
Merging SceneObjectNode and SceneObject

Licini:
The relation between `SceneObjectNode` and `SceneObject` was too convoluted. This PR merge them into one by having `SceneObject` inherent from `TreeNode`. This simplifies a lot on operations to access parent/ children object or a tree that an object is attachted too. While the most-outer level user APIs stays untouched.  This should also make it easier from creating a `compas_model` object.

```python
from compas.geometry import Sphere
from compas.geometry import Cylinder
from compas.geometry import Sphere



box = Box.from_width_height_depth(1, 1, 1)
cylinder = Cylinder(1, 1, Frame.worldXY())
sphere = Sphere(1)

scene = Scene()
scene.clear()

obj = scene.add(box).add(cylinder).add(sphere)

scene.print_hierarchy()


print(obj.tree)
```
```
└── <TreeNode: ROOT>
    └── <GeometryObject: Box>
        └── <GeometryObject: Cylinder>
            └── <GeometryObject: Sphere>
<Tree with 4 nodes>
```

### Checklist

### DIFF
--- a/src/compas/scene/scene.py
+++ b/src/compas/scene/scene.py
@@ -1,6 +1,5 @@
 import compas.datastructures  # noqa: F401
 import compas.geometry  # noqa: F401
-from compas.data import Data
 from compas.datastructures import Tree
 from compas.datastructures import TreeNode
 
@@ -147,4 +146,3 @@ class Scene(Tree):
         after_draw(drawn_objects)
 
         return drawn_objects
-
--- a/src/compas/scene/scene.py
+++ b/src/compas/scene/scene.py
@@ -130,6 +130,14 @@ class Scene(Data):
         if isinstance(item, SceneObject):
             sceneobject = item
         else:
+            if "context" in kwargs:
+                if kwargs["context"] != self.context:
+                    raise Exception(
+                        "Object context should be the same as scene context: {} != {}".format(
+                            kwargs["context"], self.context
+                        )
+                    )
+                del kwargs["context"]  # otherwist the SceneObject receives "context" twice, which results in an error
             sceneobject = SceneObject(item, context=self.context, **kwargs)  # type: ignore
         self.tree.add(sceneobject, parent=parent)
         return sceneobject
--- a/src/compas/scene/scene.py
+++ b/src/compas/scene/scene.py
@@ -9,81 +9,6 @@ from .context import detect_current_context
 from .sceneobject import SceneObject
 
 
-class SceneObjectNode(TreeNode):
-    """A node representing a scene object in a scene tree. The SceneObjectNode should only be used internally by the SceneTree.
-
-    Parameters
-    ----------
-    object : :class:`compas.scene.SceneObject`
-        The scene object associated with the node.
-
-    Attributes
-    ----------
-    name : str
-        The name of the node, same as the underlying scene object.
-    object : :class:`compas.scene.SceneObject`
-        The scene object associated with the node.
-    parentobject : :class:`compas.scene.SceneObject`
-        The scene object associated with the parent node.
-    childobjects : list[:class:`compas.scene.SceneObject`]
-        The scene objects associated with the child nodes.
-
-    """
-
-    @property
-    def __data__(self):
-        return {
-            "item": str(self.object.item.guid),
-            "settings": self.object.settings,
-            "children": [child.__data__ for child in self.children],
-        }
-
-    @classmethod
-    def __from_data__(cls, data):
-        raise TypeError("SceneObjectNode cannot be created from data. Use Scene.__from_data__ instead.")
-
-    def __init__(self, sceneobject, name=None):
-        super(SceneObjectNode, self).__init__(name=name)
-        self.object = sceneobject
-
-    @property
-    def name(self):
-        if self.object:
-            return self.object.name
-
-    @property
-    def parentobject(self):
-        if self.parent and isinstance(self.parent, SceneObjectNode):
-            return self.parent.object
-        return None
-
-    @property
-    def childobjects(self):
-        return [child.object for child in self.children]
-
-    def add_item(self, item, **kwargs):
-        """Add an child item to the node.
-
-        Parameters
-        ----------
-        item : :class:`compas.data.Data`
-            The item to add.
-        **kwargs : dict
-            Additional keyword arguments to create the scene object for the item.
-
-        Returns
-        -------
-        :class:`compas.scene.SceneObject`
-            The scene object associated with the item.
-
-        """
-        sceneobject = SceneObject(item, **kwargs)
-        node = SceneObjectNode(sceneobject)
-        self.add(node)
-        sceneobject._node = node  # type: ignore
-        return sceneobject
-
-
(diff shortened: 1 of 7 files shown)


tomvanmele:
i don't think this helps with the model problem. from an item you can access the parent node and from the parent node you can reach the tree. but you still can't reach the scene from the tree.

perhaps it would make sense to make `Scene` a subclass from `Tree`? might feel more logical since a scene object is a tree node. the problem of reaching the scene is them immediately solved.

src/compas/scene/sceneobject.py
```diff
@@ -170,10 +167,12 @@ def add(self, item, **kwargs):
...
             If the scene object does not have an associated scene node.
         """
-        if self.node:
-            return self.node.add_item(item, **kwargs)
+        if isinstance(item, SceneObject):
+            sceneobject = item
         else:
-            raise ValueError("Cannot add items to a scene object without a node.")
+            sceneobject = SceneObject(item, **kwargs)
+        super().add(sceneobject)
```
tomvanmele: doesn't work in IronPython...

src/compas/scene/scene.py
```diff
@@ -261,8 +114,14 @@ def add(self, item, parent=None, **kwargs):
...
-        sceneobject = SceneObject(item, context=self.context, **kwargs)
-        self.tree.add_object(sceneobject, parent=parent)
+
+        parent = parent or self.tree.root
+
+        if isinstance(item, SceneObject):
+            sceneobject = item
+        else:
+            sceneobject = SceneObject(item, context=self.context, **kwargs)
+        self.tree.add(sceneobject, parent=parent)
```
tomvanmele: before, the problem was here. this might work now for the model...

tomvanmele:
i added a few changes that make the scene work for recursive addition of objects in, for example a model.
added the type annotations for debugging.
we can remove them again...

tomvanmele:
i removed viewer detection and the viewer instance attribute on the scene. as explained, i think we should rather do the opposite...

Licini:
> i don't think this helps with the model problem. from an item you can access the parent node and from the parent node you can reach the tree. but you still can't reach the scene from the tree.
> 
> perhaps it would make sense to make `Scene` a subclass from `Tree`? might feel more logical since a scene object is a tree node. the problem of reaching the scene is them immediately solved.

I also think making `Scene` inherent from `Tree` is a good idea. Especially when SceneObject is the node. Will give this a try now on this branch

Licini:
@tomvanmele 

1. How about now? `Scene` is now inheriting directly from `Tree`.
2. Regarding documenting the kwargs. Should we also explicitly write all of them out then pass on to super init, so it can be suggested pylance? For example:
```python
def __init__(kwag_x=None, kwag_y=None):
    super(XXX, self).__init__(kwag_x=kwag_x, kwag_y=kwag_y)
```

tomvanmele:
regarding`**kwargs`...

in my opinion this should only be used when additional named arguments can be expected that are not known to the current class and need to be passed on to the class ancestors. arguments that are expected and used in the current scope should always be listed, processed, and documented explicitly...

what i think we should most certainly NOT do is the following:

```python
def __init__(self, **kwargs):
    super().__init__(**kwargs)
    self.a = kwargs.get('a')
    self.b = kwargs.get('b')
```

i would also vote for using ONLY named arguments in `__init__` functions in the future, because passing of arguments to parent classes is complicated when there is a mixture of positional and named arguments...

src/compas/scene/sceneobject.py
```diff
@@ -7,21 +7,40 @@
...
 from functools import reduce
 from operator import mul
 
 
-class SceneObject(object):
+# if we know the parameters
+# we should document them
+# and not just make everything kwargs
+# not only is this not very helpful for the user
+# it does not allow for proper propagation of parameters across the inheritance chain
```
tomvanmele: perhaps remove this?

tests/compas/scene/test_scene.py
```diff
@@ -80,34 +80,34 @@ def test_sceneobject_auto_context_discovery(mocker):
...
+    #     context.ITEM_SCENEOBJECT["Viewer"] = {FakeItem: FakeViewerSceneObject}
+    #     context.ITEM_SCENEOBJECT["Plotter"] = {FakeItem: FakePlotterSceneObject}
 
-        item = FakeSubItem()
-        sceneobject = SceneObject(item)
+    #     item = FakeSubItem()
+    #     sceneobject = SceneObject(item)
 
-        assert isinstance(sceneobject, FakeViewerSceneObject)
+    #     assert isinstance(sceneobject, FakeViewerSceneObject)
```
tomvanmele: let's remove this, no?

tests/compas/scene/test_scene.py
```diff
@@ -80,34 +80,34 @@ def test_sceneobject_auto_context_discovery(mocker):
...
 
     def test_sceneobject_auto_context_discovery_no_context(mocker):
-        mocker.patch("compas.scene.context.is_viewer_open", return_value=False)
+        # mocker.patch("compas.scene.context.is_viewer_open", return_value=False)
```
tomvanmele: same

src/compas/scene/sceneobject.py
```diff
@@ -148,8 +189,26 @@ def contrastcolor(self):
...
+            "opacity": self.opacity,
+            "show": self.show,
+        }
+
+        if self.frame:
+            settings["frame"] = self.frame
+        if self.transformation:
+            settings["transformation"] = self.transformation
+
+        return settings
```
tomvanmele: perhaps the existence of this container should be documented properly

src/compas/scene/sceneobject.py
```diff
@@ -7,21 +7,40 @@
...
-    **kwargs : dict
-        Additional keyword arguments for constructing SceneObject.
+    name
+    color
+    opacity
+    show
+    frame
+    transformation
+    context
+    **kwargs
```
tomvanmele: still need to fill these in...

Licini:
> regarding`**kwargs`...
> 
> in my opinion this should only be used when additional named arguments can be expected that are not known to the current class and need to be passed on to the class ancestors. arguments that are expected and used in the current scope should always be listed, processed, and documented explicitly...
> 
> what i think we should most certainly NOT do is the following:
> 
> ```python
> def __init__(self, **kwargs):
>     super().__init__(**kwargs)
>     self.a = kwargs.get('a')
>     self.b = kwargs.get('b')
> ```
> 
> i would also vote for using ONLY named arguments in `__init__` functions in the future, because passing of arguments to parent classes is complicated when there is a mixture of positional and named arguments...

Sounds good, will make everything named argument in the scope of this PR and make a separate one after for rest of places. And will fix all the docs and remove the code commented out

Licini:
@tomvanmele cleaned up, shall we merge?

Licini:
@tomvanmele let's merge this?

PR #1313, opened 2024-03-26
Rhino brep trimmed

chenkasirer:
* `RhinoBrep.trimmed` returns now a single brep as a result instead of a list of breps. If no result available a `BrepTrimmingError` is raised.

This fixes a discrepancy between the `RhinoBrep` implementation and the documented behavior. 

### Checklist


jf---:
@gonzalocasas || @tomvanmele this looks good, only thing that comes to mind if handling precision is following the new `compas.precision` idiom? 

The description is not super clear -- @chenkasirer this is fusing a nr of `faces`, right?
In that sense the method is a little tricky to understand, and if topological changes (if faces are fused) are invoked, a little more on the intent of the method is needed I think. fused faces when invoking `trimmed` is semantically not crystal clear I feel.

tomvanmele (dismissed):
LGTM

tomvanmele:
actually, sorry i take that back, tolerance handling is indeed wrong...

src/compas_rhino/geometry/brep/brep.py
```diff
@@ -394,26 +404,38 @@ def trimmed(self, plane, tolerance=TOLERANCE):
...
-        for result in results:
-            capped = result.CapPlanarHoles(TOLERANCE)
-            if capped:
-                result = capped
-            breps.append(RhinoBrep.from_native(result))
-        return breps
+        if not results:
+            raise BrepTrimmingError("Trim operation ended with no result")
+        result = results[0]
+        capped = result.CapPlanarHoles(TOLERANCE)
```
tomvanmele: in the signature, `tolerance=None`. then `tolerance = tolerance or TOL.absolute`, with at the top (if it is not already there) `from compas.tolerance import TOL`

tomvanmele: and then, finally, `capped = result.CapPlanarHoles(tolerance)`

tomvanmele (changes requested):
use of tolerance should be updated

tomvanmele:
@chenkasirer ping

src/compas_rhino/geometry/brep/brep.py
chenkasirer: @tomvanmele done.
Just curious why not `tolerance=TOL.absolute` in signature? as it returns a `float` (immutable)

chenkasirer:
> @gonzalocasas || @tomvanmele this looks good, only thing that comes to mind if handling precision is following the new `compas.precision` idiom?
> 
> The description is not super clear -- @chenkasirer this is fusing a nr of `faces`, right? In that sense the method is a little tricky to understand, and if topological changes (if faces are fused) are invoked, a little more on the intent of the method is needed I think. fused faces when invoking `trimmed` is semantically not crystal clear I feel.

Hey @jf--- !
Not sure what you mean by fusing faces here, could you explain? `Brep.trimmed` just returns all the results of a trim operation. trimming with plane would then typically result in two breps (one on each side of the plane), but maybe I'm missing something.

tomvanmele (approved):
LGTM

src/compas_rhino/geometry/brep/brep.py
tomvanmele: that would have also been an option indeed. as long as it is not a hard coded tolerance value...

PR #1314, opened 2024-03-28
Use `__str__` to replaced `print_hierarchy`

Licini:
According to @chenkasirer there are some discussion about remove all `print` in core library, instead use `__str__` for this type of functionalities. 
https://github.com/BlockResearchGroup/compas_model/pull/31#discussion_r1519911684

I made this change for `Tree`, let me know what you guys think.

### Checklist


jf---:
this looks great, but curious what's @tomvanmele @Licini  take on __str__ vs __repr__: could __repr__ just invoke __str__?

tomvanmele:
the point of `__repr__` is to allow for (re)construction of an object using `eval` of its string "repr"esentation. in the case of making something print-friendly we should use `__str__` in my opinion. if i remember correctly, we only implement `__repr__` if reconstruction is indeed possible...

jf---:
right, __repr(oduction)__ its a neat feature, so __repr__ can echo __str__ when reconstruction isn't viable

src/compas/datastructures/tree/tree.py
```diff
@@ -450,22 +449,27 @@ def print_hierarchy(self, max_depth=None):
...
+        hierarchy = {"string": ""}
+
+        def traverse(node, hierarchy, prefix="", last=True, depth=0):
 
             if max_depth is not None and depth > max_depth:
                 return
 
             connector = "└── " if last else "├── "
-            print("{}{}{}".format(prefix, connector, node))
+            hierarchy["string"] += "{}{}{}\n".format(prefix, connector, node)
```
gonzalocasas: why do you pass around a dictionary with a single, hard-coded key?

gonzalocasas (approved):
lgtm! only one small q in comments

chenkasirer:
@Licini sorry to ping, I'd just like to merge https://github.com/BlockResearchGroup/compas_model/pull/31 soon

tomvanmele:
@Licini would indeed be good to finish this...

src/compas/datastructures/tree/tree.py
Licini: This was to make the string a mutable object (like a pointer). In python 3 i can simply use `nonlocal` and just say `hierarchy_string`, but that option is not available for 2.7. Maybe now looks better with the list?

Licini:
@chenkasirer @gonzalocasas Sorry for late update on this folks. But now I can declare this is done 😂. If you guys give me a go I will merge now

src/compas/datastructures/tree/tree.py
```diff
@@ -450,22 +449,27 @@ def print_hierarchy(self, max_depth=None):
 
         Returns
         -------
-        None
+        str
+            The spatial hierarchy of the tree.
```
tomvanmele: fwiw, i don't agree with this. why would anyone expect `Tree.hierarchy` to return a string?

tomvanmele: if the only purpose of this method is to be a helper for `__str__`, i think it should be named differently and made private.

Licini: How about something more expressive like `get_hierarchy_string` ? Tho I'm more on the side for still keeping it public because I call this function a lot with different `max_depth` for very deep trees, I can imagine people might also want to do something like this

Licini: @tomvanmele any comments on this? would like to resolve it

tomvanmele (approved):

Issue #1208, opened 2023-11-09
Spherical Linear Interpolation Function (slerp) for Frames


yck011522:
# Feature Request

In order to develop robotic motion planners, I want a slerp function between two frames so that I can create interpolated frames between two frames that have not only positional differences but also orientation differences.

## Details

**Is your feature request related to a problem? Please describe.**
Given two frames with different origin positions and orientations. I want to interpolate between them linearly, controlled by a single parameter. Given a series of uniformly spaced parameters, the resulting set of interpolated Frames should have equal distances (origin points) and natural-looking gradual orientation change.

**Describe the solution you'd like**

I would imagine a function signature like this:
`Frame.slerp_interpolation(start_frame, end_frame, parameter) -> Frame`
where: 
`parameter` is a float between 0.0 and 1.0

**Additional context**
I do not know how this is done properly, but I believe animation software has algorithms like this where the position and orientation of the Frames are interpolated separately and combined later. The origin position can be a simple interpolation and the orientation is interpolated by slerp in quaternion representation. https://en.wikipedia.org/wiki/Slerp

yck011522:
https://docs.unity3d.com/ScriptReference/Quaternion.Slerp.html

Maybe this provides some idea about the slerp function

https://github.com/Unity-Technologies/UnityCsReference/blob/d0e1a5b25e8d3c8e1b4e6baf0073032d97254a0c/Editor/Mono/Utils/MathUtils.cs#L222-L262

yck011522:
https://splines.readthedocs.io/en/latest/rotation/slerp.html

This webpage explains the importance of Slerp having the property of constant angular velocity.

petrasvestartas:
@yck011522  this can be useful for slerp:

**t - [0.0, 1.0]**

```python
 def interpolate_frames(frame0, frame1, t):
                        """
                        Interpolate smoothly between two frames using Slerp and Lerp.

                        Parameters:
                            frame0 (Frame): The starting frame.
                            frame1 (Frame): The ending frame.
                            t (float): Interpolation parameter between 0 and 1.

                        Returns:
                            Frame: The interpolated frame.
                        """
                        # Perform linear interpolation for the position (point) component
                        point = Point(*frame0.point) * (1 - t) + Point(*frame1.point) * t

                        # Perform spherical linear interpolation (Slerp) for the orientation (quaternion) component
                        q0 = Quaternion.from_frame(frame0)
                        q1 = Quaternion.from_frame(frame1)

[...]

PR #1316, opened 2024-04-06
implements slerp (#1208)

jf---:
implements interpolation between frame(s)

~the updated scipy dependency is due to the Slerp being added in 1.2~
~see: https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.transform.Slerp.html~

- [x] New feature 

### Checklist


jf---:
Wow

> Running flake8 python linter...

It's not encouraging to see a build fail on seriously minor formatting.
@tomvanmele why not format using `black` an action like below could format code with black and perform an autocommit:

```

on: [pull_request, push]

jobs:
  black:
    runs-on: ubuntu-latest
    steps:
      - name: Check out code
        uses: actions/checkout@v2
        with:
          ref: ${{ github.head_ref }}

      - name: Set up Python
        uses: actions/setup-python@v2
        with:
          python-version: '3.8'

      - name: Install Black
        run: pip install black

      - name: Run Black Check
        run: black --check .

      - name: Run Black
        run: black .

      - name: Commit changes
        uses: stefanzweifel/git-auto-commit-action@v4
        with:
          commit_message: Apply Black Formatting
          branch: ${{ github.head_ref }}
          file_pattern: .
```

or a precommit:

```
pip install pre-commit
```

add a .pre-commit-config.yaml

```
repos:
  - repo: https://github.com/psf/black
    rev: stable  # Use a specific version of Black or "stable" for the latest stable release
    hooks:
      - id: black
        language_version: python3  # Adjust this to your project's Python version
```

and hit 

```
pre-commit install
```

jf---:
```
Running flake8 python linter...
Running black python linter...
```

@tomvanmele this is too much, can I suggest to use either one?
flake8 and black are mutex, so this isn't making sense 
moreover, what's really bad here is linting is not a separate concern ( like `verify-pr-checklist` is ) 
so you don't want linting to obscure where the testsuite passed.

to conflate functional test and micromanaging stylistic issues is pretty wicked

let's make it easy that PR's are in the green, so you can hit that sweet merge button relentlessly

tomvanmele:
will check with @Licini because this has been replaced on many repos by `ruff`. i guess we have not yet gotten to the main `compas` repo yet.

that said though, it is rather important that code is formatted consistently. otherwise half of all PRs end up being about formatting diffs. unfortunately, not all formatting errors can be auto-corrected. therefore, we have left it up to the contributor to make sure this is taken care of as part of their local development process.

will have a look if the process can be simplified when we make the switch to `ruff` on the main repo...

jf---:
yeah `ruff` seems the new `black`

well I suppose my gripe is the conflation of linting & testing

I don't contest the value of consistent formatting: especially when it comes to the diffs in PR's

any thoughts on the .pre-commit route?

tomvanmele:
don't know the tool. @Licini could you have a look?

tomvanmele:
we should also switch `compas` to `pyproject.toml` and the new linting/formatting/sorting setup...

jf---:
@gonzalocasas & @tomvanmele -- wait a minute this is quite upsetting: the entire CI build grinds to a halt, because of some  trivial formatting issue?



image

tomvanmele:
the repo and CI infrastructure will be updated, but not in the next few days. perhaps in the meantime, you could just install and run black/ruff locally?

jf---:
@tomvanmele sure thing -- and sorry for being a little non-pragmatic here.
There's a nice plugin for pycharm that formats with black when saving a file, pretty sweet.
An apology is in order -- my comments above were pretty whiny.
Seems I'm PTSS'y not seeing green flags 👺🙈

jf---:
hi @tomvanmele || @gonzalocasas, curious if your happy with the implementation?

src/compas/geometry/frame.py
```diff
@@ -24,6 +24,10 @@
 from .point import Point
 from .quaternion import Quaternion
 
+from scipy.spatial.transform import Rotation as R
+from scipy.spatial.transform import Slerp
+from ..itertools import linspace
```
tomvanmele: usually, we don't use relative imports from packages/modules that cant be "seen" by the current packages/module. would prefer `compas.itertools` instead...

src/compas/geometry/frame.py
```diff
@@ -592,6 +596,73 @@ def to_transformation(self):
     # Methods
     # ==========================================================================
 
+    def interpolate_frame(self, other, t):
+        """Interpolates between two frames at a given parameter t in the range [0, 1]
+
+        Parameters
+        ----------
+        other : :class:`compas.geometry.Frame`
+        t: float
```
tomvanmele: missing space after `t`

src/compas/geometry/frame.py
```diff
@@ -592,6 +596,73 @@ def to_transformation(self):
...
+        -------
+        :class:`compas.geometry.Frame`
+            the interpolated frame
+
+        Examples
+        --------
+        >>> frame1 = Frame(Point(0, 0, 0), Vector(1, 0, 0), Vector(0, 1, 0))
+        >>> frame2 = Frame(Point(1, 1, 1), Vector(0, 0, 1), Vector(0, 1, 0))
+        >>> start_frame = frame1.interpolate_frame(frame2, 0)
+        >>>> allclose(start_frame.point, frame1.point) and allclose(start_frame.quaternion, frame1.quaternion)
```
tomvanmele: `>>>>` is one too many

src/compas/geometry/frame.py
```diff
@@ -592,6 +596,73 @@ def to_transformation(self):
     # Methods
     # ==========================================================================
 
+    def interpolate_frame(self, other, t):
+        """Interpolates between two frames at a given parameter t in the range [0, 1]
```
tomvanmele, line 598: period missing at the end

src/compas/geometry/frame.py
```diff
@@ -592,6 +596,73 @@ def to_transformation(self):
...
+        Parameters
+        ----------
+        other : :class:`compas.geometry.Frame`
+        t: float
+            a parameter in the range [0-1]
+
+        Returns
+        -------
+        :class:`compas.geometry.Frame`
+            the interpolated frame
```
tomvanmele: my personal preference is that these things are sentences...

tomvanmele: meaning starting with a capital and ending a period

src/compas/geometry/frame.py
```diff
@@ -24,6 +24,10 @@
 from .point import Point
 from .quaternion import Quaternion
 
+from scipy.spatial.transform import Rotation as R
+from scipy.spatial.transform import Slerp
```
tomvanmele: importing stuff from `scipy` like this will make the code unusable in IronPython. unfortunately, we are not dropping support for IronPython in COMPAS 2...

src/compas/geometry/frame.py
```diff
@@ -592,6 +596,73 @@ def to_transformation(self):
...
+        interpolated_frame = Frame.from_quaternion(rot, origin_interpolated)
+
+        return interpolated_frame
+
+    def interpolate_frames(self, other, steps):
+        """Generates a specified number of interpolated frames between two given frames
+
+        Parameters
+        ----------
+        other: Frame
```
tomvanmele:
```
other : :class:`compas.geometry.Frame`
```

src/compas/geometry/frame.py
```diff
@@ -592,6 +596,73 @@ def to_transformation(self):
...
+        steps: int
```
tomvanmele: `steps : int` (not sure this gets rendered properly without the space)

tests/compas/geometry/test_frame.py
```diff
@@ -68,3 +68,20 @@ def test_frame_predefined():
     assert frame.point == Point(0, 0, 0)
     assert frame.xaxis == Vector(0, 0, 1)
     assert frame.yaxis == Vector(1, 0, 0)
+
```
tomvanmele, line 75: perhaps some other "well-known" interpolation values could be added? e.g. `t=0.25`, `t=0.5`, `t=0.75`

yck011522:
Thanks for all the love.

gonzalocasas:
Not long ago I wrote a frame interpolator using nurbs curves interpolation instead, would it make sense to combine these things? Is there any benefits of using slerp over (perhaps a more mundane) nurbs curve interpolation?

yck011522:
@gonzalocasas 
Frame interpolation with slerp produce a more natural looking interpolation. It also don't have gimbal lock problems, don't have numerical instability problems. I don't know enough to explain it in mathematical terms, but it is just better. 

I think other interpolation methods have their place too. Most animation software will provide a number of choices for the user. 

Some further reading:
https://www.occasoftware.com/blog/unity-slerp-a-comprehensive-guide

yck011522:
Some visual comparison here between Lerp (Linear Interpolation)  and Slerp (Spherical Linear Interpolation): https://youtu.be/IymDGkzwJts?si=W0VRcv38-7XRzywz&t=367

tomvanmele:
perhaps we can provide "all" options through a `method` parameter with `"slerp"` as default? in any case we will have to move the `scipy` imports to the method body, and in case of multiple implementations, perhaps make multiple functions and make the method of the frame a "switch" based on the "method" parameter...

jf---:
hi @gonzalocasas, @yck011522 is right, for interpolation of frames, its hard to get this numerically stable. a `nurbs` interpolation would make a useful contribution, but is something completely different altogether. Thanks for the crystal clear reference to the YouTube video!

When writing tool paths, to robot code, it's a good idea to write a frame every couple of centimeter, and especially for large rotations, it's good practice to interpolate these. In practice that leads to more robust tooling paths.

> perhaps we can provide "all" options 

no, I don't think that's a good idea: slerp is perfectly reasonable for the interpolation of frames.
you could get numerically wildly differing results, due to an optional flag.
moreover, interpolation of frames has a clear and specific significance, and slerp is the accepted method to achieve this.
I don't think the idea of a various implementations of a fundamental algorithm is welcome: different code path lead to different results potentially.

@gonzalocasas it would be great to have an interpolation method on the `Curve` class 👏🏻

let's Keep It Simple Somehow ;)

tomvanmele:
@jf--- `compas.geometry.NurbsCurve.from_interpolation` creates a curve by interpolating given point locations. to get multiple points on the curve by interpolating the curve domain, you can do

```python
points = [curve.point_at(t) for t in linspace(*curve.domain, 100)]
```

src/compas/geometry/frame.py
jf---: ah in discussing the issue #1208  you stated that:

> in general, if we can use something that already exists, we should. and since we stopped caring about IronPython, there is no need to make a pure Python version...

but I see the problem, and let's settle on a pure python version over conditional import of dependency ( eg switching between implementations depending on what python interpreter )

tomvanmele: as you want, but there are many algorithms in `compas` core that use `numpy` and `scipy`. you don't have to make the import conditional. just move it in to the function body. in most cases, the functionality will be available since many people will be using `compas` with CPython now...

tomvanmele: you could also make the function a pluggable and add the `scipy` implementation as a plugin. then there is no conditionality and there is no need for the import being moved to the function/method body...

tomvanmele: only thing we can't do is break compatibility with IronPython in the core of COMPAS 2

jf---: > you don't have to make the import conditional. just move it in to the function body

Shoot, I spend this morning reimplementing this in pure python 🐍
OK, it could be cool if that convention is made more explicit
Not sure if I like it: it creates an easy to miss discrepancy between what is supported across interpreters.

tomvanmele: again, as you prefer, but it is an unfortunate reality that not everything is available when working in IronPython. things like PCA, ICP, Oriented Bounding Box, ... are also only available if `numpy`/`scipy` are available.

jf---: Understood. I'm happy as things are: this is well tested and ready to be merged, if you are content

tomvanmele:
@jf---  just for the record, the COMPAS 2 way of dealing with multiple implementations that have different dependencies that are not available on all interpreters, would be to add different implementations in separate modules as plugins for a pluggable entry point. the method of the frame would then call the pluggable and therefore automatically switch between implementations based on the availability of requirements.

something like the following:

```python
# compas.geometry.interpolation_frame.py

@pluggable(category='geometry')
def interpolate_frame(...):
    <add pure python implementation here if it exists>

```

```python
# compas.geometry.interpolation_frame_numpy.py

@plugin(category="geometry", requires=['scipy']):
    <add scipy implementation here>

```

register the plugin in `compas.__init__.py`

```python
# compas.__init__.py

__all_plugins__ = [
    ...
    'compas.geometry.interpolation_frame_numpy'
]

```

import and use the pluggable in `compas.geometry.Frame`...

src/compas/geometry/frame.py
tomvanmele: well with the imports at the top of the file, the entire `Frame` class is no longer available in IronPython. this should not be the case. so unfortunately, you will have to move the imports to the function body, or use a pur Python implementation, or use the a pluggable/plugin...

jf---: thanks for the pluggable example!

tomvanmele: currently the workflows seem to fail because you still have the `scipy` imports in `Frame.py` even though they are no longer being used...

src/compas/geometry/quaternion.py
```diff
@@ -471,3 +474,72 @@ def conjugated(self):
...
+        >>> interpolated_quaternion = Quaternion.slerp(q1, q2, t)
+
+
+        """
+        epsilon = float_info.epsilon
+
+        q1 = self.unitized()
+        q2 = other.unitized()
+
+        cosom = self.dot(q2)
```
gonzalocasas: Shouldn't it be `q1.dot(q2)` here?

src/compas/geometry/quaternion.py
```diff
@@ -471,3 +474,72 @@ def conjugated(self):
...
+        return self.w * other.w + self.x * other.x + self.y * other.y + self.z * other.z
+
+    def slerp(self, other, t):
+        """Slerp: spherical interpolation of two quaternions.
+
+        Parameters
+        ----------
+        other : :class:`compas.geometry.Quaternion`
+            The other quaternion to interpolate between.
+
```
gonzalocasas:
```suggestion
```

I'm not sure, but since rst is super picky, I would remove this new-line.

src/compas/geometry/quaternion.py
```diff
@@ -471,3 +474,72 @@ def conjugated(self):
...
+        else:
+            sc1 = 1.0 - t
+            sc2 = t
+
+        qW_interp = sc1 * quat[0] + sc2 * q2[0]
+        qX_interp = sc1 * quat[1] + sc2 * q2[1]
+        qY_interp = sc1 * quat[2] + sc2 * q2[2]
+        qZ_interp = sc1 * quat[3] + sc2 * q2[3]
+
+        return Quaternion(qW_interp, qX_interp, qY_interp, qZ_interp)
```
gonzalocasas: Nitpicking, but I'd use lowercase here:
```suggestion
        qw_interp = sc1 * quat[0] + sc2 * q2[0]
        qx_interp = sc1 * quat[1] + sc2 * q2[1]
        qy_interp = sc1 * quat[2] + sc2 * q2[2]
        qz_interp = sc1 * quat[3] + sc2 * q2[3]

        return Quaternion(qw_interp, qx_interp, qy_interp, qz_interp)
```

tests/compas/geometry/test_frame.py
```diff
@@ -1,14 +1,17 @@
 from __future__ import division
+
+import math
+
 import pytest
 import json
 import compas
 from random import random
-from compas.geometry import allclose
+from compas.geometry import allclose, Rotation
```
gonzalocasas: We have this `isort` setting that forces all imports to be single-line. The setting is at the top of the repo, so, if you have installed isort, it should auto-sort them. In VSCode, you can do `right-click -> Organizer imports`:

```suggestion
from compas.geometry import allclose
from compas.geometry import Rotation
```

src/compas/geometry/frame.py
```diff
@@ -24,6 +24,9 @@
 from .point import Point
 from .quaternion import Quaternion
 
+from scipy.spatial.transform import Rotation as R
+from scipy.spatial.transform import Slerp
```
gonzalocasas: Since the implementation is now pure-python, please remove these ones, because they are causing the build to break.
```suggestion
```

requirements.txt
```diff
@@ -2,6 +2,6 @@
 jsonschema
 networkx >= 3.0
 numpy >= 1.15.4
-scipy >= 1.1
+scipy >= 1.2
```
gonzalocasas: If we don't need this anymore, I would revert to 1.1.

gonzalocasas (commented):
looking good! Added a few comments here and ther

src/compas/geometry/frame.py
jf---: shoot I've been force pushing for too long, some messy things like the above were fixed, but I messed up rebasing the branch

src/compas/geometry/quaternion.py
jf---: nice catch @gonzalocasas  💎

requirements.txt
jf---: 🙏🏼

tests/compas/geometry/test_frame.py
jf---: @gonzalocasas thx for the VSCode heads up, alas, no pycharm equivalent.
what about embracing `pre-commit` hooks?
[seems a super clean way of enforcing code standards?](https://gist.github.com/guptarohit/35f49bc663620d03ca00d2a0ff7429df)

jf---:
I think we're there -- many thanks for the code review @gonzalocasas & @tomvanmele, that's super fun & welcoming

tests/compas/geometry/test_frame.py
gonzalocasas: Totally in for `pre-commit` hooks, I just didn't have the time to set that up yet

gonzalocasas:
We're *almost* there! The linter is still complaining about tiny little things:
```
src/compas/geometry/frame.py:29:1: E302 expected 2 blank lines, found 1
src/compas/geometry/quaternion.py:545:1: W391 blank line at end of file
tests/compas/geometry/test_frame.py:10:1: F401 'compas.geometry.Rotation' imported but unused
tests/compas/geometry/test_frame.py:17:1: E302 expected 2 blank lines, found 1
```

jf---:
@gonzalocasas thx for the help, nice to see lots of green flags, weirdly encouraging

gonzalocasas:
Yay!! 🚀 Awesome!

gonzalocasas (approved):

yck011522:
Thanks again for all the love. This is a very important function for robotics application. Maybe for people who do animations too.

PR #1325, opened 2024-04-19
BUG FIX is_coplanar

petrasvestartas:
Little culprit, when there are 4 points, the while loop fails while popping points in predicates_3.py -> is_coplanar. The previous if statement in that method is not enough, which I guess was an initial assumption.

### Checklist


tomvanmele (approved):

tomvanmele:
would be nice to add a test for the case that previously failed :)

petrasvestartas:
@tomvanmele just added.

This method was used in compas_viewer hide_coplanar_edges. When you input meshes with extremely sharp faces, due to floating point errors they become almost co-linear. I added this IndexError test that pass through and should not cause issues. 

Now it works, 3rd line from this screenshot:

Screenshot from 2024-04-20 12-26-19

PR #1326, opened 2024-04-22
Fix unify cycles

tomvanmele:

### Checklist

### DIFF
--- a/src/compas/datastructures/mesh/mesh.py
+++ b/src/compas/datastructures/mesh/mesh.py
@@ -58,6 +58,7 @@ from compas.utilities import window
 from compas.topology import breadth_first_traverse
 from compas.topology import face_adjacency
 from compas.topology import connected_components
+from compas.topology import unify_cycles
 
 from compas.datastructures.datastructure import Datastructure
 from compas.datastructures.attributes import VertexAttributeView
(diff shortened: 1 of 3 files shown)


src/compas/datastructures/mesh/mesh.py
```diff
@@ -2953,42 +2953,22 @@ def unify_cycles(self, root=None):
...
-                        # is in the same direction
-                        # flip the neighbor
-                        self.face[nbr][:] = self.face[nbr][::-1]
-                        return
-
-        if root is None:
-            root = self.face_sample(size=1)[0]
-
+        vertex_index = {vertex: index for index, vertex in enumerate(self.vertices())}
+        index_vertex = {index: vertex for index, vertex in enumerate(self.vertices())}
```
gonzalocasas: Besides the nice terseness of this, wouldn't it be more efficient (and marginally more readable also) to do a single `for` loop enumerating over the vertices, and create the two indices in one go?  Something like

```suggestion
        vertex_index = {}
        index_vertex = {}
        for index, vertex in enumerate(self.vertices()):
                vertex_index[vertex] = index
                index_vertex[index] = vertex
```

src/compas/topology/orientation.py
```diff
@@ -88,7 +88,7 @@ def _face_adjacency(vertices, faces, nmax=10, radius=10.0):
                         found.add(nbr)
                         break
 
-        adjacency[face] = nbrs
+        adjacency[index] = nbrs
```
gonzalocasas, line 91: I think this is introducing a bug or at least, a breaking change. 

The public function `face_adjacency` (right below on this file) returns the result of `_face_adjacency` for `len(faces) > 100` but that function is documented as returning face identifiers, not indices. On that note, the public function is incorrectly documented as taking a `Mesh` as param, when it's actually taking `points, faces` lists.

gonzalocasas (changes requested):

src/compas/topology/orientation.py
tomvanmele: no it is correct like this. previous version was wrong. not sure how that happened. `face` is a list of vertices. it is not an identifier. `index` is the index of the list of vertices defining the face in the list of faces. so the adjacency dict is in fact returning identifiers as keys as you say...

tomvanmele: will correct the docs.

src/compas/datastructures/mesh/mesh.py
tomvanmele: happy to make the change but not sure if it will be faster. comprehensions are highly optimised in recent python versions...

src/compas/topology/orientation.py
tomvanmele: > no it is correct like this. previous version was wrong. not sure how that happened. `face` is a list of vertices. it is not an identifier. `index` is the index of the list of vertices defining the face in the list of faces. so the adjacency dict is in fact returning identifiers as keys as you say...

now it is the same as `face_adjacency` a few lines below, where it is perhaps a bit clearer because the enumeration says `for i, vertices in enumerate(faces)` instead of `for index, face in enumerate(faces)`

src/compas/datastructures/mesh/mesh.py
gonzalocasas: ah ok, in that case, no idea what's faster

gonzalocasas (approved):

PR #1327, opened 2024-04-22
Minor additions

tomvanmele:

### Checklist


src/compas/geometry/brep/brep.py
```diff
@@ -489,6 +487,26 @@ def from_pipe(cls, curve, radius, thickness=None):
         """
         return from_pipe(curve, radius, thickness=thickness)
 
+    @classmethod
+    def from_plane(cls, plane, domain_u=(-1, +1), domain_v=(-1, +1)):
+        """Construct a Brep from a set of planes.
```
gonzalocasas: From one plane of from a set of planes? If the latter, the method should be called `from_planes`

src/compas/geometry/brep/__init__.py
```diff
@@ -72,6 +72,11 @@ def from_pipe(*args, **kwargs):
     raise PluginNotInstalledError
 
 
+@pluggable(category="factories")
+def from_plane(*args, **kwargs):
+    raise PluginNotInstalledError
+
```
gonzalocasas, line 78: Is there a rhino implementation that we could add right away as well? Or will this be only OCC?

gonzalocasas (approved):
lgtm + a couple of questions

src/compas/geometry/brep/__init__.py
tomvanmele: dunno. will check

tomvanmele: no, Rhino doesn't have a `CreateFromPlane` or similar...

PR #1332, opened 2024-04-23
Update format settings

tomvanmele:

### Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -118,20 +118,15 @@ replace = "[{new_version}] {now:%Y-%m-%d}"
 # ============================================================================
 
 [tool.black]
-# we should set this to 179
-line-length = 120
+line-length = 179
 
 [tool.ruff]
-# we should set this to 179
-line-length = 120
+line-length = 179
 indent-width = 4
 target-version = "py39"
 
 [tool.ruff.lint]
-# we should add "I"
-select = ["E", "F"]
-# we should remove this
-ignore = ["E501"]
+select = ["E", "F", "I"]
 
 [tool.ruff.lint.per-file-ignores]
 "__init__.py" = ["I001"]
@@ -151,8 +146,7 @@ known-first-party = [
 convention = "numpy"
 
 [tool.ruff.lint.pycodestyle]
-# we should set this to 179
-max-doc-length = 120
+max-doc-length = 179
 
 [tool.ruff.format]
 docstring-code-format = true
--- a/.editorconfig
+++ b/.editorconfig
@@ -8,7 +8,7 @@ insert_final_newline = true
 indent_style = space
 indent_size = 4
 charset = utf-8
-max_line_length = 120
+max_line_length = 179
 
 [*.{bat,cmd,ps1}]
 end_of_line = crlf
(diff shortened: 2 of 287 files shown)


tomvanmele:
check the first commit for the settings

tomvanmele:
```toml
[tool.black]
line-length = 179

[tool.ruff]
line-length = 179
indent-width = 4
target-version = "py39"

[tool.ruff.lint]
select = ["E", "F", "I"]

[tool.ruff.lint.pycodestyle]
max-doc-length = 179
```

pyproject.toml
```diff
@@ -118,20 +118,15 @@ replace = "[{new_version}] {now:%Y-%m-%d}"
 # ============================================================================
 
 [tool.black]
-# we should set this to 179
-line-length = 120
+line-length = 179
```
gonzalocasas, line 121: There's a paragraph in the devguide that needs to be updated to reflect this change as well. And we need to create follow up issues to update this across the ecosystem (compas theme, and other core packages)

pyproject.toml
```diff
@@ -118,20 +118,15 @@ replace = "[{new_version}] {now:%Y-%m-%d}"
...
 indent-width = 4
 target-version = "py39"
 
 [tool.ruff.lint]
-# we should add "I"
-select = ["E", "F"]
-# we should remove this
-ignore = ["E501"]
+select = ["E", "F", "I"]
```
gonzalocasas, line 129: Without having to read the ruff docs, does this imply we enable all? What is excluded? Maybe drop a line indicating why/what is done here

gonzalocasas (approved):
Besides a tiny comment, it lgtm!

I do wonder though, why raising the max line length? Any concrete reason?

pyproject.toml
gonzalocasas: In that line, I think the devguide needs to be update to indicate the move to `ruff`, right?

pyproject.toml
tomvanmele: there is also "B", "Q", ... which are things we don't need/want

pyproject.toml
tomvanmele: > There's a paragraph in the devguide that needs to be updated to reflect this change as well. And we need to create follow up issues to update this across the ecosystem (compas theme, and other core packages)

am already doing this for `compas_cgal`, `compas_occ`, `compas_libigl`, `compas_gmsh`, `compas_viewer`, `compas_notebook`, `compas_model`

tomvanmele: re the devguide: yes indeed, we should. will make the change...

tomvanmele:
> Besides a tiny comment, it lgtm!
> 
> I do wonder though, why raising the max line length? Any concrete reason?

i just very often bump into the fact that 120 is too short. many comprehensions, or nested function calls get formatted really awkwardly because of it. also in docstrings it happens a lot (ruff doesn't properly differentiate those afaik). if you don't like it we can change it back :)

gonzalocasas:
> > Besides a tiny comment, it lgtm!
> > I do wonder though, why raising the max line length? Any concrete reason?
> 
> i just very often bump into the fact that 120 is too short. many comprehensions, or nested function calls get formatted really awkwardly because of it. also in docstrings it happens a lot (ruff doesn't properly differentiate those afaik). if you don't like it we can change it back :)

nah, it's fine, I wasn't sure if there was an underlying technical reason for it (like ruff defaulting to it, or something like that)

PR #1340, opened 2024-04-24
Fix the doctests

tomvanmele:

### Checklist

### DIFF
--- a/src/compas/data/encoders.py
+++ b/src/compas/data/encoders.py
@@ -181,13 +181,14 @@ class DataDecoder(json.JSONDecoder):
 
     >>> import json
     >>> from compas.data import DataDecoder
-    >>> with open("point.json", "r") as f:
-    ...     point = json.load(f, cls=DataDecoder)
+    >>> with open('point.json', 'r') as f:        # doctest: +SKIP
+    ...     point = json.load(f, cls=DataDecoder) # doctest: +SKIP
+    ...
 
     Implicit use case.
 
     >>> from compas.data import json_load
-    >>> point = json_load("point.json")
+    >>> point = json_load('point.json')  # doctest: +SKIP
 
     """
 
--- a/src/compas/geometry/curves/hyperbola.py
+++ b/src/compas/geometry/curves/hyperbola.py
@@ -99,19 +99,19 @@ class Hyperbola(Conic):
 
     Construct a hyperbola such that the Z axis of its frame aligns with a given line.
 
-    >>> from compas.geometry import Line, Frame, Hyperbola
+    >>> from compas.geometry import Line, Plane, Frame, Hyperbola
     >>> line = Line([0, 0, 0], [1, 1, 1])
     >>> plane = Plane(line.end, line.direction)
     >>> hyperbola = Hyperbola(major=3, minor=2, frame=Frame.from_plane(plane))
 
     Visualise the line, hyperbola, and frame of the hyperbola with the COMPAS viewer.
 
-    >>> from compas_viewer import Viewer  # doctest: +SKIP
-    >>> viewer = Viewer()                    # doctest: +SKIP
-    >>> viewer.scene.add(line)                  # doctest: +SKIP
+    >>> from compas_viewer import Viewer          # doctest: +SKIP
+    >>> viewer = Viewer()                         # doctest: +SKIP
+    >>> viewer.scene.add(line)                    # doctest: +SKIP
     >>> viewer.scene.add(hyperbola)               # doctest: +SKIP
     >>> viewer.scene.add(hyperbola.frame)         # doctest: +SKIP
-    >>> viewer.show()                      # doctest: +SKIP
+    >>> viewer.show()                             # doctest: +SKIP
 
     """
 
--- a/src/compas/geometry/intersection.py
+++ b/src/compas/geometry/intersection.py
@@ -18,15 +18,15 @@ class Intersection(object):
 
     Examples
     --------
-    >>> from compas.geometry import Line
-    >>> from compas.geometry import Intersection
-    >>> a = Line([0, 0, 0], [2, 0, 0])
-    >>> b = Line([1, 0, 0], [1, 1, 0])
-    >>> intersection = Intersection()
-    >>> intersection.line_line(a, b)
-    >>> intersection.number_of_intersections
+    >>> from compas.geometry import Line          # doctest: +SKIP
+    >>> from compas.geometry import Intersection  # doctest: +SKIP
+    >>> a = Line([0, 0, 0], [2, 0, 0])            # doctest: +SKIP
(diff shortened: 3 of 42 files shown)


tomvanmele:
* not all objects have `__str__`
* having to use `print(...)` in docstring examples to make sure they don't fail due to precision errors is annoying
* would be good to have a template structure of some sorts, or at least a few "rules", for docstring examples

other things i have noticed
* shapes have u and v even when only u makes sense. perhaps that should be changed
* many attributes could be cached. would be good to have a system for that so that the cache gets invalidated when the parameters of an object are changed (could be a decorator for some attributes linked to an invalidate_cache method)
* would be useful to further extend the basic infrastructure of with a bounding box attribute
* point (and vector), line, polyline, polygon, polyhedron share the fact that they are defined explicitly by points. they therefore also don't have a frame. would be good to document this as such.
* the parametric objects do have a frame, would be good to point this out as well in the docs
* some intersections are still missing
* an intersection class that combines all intersection calculations would be useful.
* in general there should be a rule for binary operations between geometric entities about which entity is "left" and which is "right". for example the lower-dimensional object could always be on the left...

tomvanmele:
@jf--- since i can't add you as reviewer... ping

tomvanmele:
my mouse and keyboard don't want to connect anymore so i will take that as a sign that it is time to give up for today...

tomvanmele:
all are fixed.
some had to be skipped because of some weird behaviour, others because of an actual bug. we can do a search for `# doctes: +SKIP` to investigate in more detail later...

tomvanmele:
the tests are failing because of something weird going on on the github side

src/compas/data/encoders.py
```diff
@@ -182,14 +182,14 @@ class DataDecoder(json.JSONDecoder):
 
     >>> import json
     >>> from compas.data import DataDecoder
-    >>> with open('point.json', 'r') as f:
-    ...     point = json.load(f, cls=DataDecoder)
+    >>> with open('point.json', 'r') as f:        # doctest: +SKIP
+    ...     point = json.load(f, cls=DataDecoder) # doctest: +SKIP
```
Licini, line 185: just out of curiosity why these needs to be skipped? So in doctest the read and writing to disk is not possible? If that's the case perhaps these kind of stuff can be `loads` and `dumps` instead

src/compas/geometry/curves/hyperbola.py
```diff
@@ -99,19 +99,19 @@ class Hyperbola(Conic):
...
-    >>> from compas_viewer import Viewer  # doctest: +SKIP
-    >>> viewer = Viewer()                    # doctest: +SKIP
-    >>> viewer.scene.add(line)                  # doctest: +SKIP
+    >>> from compas_viewer import Viewer          # doctest: +SKIP
+    >>> viewer = Viewer()                         # doctest: +SKIP
+    >>> viewer.scene.add(line)                    # doctest: +SKIP
     >>> viewer.scene.add(hyperbola)               # doctest: +SKIP
     >>> viewer.scene.add(hyperbola.frame)         # doctest: +SKIP
-    >>> viewer.show()                      # doctest: +SKIP
+    >>> viewer.show()                             # doctest: +SKIP
```
Licini, line 114: Maybe in future we should remove them from API reference of core? (After we clean up and improve the documents of viewer)

src/compas/geometry/intersection.py
```diff
@@ -18,15 +18,15 @@ class Intersection(object):
...
-    >>> intersection = Intersection()
-    >>> intersection.line_line(a, b)
-    >>> intersection.number_of_intersections
+    >>> from compas.geometry import Line          # doctest: +SKIP
+    >>> from compas.geometry import Intersection  # doctest: +SKIP
+    >>> a = Line([0, 0, 0], [2, 0, 0])            # doctest: +SKIP
+    >>> b = Line([1, 0, 0], [1, 1, 0])            # doctest: +SKIP
+    >>> intersection = Intersection()             # doctest: +SKIP
+    >>> intersection.line_line(a, b)              # doctest: +SKIP
+    >>> intersection.number_of_intersections      # doctest: +SKIP
```
Licini, line 27: plugin required for these ones?

src/compas/geometry/offset.py
```diff
@@ -138,7 +138,7 @@ def offset_polygon(polygon, distance, tol=None):
     >>> polygon = Polygon([(0.0, 0.0), (1.0, 0.0), (1.0, 1.0), (0.0, 1.0)])
     >>> offsetted_polygon = offset_polygon(polygon, 0.5)
     >>> offsetted_polygon
-    Polygon[[0.5, 0.5, 0.0], [0.5, 0.5, 0.0], [0.5, 0.5, 0.0], [0.5, 0.5, 0.0]]
+    [[0.5, 0.5, 0.0], [0.5, 0.5, 0.0], [0.5, 0.5, 0.0], [0.5, 0.5, 0.0]]
```
Licini, line 141: shouldn't this function return a Polygon object instead of list to align with the input?

src/compas/geometry/polyhedron.py
```diff
@@ -491,7 +491,7 @@ def from_convex_hull(cls, points):
         --------
         >>> from compas.geometry import Polyhedron
         >>> points = [[0, 0, 0], [1, 0, 0], [0, 1, 0]]
-        >>> p = Polyhedron.from_convex_hull(points)
+        >>> p = Polyhedron.from_convex_hull(points)  # doctest: +SKIP
```
Licini, line 494: It would be nice if we also have a way document the minimal plugins needed for certain functions

Licini (approved):
LGTM! Just some random comments!

src/compas/geometry/curves/hyperbola.py
tomvanmele: yes i agree. viewer stuff should not be in the docs. will already remove them...

src/compas/data/encoders.py
tomvanmele: perhaps this works now actually. will try again...

src/compas/geometry/intersection.py
tomvanmele: no, this was more like an API test example. the class is not yet part of the public API...

src/compas/geometry/offset.py
tomvanmele: no, the lower level function shouldn't. the version attached to the polygon class itself should...

jf---:
@tomvanmele this is exciting, is this PR tangential to #1340?

tomvanmele:
which one do you mean? this one is #1340...

src/compas/geometry/offset.py
Licini: i see

gonzalocasas:
@tomvanmele there's quite a bunch of conflicts now, but it would be cool to merge this

Issue #1343, opened 2024-04-28
Cannot convert from Rhino.Point to compas point using `point_to_compas`


xarthurx:
**Describe the bug**

`point_to_compas` can convert `Point3d` to compas-point object.
However, when the point is converted from `rs.coercerhinoobject(pt).Geometry` inside a GHpython script, the default result is `Rhino.Geometry.Point` instead of `Rhino.Geometry.Point3d`

using the current [code](https://github.com/compas-dev/compas/blob/c2516e9919fa1457aa9f16433b2d7586d6035a8f/src/compas_rhino/conversions/geometry.py#L101) of `point_to_compas`, this will result in a error: `AttributeError: 'Point' object has no attribute 'X'`

Based on [here](https://discourse.mcneel.com/t/could-not-convert-rhino-geometry-point-object-to-a-point3d/133571/2), in such case, we need to use `pt.Location` to get the coordinate, which is also a valid input for `point_to_compas`.

**To Reproduce**
1. In GHPython, make a python script (IronPython): 
![image](https://github.com/compas-dev/compas/assets/1921878/bf652452-a122-48a9-ac21-c1c8a57cfa0c)
2. Inside:
 
```python
# if Guid, convert to Rhino Object
import compas_rhino.conversions as cnv 
import Rhino.Geometry as rg
import rhinoscriptsyntax as rs
import System.Guid

if type(obj) == System.Guid:
    obj = rs.coercerhinoobject(obj).Geometry

obj_type = type(obj)
if obj_type == rg.Point:
    print("converting to point")
    res1 = cnv.point_to_compas(obj)
    res2 = cnv.point_to_compas(obj.Location)
```

**Expected behavior**
`res1` should work. But only `res` is working.

**Screenshots**
![image](https://github.com/compas-dev/compas/assets/1921878/a5c8cdb2-eb35-4db4-b4a0-723d47ca32f7)

**Desktop (please complete the following information):**
 - OS: Windows
 - Python version: IronPython

**Additional context**
Should be a simple fix. So I didn't bother to create a PR. If you need, a PR is also possible.

Licini:
@xarthurx Fixed now, let me know if there is still any issue with this.

PR #1346, opened 2024-04-29
Update/rhino point conversion

Licini:
Small fix for #1343

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,7 +34,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Changed use of `compas.geometry.allclose` to `compas.tolerance.TOL.is_allclose`.
 * Changed use of `compas.geometry.close` to `compas.tolerance.TOL.is_close`.
 * Changed imports of itertools to `compas.itertools` instead of `compas.utilities`.
-* Updated `compas_rhino.conversions.point_to_compas` to allow for `compas.geometry.Point` as input.
+* Updated `compas_rhino.conversions.point_to_compas` to allow for `Rhino.Geometry.Point` as input.
 
 ### Removed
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -34,6 +34,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Changed use of `compas.geometry.allclose` to `compas.tolerance.TOL.is_allclose`.
 * Changed use of `compas.geometry.close` to `compas.tolerance.TOL.is_close`.
 * Changed imports of itertools to `compas.itertools` instead of `compas.utilities`.
+* Updated `compas_rhino.conversions.point_to_compas` to allow for `compas.geometry.Point` as input.
```
gonzalocasas: Do you mean ```to allow for `Rhino.Geometry.Point` as input.```?

src/compas_rhino/conversions/geometry.py
```diff
@@ -103,14 +103,19 @@ def point_to_compas(point):
...
     :class:`compas.geometry.Point`
 
     """
-    return Point(point.X, point.Y, point.Z)
+    if isinstance(point, Rhino.Geometry.Point3d):
+        return Point(point.X, point.Y, point.Z)
+    elif isinstance(point, Rhino.Geometry.Point):
+        return Point(point.Location.X, point.Location.Y, point.Location.Z)
+    else:
+        raise TypeError("Expected Rhino.Geometry.Point3d or Rhino.Geometry.Point., got: {}".format(type(point)))
```
gonzalocasas: Instead of relying on type checking, I would rely on "feature checking". Something like this maybe?
```suggestion
    if hasattr(point, "X"):
        return Point(point.X, point.Y, point.Z)
    elif hasattr(point, "Location"):
        return Point(point.Location.X, point.Location.Y, point.Location.Z)
    else:
        raise TypeError("Unexpected point type, got: {}".format(type(point)))
```

Or maybe not even that, maybe just try..except idiom:
```python
    try:
        return Point(point.X, point.Y, point.Z)
    except MissingMemberException:
        try:
            return Point(point.Location.X, point.Location.Y, point.Location.Z)
        except MissingMemberException:
            raise TypeError("Unexpected point type, got: {}".format(type(point)))
```

gonzalocasas (changes requested):

CHANGELOG.md
Licini: My bad!

Licini: Updated!

src/compas_rhino/conversions/geometry.py
Licini: Sounds good! using try except now

gonzalocasas (approved):
Thanks for the updates! 
LGTM!

Issue #1344, opened 2024-04-29
Point float arithmetic not mutable?!


xarthurx:
**Describe the bug**
The code means more than a thousand words:

```python
>>> p1 = cg.Point(1.2, 4, 3)
>>> p1                           
Point(x=1.2, y=4.0, z=3.0)
>>> p1 * 0.5 
Point(x=0.6, y=2.0, z=1.5)
>>> 0.5 * p1
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: unsupported operand type(s) for *: 'float' and 'Point'
```

**Expected behavior**
No Errors

**Desktop (please complete the following information):**
 - OS: Win/Linux
 - Python version [e.g. 2.7]
 - Python package manager [e.g. macports, pip, conda]

**Additional context**
Is this by design?!

gonzalocasas:
This was discussed recently here https://github.com/compas-dev/compas/issues/1248#issuecomment-1877582973

xarthurx:
> This was discussed recently here [#1248 (comment)](https://github.com/compas-dev/compas/issues/1248#issuecomment-1877582973)

OK. as one basic math operations, I DO think this is very necessary and common... 
If you don't support sth like this, I would recommend it should be highlighted in the documentation...

Based on my experience, people may prefer to put the float in front, especially when you compute a

**"WEIGHTED * long-expression"**

Issue #1248, opened 2024-01-04
Vector operators


ZacZhangzhuo:
**Describe the bug**

for Vector peration `*`, like:
```python
>>> u = Vector(1.000, 1.000, 0.000)
>>> u * 2
Vector(2.000, 2.000, 0.000)
```
if we do:
`Vector * float`: It is fine.
`float * Vector`: TypeError: unsupported operand type(s) for *: 'float' and 'Vector'.

**Expected behavior**
I think this could be fixed with minimal effort just for a better user experience. :)

gonzalocasas:
Can it really be solved easily? I have the feeling it's not possible to define the binding of the star operator to take right hand operand as precedence...

tomvanmele:
yes, we could implement the "r" versions of the magic methods, but i don't know if there would be any unwanted side-effects...

```python
def __rmul__(self, factor):
    ...
```

gonzalocasas:
Ah true 👍

ZacZhangzhuo:
yes, it's not a big issue. mathmatically the floating number is preferred in the front. like `a = k b`, a, b are vectors.

jf---:
@tomvanmele you're right that __rmul__ would address the issue, and @ZacZhangzhuo is right in that operators in Vectors should be complete -- but `float * Vector` is quite counter intuitive?

tomvanmele:
i agree, that is why never implemented the "r" versions. i think it is easy enough for the user to be explicit about their intent by writing the compas object first...

jf---:
@tomvanmele I know I'm being frantic, but seems like this issue can be closed? The broader concept of vec ops isn't discussed.

xarthurx:
> @tomvanmele you're right that **rmul** would address the issue, and @ZacZhangzhuo is right in that operators in Vectors should be complete -- but `float * Vector` is quite counter intuitive?

It is actually quite common and intuitive, especially when you compute a "WEIGHTED * long-expression"....

PR #1347, opened 2024-04-29
Vector multiplications

Licini:
Triggered by discussion of #1344 and #1248 

I'm more on the side of allowing `__rmul__`, since a lot of people have the habit of writing equations in such order. On implementation level it would be also tricky to show a warning to say that we don't allow it (in that case `__rmul__` is anyway needed). 

Also adding support for element-wise multiplication and division for vector * vector, aligning with other libraries like Numpy, Tensorflow, Pytorch, R ... (In MATLAB this returns dot product tho)

The PR is created more for discussion purpose, let me know what you guys think

### Checklist


gonzalocasas:
I like this, and the implementation looks good to me. I'm not sure how to assess correctly the potential side-effects of this though.

src/compas/geometry/vector.py
```diff
@@ -148,11 +148,21 @@ def __add__(self, other):
...
-    def __mul__(self, n):
-        return Vector(self.x * n, self.y * n, self.z * n)
-
-    def __truediv__(self, n):
-        return Vector(self.x / n, self.y / n, self.z / n)
+    def __mul__(self, other):
+        if isinstance(other, (int, float)):
+            return Vector(self.x * other, self.y * other, self.z * other)
+        elif isinstance(other, Vector):
+            return Vector(self.x * other[0], self.y * other[1], self.z * other[2])
```
tomvanmele: if `other` is `Vector`, why not `other.x` etc?
also, this is the same as `return self.dot(other)`

src/compas/geometry/vector.py
```diff
@@ -148,11 +148,21 @@ def __add__(self, other):
...
+        else:
+            raise TypeError("Multiplication of Vector with unsupported type: {}".format(type(other)))
+
+    def __truediv__(self, other):
+        if isinstance(other, (int, float)):
+            return Vector(self.x / other, self.y / other, self.z / other)
+        elif isinstance(other, Vector):
+            return Vector(self.x / other[0], self.y / other[1], self.z / other[2])
```
tomvanmele: same here, why not `self.x / other.x`?

src/compas/geometry/vector.py
```diff
@@ -190,6 +200,9 @@ def __ipow__(self, n):
         self.z **= n
         return self
 
+    def __rmul__(self, n):
```
tomvanmele, line 207: no `__rsub__`, `__radd__`? or do these not exist?

src/compas/geometry/vector.py
Licini: you are very right.... another mistake of mine not careful about what copilot suggested! But this is not same with dot product tho, this returns element-wise multiplication, the result is still of a vector, not the sum of them

src/compas/geometry/vector.py
Licini: they do, let me try this too...

Licini:
> I like this, and the implementation looks good to me. I'm not sure how to assess correctly the potential side-effects of this though.

I found this explanation for these functions: 
```

In Python, the method __add__() and __radd__() are both part of the data model used to define how objects can be added together. The priority between these two methods depends on the operands involved in the addition operation. Here's how it works:

__add__() Method:
This method is attempted first when you use the + operator. It is a method for the left-hand operand of the addition.
For instance, in the expression a + b, Python first tries to call a.__add__(b).
__radd__() Method:
This method is the "right-side" addition method. It is called if the left operand's __add__() method does not support the addition with the right operand (either it returns NotImplemented, or the left operand doesn't have an __add__() method at all).
For instance, if a.__add__(b) returns NotImplemented, Python will check if b has a __radd__() method and if so, it will call b.__radd__(a).
In summary, __add__() has priority over __radd__() as it is attempted first. If __add__() fails to return a valid result and does not raise an exception, then __radd__() is attempted.
```

I will expand the test a bit here, to see what happens when different types are multiplied/added together

Licini:
@tomvanmele @gonzalocasas What do you guys think about this way? If `other` is not a number, then we try casting the input into a Vector using `Vector(*other)` , this can allow more flexibility for other type of iterables

Licini:
As for possible side effects of these reverse operator functions, because we don't return `NotImplemented`, this means that we will force the program to use our vector operator functions as long as our `Vector` class shows up at either side of operator. Which is not a necessarily a bad thing no?

Licini:
@tomvanmele Could you take a look again and this one and see if it can be merged?

src/compas/geometry/vector.py
```diff
@@ -148,11 +148,25 @@ def __add__(self, other):
...
 
-    def __mul__(self, n):
-        return Vector(self.x * n, self.y * n, self.z * n)
-
-    def __truediv__(self, n):
-        return Vector(self.x / n, self.y / n, self.z / n)
+    def __mul__(self, other):
+        if isinstance(other, (int, float)):
+            return Vector(self.x * other, self.y * other, self.z * other)
+        else:
```
tomvanmele: else block is not needed because there is a return in the if

src/compas/geometry/vector.py
```diff
@@ -148,11 +148,25 @@ def __add__(self, other):
...
+            try:
+                other = Vector(*other)
+                return Vector(self.x * other.x, self.y * other.y, self.z * other.z)
+            except TypeError:
+                raise TypeError("Cannot cast {} {} to Vector".format(other, type(other)))
+
+    def __truediv__(self, other):
+        if isinstance(other, (int, float)):
+            return Vector(self.x / other, self.y / other, self.z / other)
+        else:
```
tomvanmele: same

src/compas/geometry/vector.py
```diff
@@ -148,11 +148,25 @@ def __add__(self, other):
     def __sub__(self, other):
         return Vector(self.x - other[0], self.y - other[1], self.z - other[2])
 
-    def __mul__(self, n):
-        return Vector(self.x * n, self.y * n, self.z * n)
-
-    def __truediv__(self, n):
-        return Vector(self.x / n, self.y / n, self.z / n)
+    def __mul__(self, other):
+        if isinstance(other, (int, float)):
```
tomvanmele, line 152: if we don't mind the performance penalty here i am okay with it...

tomvanmele (approved):

PR #1348, opened 2024-04-29
Make the tolerance class a singleton

tomvanmele:

### Checklist


src/compas/tolerance.py
```diff
@@ -151,22 +158,34 @@ def __init__(
...
+            self._relative = None
+            self._angular = None
+            self._approximation = None
+            self._precision = None
+            self._lineardeflection = None
+            self._angulardeflection = None
+
+        self._is_inited = True
+
+        if unit is not None:
```
Licini, line 180: not important.... but i guess unit will never be None because it has a default "M", no?

src/compas/tolerance.py
```diff
@@ -108,8 +114,9 @@ class Tolerance(Data):
     """
 
     def __new__(cls, *args, **kwargs):
-        if not hasattr(cls, "_instance"):
-            cls._instance = super(Tolerance, cls).__new__(cls)
+        if not isinstance(cls._instance, cls):
```
Licini, line 117: trying to understand this one... couldn't this be `if cls_instance is not None`? Or there are some other possible type for the instance?

Licini: mmm... not sure I understand here, under what scenario is this condition triggered? Is it equivalent for `if cls._instance is None`, or it has some wider purposes?

Licini: is it intended for inherited classes?

Licini: I mean the only possible types for `_instance` are `Tolerance` and `NoneType`, no?

Licini:
btw would it be relevant to mention angular unit ?

tomvanmele:
@Licini there are indeed more things that have to be added to the tolerance object but perhaps we can already merge the switch to the singleton?

PR #1350, opened 2024-05-02
Surface from rhino

chenkasirer:
- Fixed `Surface` pluggable `from_plane` so that it's possible to create a planar surface in Rhino.
- Added `from_native` in `native_surface` in `Surface` and `RhinoSurface` to align better with the `Brep` system.
- Added deprecation warning on `RhinoSurface.from_rhino`.
- Replaced `Box` argument with `u_domain` and `v_domain` in `RhinoSurface.from_plane`.
- Changed `new_surface` pluggin to create a `RhinoSurface` object instead of `Surface`.
- Default constructors in `Surface` and `NurbsSurface` are protected. Caller must use one of the alternative constructors.

https://miro.com/app/board/uXjVK2eWeOk=/

### Checklist


src/compas/geometry/surfaces/nurbs.py
```diff
@@ -350,6 +355,10 @@ def from_fill(cls, curve1, curve2, curve3=None, curve4=None, style="stretch"):
         """
         return new_nurbssurface_from_fill(cls, curve1, curve2, curve3, curve4, style)
 
+    @classmethod
+    def from_plane(cls, plane, u_degree=1, v_degree=1):
```
tomvanmele, line 391: perhaps it would make sense to add the domain. when creating a brep from a nurbssurface, domains are needed and that would make this a bit simpler...

src/compas_rhino/geometry/surfaces/nurbs.py
```diff
@@ -361,6 +362,32 @@ def from_fill(cls, curve1, curve2):
...
+        -------
+        :class:`compas_rhino.geometry.RhinoNurbsSurface`
+
+        """
+        plane = plane_to_rhino(plane)
+        surface = cls()
+        surface.rhino_surface = Rhino.Geometry.NurbsSurface.CreateFromPlane(
+            plane,
+            Rhino.Geometry.Interval(0, 1),
+            Rhino.Geometry.Interval(0, 1),
```
tomvanmele: would be good if this can be set by the user...

src/compas_rhino/geometry/surfaces/surface.py
```diff
@@ -168,22 +168,27 @@ def from_rhino(cls, rhino_surface):
...
 
         """
         plane = plane_to_rhino(plane)
-        box = Rhino.Geometry.BoundingBox(box.xmin, box.ymin, box.zmin, box.xmax, box.ymax, box.zmax)
-        rhino_surface = Rhino.Geometry.PlaneSurface.CreateThroughBox(plane, box)
+        if box:
+            box = Rhino.Geometry.BoundingBox(box.xmin, box.ymin, box.zmin, box.xmax, box.ymax, box.zmax)
+            rhino_surface = Rhino.Geometry.PlaneSurface.CreateThroughBox(plane, box)
+        else:
+            rhino_surface = Rhino.Geometry.PlaneSurface(plane, Rhino.Geometry.Interval(0, 1), Rhino.Geometry.Interval(0, 1))
```
tomvanmele: also in this case, it would be helpful if the user could set this. would be the alternative to providing a bounding box...

src/compas_rhino/geometry/surfaces/__init__.py
```diff
@@ -1,16 +1,13 @@
...
 from compas.geometry import NurbsSurface
 from compas.plugins import plugin
 
 
 @plugin(category="factories", requires=["Rhino"])
 def new_surface(cls, *args, **kwargs):
-    surface = super(Surface, cls).__new__(cls)
-    surface.__init__(*args, **kwargs)
-    return surface
+    return object.__new__(RhinoSurface)
```
tomvanmele: i agree that the call to `__init__` should be removed since it is not the responsibility of the `__new__` replacement. however, i think we should use `super(...).__new__` instead of `object.__new__`. in my experience, not using `super` will not work correctly in recent CPython versions...

src/compas_rhino/geometry/surfaces/surface.py
```diff
@@ -37,12 +37,19 @@ class RhinoSurface(Surface):
...
+    native_surface: :class:`Rhino.Geometry.Surface`
+        The underlying Rhino surface object.
 
     """
 
     def __init__(self, name=None):
         super(RhinoSurface, self).__init__(name=name)
-        self._rhino_surface = None
+        # empty surface is not really a valid RhinoSurface object.
+        self._rhino_surface = Rhino.Geometry.PlaneSurface(Rhino.Geometry.Plane.WorldXY, Rhino.Geometry.Interval(0, 1), Rhino.Geometry.Interval(0, 1))
```
tomvanmele: not so sure i like this. how about we add the native surface as an `__init__` parameter instead? potentially with default `None` (since i think using keyword arguments makes multiple inheritance more transparent) and a check of the value of the native surface that raises an errror if it is not a Rhino Surface...

i think the reason this didn't work properly before was because we called `__init__` in the "new plugin" which is not correct, but should be resolved now.

chenkasirer: from an API point of view, shouldn't there be just one way to do this?
`Surface(native_surface)` makes `from_native` redundant for me. 
I find `Surface.from_native(native_surface)` while a bit verbose, makes it  clear that you need a native CAD surface in order to instantiate a surface object. followin the brep convention.

Looking forward, it also leave the opening to do `Surface()` and not get an error if/when we have our own implementation of it. Same with brep.

tomvanmele: then i would just leave it as it is. creating an empty surface should not be possible in my opinion. we also don't provide a way to modify a "base" surface, so the only correct way to construct a surface is to use one of the factory functions. i find the implicit creation of planar surface rather opaque behaviour.

chenkasirer: agree!

tomvanmele: just to make sure we are talking about the same thing, with "leave it as it is", i mean as it was before the proposed change :)

chenkasirer: sorry this is taking forever, gonna try and finish this today.

To conclude my understanding and the changes in this PR:

default constructor is not allowed:
```python
from compas.geometry import Surface
Surface()

# this results in
# Runtime error (NotImplementedException): Surface() is protected. Use Surface.from_...() to construct a Surface object.
```

```python
surface = Surface.from_native(surface)
type(surface)
# <class 'compas_rhino.geometry.surfaces.surface.RhinoSurface'>
```

src/compas/geometry/surfaces/nurbs.py
```diff
@@ -94,36 +95,66 @@ def __dtype__(self):
...
-            data["weights"],
-            data["knots_u"],
-            data["knots_v"],
-            data["mults_u"],
-            data["mults_v"],
-            data["degree_u"],
-            data["degree_v"],
-            data["is_periodic_u"],
-            data["is_periodic_v"],
+            frame,
```
tomvanmele, line 141: i am not so sure that any of the methods of the surface class take this frame into account. currently, the frame is not settable and always defaults to world XY...

src/compas/geometry/surfaces/surface.py
```diff
@@ -43,15 +41,41 @@ class Surface(Geometry):
...
+            "domain_u": list(self.domain_u),
+            "domain_v": list(self.domain_v),
+        }
+
+    @classmethod
+    def __from_data__(cls, data):
+        frame = Frame.__from_data__(data["frame"])
+        domain_u = tuple(data["domain_u"])
+        domain_v = tuple(data["domain_v"])
+        return cls.from_plane(frame, domain_u, domain_v)
```
tomvanmele, line 70: since the base surface is abstract and since it should not be possible to create instances of it directly, i don't think it should provide the possibility to create one from data...

src/compas_rhino/geometry/surfaces/surface.py
```diff
@@ -27,42 +33,56 @@ class RhinoSurface(Surface):
...
         super(RhinoSurface, self).__init__(name=name)
         self._rhino_surface = None
 
+    def _get_frame_from_planesurface(self):
+        u_start = self.domain_u[0]
+        v_start = self.domain_v[0]
+        success, frame = self.native_surface.FrameAt(u_start, v_start)
+        if not success:
+            raise ValueError("Failed to get frame at u={} v={}".format(u_start, v_start))
+        return plane_to_compas_frame(frame)
```
tomvanmele, line 62: i don't think this is correct. control points are usually in world coordinates, not in local coordinates wrt to a frame positioned at the beginning of the domain.

src/compas_rhino/geometry/surfaces/surface.py
```diff
@@ -163,57 +186,64 @@ def from_rhino(cls, rhino_surface):
...
         """
-        plane = plane_to_rhino(plane)
-        box = Rhino.Geometry.BoundingBox(box.xmin, box.ymin, box.zmin, box.xmax, box.ymax, box.zmax)
-        rhino_surface = Rhino.Geometry.PlaneSurface.CreateThroughBox(plane, box)
-        return cls.from_rhino(rhino_surface)
+        instance = cls()
+        instance._rhino_surface = native_surface
+        instance.frame = instance._get_frame_from_planesurface()
+        instance._domain_u = native_surface.Domain(0)[0], native_surface.Domain(0)[1]
+        instance._domain_v = native_surface.Domain(1)[0], native_surface.Domain(1)[1]
```
tomvanmele, line 212: this doesn't make sense to me

src/compas/geometry/surfaces/surface.py
chenkasirer: `from_plane` will invoke the plugin, if available. it just seems to be the canonical surface representation (maybe with exception of `frame` which it seems I didn't quite get)

src/compas_rhino/geometry/surfaces/surface.py
chenkasirer: the domain part also?

src/compas/geometry/surfaces/surface.py
tomvanmele: but why should we provide functionality for making a "canonical" surface from data, when the surface can't actually by instantiated in the normal way. i though we agreed that only the spacial constructors can be used to create an actual surface and that the base class only provides the shared interface...

src/compas_rhino/geometry/surfaces/surface.py
tomvanmele: afaik the u and v domain are read from the native surface by the corresponding properties. they are never set...

tomvanmele: but my comment was mostly about the frame thing indeed

src/compas/geometry/surfaces/surface.py
chenkasirer: for CAD agnostic (de)serialization? maybe what's confusing me is that lack of other kinds of surface types
maybe this should be in a `PlanarSurface` like in Rhino..

tomvanmele: there are plenty of other surface types (including planar surface). they just don't require a Rhino or OCC plugin...

tomvanmele: `compas.geometry.surfaces` has `CylindricalSurface`, `ConicalSurface`, `PlanarSurface`, `SphericalSurface`, `ToroidalSurface`

chenkasirer:
replaced by https://github.com/compas-dev/compas/pull/1375

PR #1353, opened 2024-05-13
item as kwarg

Licini:
In SceneObject __init__, changing the `item` to be a `kwarg`. This will be very helpful in handling complicated multi-inheritance situations, in places like `Viewer`, where this is becoming very tricky.


src/compas/scene/sceneobject.py
```diff
@@ -82,13 +83,14 @@ class SceneObject(TreeNode):
...
 
-    def __new__(cls, item, **kwargs):
+    def __new__(cls, item=None, **kwargs):
         sceneobject_cls = get_sceneobject_cls(item, **kwargs)
         return super(SceneObject, cls).__new__(sceneobject_cls)
 
-    def __init__(self, item, name=None, color=None, opacity=1.0, show=True, frame=None, transformation=None, context=None, **kwargs):  # fmt: skip
-        # type: (compas.geometry.Geometry | compas.datastructures.Datastructure, str | None, compas.colors.Color | None, float, bool, compas.geometry.Frame | None, compas.geometry.Transformation | None, str | None, dict) -> None
-        name = name or item.name
+    def __init__(self, item=None, name=None, color=None, opacity=1.0, show=True, frame=None, transformation=None, context=None, **kwargs):  # fmt: skip
```
tomvanmele, line 90: is this to allow for the creation of a `SceneObject` without a data item? in that case, would it not make more sense to add something like `EmptySceneObject`?

Licini:
@tomvanmele @gonzalocasas @chenkasirer  Hey guys, I expanded this PR a bit to make a rather big update to clean up how data items are assigned to `SceneObjects`. 

Previously, in `SceneObject` class `__init__`, we accept the geometry/datastructrue item as the first positional argument, in form of: 
```python
class SceneObject:
    def __init__(self, item, **kwargs):
        self.item=item
        ...
```
Then its subclasses like `MeshObject` we add on top:
```python
class MeshObject(SceneObject):
    def __init__(self, mesh, **kwargs):
        super().init(item=mesh, **kwargs)
        self.mesh=mesh
        ...
```
Meanwhile we also created general sceneobject class for different context like `RhinoSceneObject` and `BlenderSceneObject` where their `__init__` does not take positional arguments, only keyword arguments:
```python
class RhinoSceneObject(SceneObject):
    def __init__(self, **kwargs):
        super().init(**kwargs)
        ...
```
At last we need to make specific sceneobject for context such as `MeshObject` for Rhino, which looks like:
```python
class MeshObject(RhinoSceneObject, BaseMeshObject):
    def __init__(self, mesh, **kwargs):
        super().init(mesh=mesh, **kwargs)
        ...
```

Especially for the last one which involves multi-inheritance, it becomes very difficult to know how to pass in data item to the super init function, whether as positional or kwargs.  And it is very often we encounter error of conflicted args or kwargs while extending further these classes. 

So the idea here is to remove all this complexity. We do not use positional arguments at all. And we only catch `item` as kwarg ONCE at the root `SceneObject` class. And if we need alias such as `MeshObject.mesh`, we make it a property that fetches `self.item`, which is the "single source of truth".  

I didn't do it for all, just few of them as a demonstration how it can look like. Please let me know if you guys think this is the good direction, then I will apply the same to all rest of `SceneObject` classes.

src/compas/scene/geometryobject.py
```diff
@@ -41,17 +41,24 @@ class GeometryObject(SceneObject):
...
+    def __init__(self, pointcolor=None, linecolor=None, surfacecolor=None, pointsize=1.0, linewidth=1.0, show_points=False, show_lines=True, show_surfaces=True, **kwargs):
+        super(GeometryObject, self).__init__(**kwargs)
+        self.pointcolor = pointcolor or self.color
+        self.linecolor = linecolor or self.color
+        self.surfacecolor = surfacecolor or self.color
+        self.pointsize = pointsize
+        self.linewidth = linewidth
+        self.show_points = show_points
+        self.show_lines = show_lines
+        self.show_surfaces = show_surfaces
```
Licini: So after adjustment, any `__init__` of subclasses should only care about additional parameters that its parent classes do not have, the data item will be passed along automatically within `**kwargs` and eventualy received by the root `__init__` of `SceneObject`

src/compas/scene/geometryobject.py
```diff
@@ -41,17 +41,24 @@ class GeometryObject(SceneObject):
...
+
+    @property
+    def geometry(self):
+        return self.item
```
Licini: If we need alias we create a property that points to `self.item` instead of creating another attribute.

src/compas/scene/graphobject.py
```diff
@@ -60,13 +58,7 @@ def __init__(self, graph, **kwargs):
 
     @property
     def graph(self):
-        return self._graph
-
-    @graph.setter
-    def graph(self, graph):
```
Licini: These setters I removed for now, because `item` in `SceneObject` is currently read-only

src/compas_rhino/scene/brepobject.py
```diff
@@ -12,17 +12,7 @@
...
-    Parameters
-    ----------
-    brep : :class:`compas_rhino.geometry.RhinoBrep`
-        The Brep to draw.
-
-    """
-
-    def __init__(self, brep, **kwargs):
-        super(RhinoBrepObject, self).__init__(geometry=brep, **kwargs)
+    """A scene object for drawing a RhinoBrep."""
```
Licini, line 15: In many cases the `__init__` function of subclass can be entirely omitted, because the don't do anything else than passing in the positional argument using a different name.

src/compas/scene/graphobject.py
```diff
@@ -46,11 +46,9 @@ class GraphObject(SceneObject):
     nodecolor = ColorDictAttribute()
     edgecolor = ColorDictAttribute()
 
-    def __init__(self, graph, **kwargs):
-        super(GraphObject, self).__init__(item=graph, **kwargs)
-        self._graph = None
+    def __init__(self, **kwargs):
```
tomvanmele: this should list `nodecolor`, `edgecolor`, `nodesize`, and not get them from `**kwargs`

src/compas/scene/meshobject.py
```diff
@@ -56,11 +56,9 @@ class MeshObject(SceneObject):
     edgecolor = ColorDictAttribute()
     facecolor = ColorDictAttribute()
 
-    def __init__(self, mesh, **kwargs):
-        super(MeshObject, self).__init__(item=mesh, **kwargs)
-        self._mesh = None
+    def __init__(self, **kwargs):
```
tomvanmele: include `vertexcolor`, `edgecolor`, `facecolor`, ... explicitly instead of processing `**kwargs`

src/compas/scene/volmeshobject.py
```diff
@@ -64,11 +64,9 @@ class VolMeshObject(SceneObject):
     facecolor = ColorDictAttribute()
     cellcolor = ColorDictAttribute()
 
-    def __init__(self, volmesh, **kwargs):
-        super(VolMeshObject, self).__init__(item=volmesh, **kwargs)
-        self._volmesh = None
+    def __init__(self, **kwargs):
```
tomvanmele: same

src/compas/scene/graphobject.py
Licini: agree, leftover from the past, will correct all of these in this PR

src/compas_blender/scene/meshobject.py
```diff
@@ -120,7 +140,7 @@ def draw_vertices(self) -> List[bpy.types.Object]:
         for vertex in vertices:
             name = f"{self.mesh.name}.vertex.{vertex}"
             color = self.vertexcolor[vertex]
-            point = self.vertex_xyz[vertex]
+            point = self.mesh.vertex_coordinates(vertex)
```
Licini, line 143: I did this here because the `update_object` will apply transformation afterwards. If using vertex_xyz, the transformation will be applied twice, similar changes are done for both `BlenderMeshObject` and `BlenderVolmeshObjectr`

Licini:
@tomvanmele Ok this one is done and ready. Could you take a look again? 
I know it's a lot files but all following same pattern.

src/compas/scene/geometryobject.py
```diff
@@ -41,16 +43,23 @@ class GeometryObject(SceneObject):
...
+        self.show_surfaces = show_surfaces
+
+    @property
+    def geometry(self):
+        """The geometry of the geometry object."""
+        return self._item
+
+    def draw(self):
+        """Draw the geometry. Implemented by child classes."""
+        raise NotImplementedError
```
tomvanmele: i don't think this is needed because the scene object already defines it

src/compas/scene/geometryobject.py
```diff
@@ -41,16 +43,23 @@ class GeometryObject(SceneObject):
...
+        self.pointsize = pointsize
+        self.linewidth = linewidth
+        self.show_points = show_points
+        self.show_lines = show_lines
+        self.show_surfaces = show_surfaces
+
+    @property
+    def geometry(self):
+        """The geometry of the geometry object."""
+        return self._item
```
tomvanmele: `self.item`

src/compas/scene/volmeshobject.py
```diff
@@ -81,11 +80,9 @@ def __init__(
         edgewidth=1.0,
         **kwargs,
```
tomvanmele: perhaps we could add `# fmt: skip` so i can add back the test for IPY

src/compas_blender/scene/circleobject.py
```diff
@@ -1,29 +1,15 @@
...
-    **kwargs : dict, optional
-        Additional keyword arguments.
-
-    """
-
-    def __init__(self, circle: Circle, **kwargs: Any):
-        super().__init__(geometry=circle, **kwargs)
+    """Scene object for drawing circles in Blender."""
 
     def draw(self, color: Optional[Color] = None, collection: Optional[str] = None) -> list[bpy.types.Object]:
```
tomvanmele, line 14: here we should also remove the parameters, but this can be in the next PR

src/compas_blender/scene/meshobject.py
```diff
@@ -153,7 +173,7 @@ def draw_edges(self) -> List[bpy.types.Object]:
         for u, v in edges:
             name = f"{self.mesh.name}.edge.{u}-{v}"
             color = self.edgecolor[u, v]
-            curve = conversions.line_to_blender_curve(Line(self.vertex_xyz[u], self.vertex_xyz[v]))
+            curve = conversions.line_to_blender_curve(Line(self.mesh.vertex_coordinates(u), self.mesh.vertex_coordinates(v)))
```
tomvanmele, line 176: why do you change this back? `vertex_xyz` are the cached coordinates...

src/compas_blender/scene/meshobject.py
```diff
@@ -177,7 +197,7 @@ def draw_faces(self) -> List[bpy.types.Object]:
         for face in faces:
             name = f"{self.mesh.name}.face.{face}"
             color = self.facecolor[face]
-            points = [self.vertex_xyz[vertex] for vertex in self.mesh.face_vertices(face)]
+            points = [self.mesh.vertex_coordinates(vertex) for vertex in self.mesh.face_vertices(face)]
```
tomvanmele, line 200: same here

src/compas_blender/scene/volmeshobject.py
```diff
@@ -152,7 +171,7 @@ def draw_edges(self) -> List[bpy.types.Object]:
         for u, v in edges:
             name = f"{self.volmesh.name}.edge.{u}-{v}"
             color = self.edgecolor[u, v]
-            curve = conversions.line_to_blender_curve(Line(self.vertex_xyz[u], self.vertex_xyz[v]))
+            curve = conversions.line_to_blender_curve(Line(self.volmesh.vertex_coordinates(u), self.volmesh.vertex_coordinates(v)))
```
tomvanmele, line 174: why do you change this back?

tomvanmele (approved):
LGTM. just a few notes/reminders

* `show_wires` (Blender) is actually the same as `show_lines`. `show_wires` is a better name in my opinion
* `disjoint` (Rhino) and `not shade_smooth` (Blender) are the same
* `Frame`, `Plane`, `Shape`, `Curve`, `Surface` should have dedicated base classes
* points of `Line`, `Polyline`, `Polygon`, `Polyhedron`, `Curve`, `Surface` are in fact "control points" and should perhaps have a dedicated color attribute for when they are turned "on".

tomvanmele:
the revert of using `self.vertex_xyz` is not correct in my opinion...

Licini:
> the revert of using `self.vertex_xyz` is not correct in my opinion...

This was because the `update_object` that is called later inside these function will apply the transformation again. So the object will be transformed twice:
```python
from compas.datastructures import Graph

graph = Graph.from_nodes_and_edges([(0, 0, 0), (0, -1.5, 0), (-1, 1, 0), (1, 1, 0)], [(0, 1), (0, 2), (0, 3)])



scene = Scene()

obj = scene.add(graph, nodesize=0.2)
obj.transformation = Translation.from_vector([10, 0, 0])
scene.draw()

```

See in picture it is moved 20 instead of 10
image

src/compas/scene/volmeshobject.py
```diff
@@ -78,7 +78,7 @@ def __init__(
         cellcolor=None,
         vertexsize=1.0,
         edgewidth=1.0,
-        **kwargs, # fmt: skip
+        **kwargs,  # fmt: skip
```
tomvanmele, line 81: this is only useful if you remove the comma after `**kwargs`

PR #1375, opened 2024-07-03
Surface API alternative PR

tomvanmele:
This PR is a refactored version of #1350 ...

### Checklist

### DIFF
--- a/src/compas/geometry/surfaces/nurbs.py
+++ b/src/compas/geometry/surfaces/nurbs.py
@@ -12,11 +12,6 @@ from .surface import Surface
 
 
 @pluggable(category="factories")
-def new_nurbssurface(cls, *args, **kwargs):
-    raise PluginNotInstalledError
-
-
-@pluggable(category="factories")
 def nurbssurface_from_cylinder(cls, *args, **kwargs):
     raise PluginNotInstalledError
 
@@ -38,6 +33,11 @@ def nurbssurface_from_frame(cls, *args, **kwargs):
 
 @pluggable(category="factories")
 def nurbssurface_from_interpolation(cls, *args, **kwargs):
+    raise PluginNotInstalledError
+
+
+@pluggable(category="factories")
+def nurbssurface_from_native(cls, *args, **kwargs):
     raise PluginNotInstalledError
 
 
@@ -157,7 +157,7 @@ class NurbsSurface(Surface):
             The constructed surface.
 
         """
-        points = [[Point.__from_data__(point) for point in row] for row in data["points"]]
+        points = data["points"]  # conversion is not needed because point data can be provided in their raw form as well
         weights = data["weights"]
         knots_u = data["knots_u"]
         knots_v = data["knots_v"]
@@ -180,9 +180,10 @@ class NurbsSurface(Surface):
             is_periodic_v,
         )
 
-    # this may not really be necessary
     def __new__(cls, *args, **kwargs):
-        return new_nurbssurface(cls, *args, **kwargs)
+        if cls is NurbsSurface:
+            raise TypeError("Instantiating the base NURBS Surface class directly is not allowed.")
+        return object.__new__(cls)
 
     def __repr__(self):
         return "{0}(points={1!r}, weigths={2}, knots_u={3}, knots_v={4}, mults_u={5}, mults_v={6}, degree_u={7}, degree_v={8}, is_periodic_u={9}, is_periodic_v={10})".format(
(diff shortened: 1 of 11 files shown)


tomvanmele:
Some explanation...

* `compas.geometry.surfaces.surface.Surface` is like an ABC for parametric surfaces.
* `compas.geometry.surfaces.surface.Surface.__new__` has been replaced such that only subclasses can be instantiated.
* `__new__` re-overwrites of subclasses are therefore no longer necessary.
* `Surface` has many implementations (Conical, Cylindrical, Planar, Spherical, Toroidal, Nurbs). only `NurbsSurface` requires a plugin implementation.
* `from_native` has been added to `Surface` since CAD engines can provide a "native" surface from which all parametric functionality can be derived, even without knowing what kind of surface it is.
* all specific factory methods have been moved to `NurbsSurface` since they indeed produce a Nurbs surface.
* elementary surfaces can be instantiated directly from the specific pure Python classes (`ConicalSurface`, `CylindricalSurface`, ...)
* a few small bugs have been fixed along the way.
* similar changes should be made to the Curve API...

all of this has been tested with OCC as plugin and using the viewer for visualisation. Rhino should still be verified.

src/compas/geometry/surfaces/nurbs.py
```diff
@@ -109,25 +144,46 @@ def __data__(self):
...
+            knots_v,
+            mults_u,
+            mults_v,
+            degree_u,
+            degree_v,
+            is_periodic_u,
+            is_periodic_v,
         )
 
+    # this may not really be necessary
```
chenkasirer: shouldn't `NurbsSurface` be made ABC as well?

src/compas/geometry/surfaces/nurbs.py
```diff
@@ -109,25 +144,46 @@ def __data__(self):
 
     @classmethod
     def __from_data__(cls, data):
+        """Construct a Nurbs surface from its data representation.
```
chenkasirer, line 147: is the domain not part of the representation? can it somehow be inferred from another piece of data?

src/compas/geometry/surfaces/nurbs.py
```diff
@@ -187,26 +244,155 @@ def degree_u(self):
...
+    def domain_v(self):
+        raise NotImplementedError
+
     # ==============================================================================
     # Constructors
     # ==============================================================================
 
+    @classmethod
+    def from_cylinder(cls, cylinder, *args, **kwargs):
+        """Construct a surface from a sphere.
```
chenkasirer:
```suggestion
        """Construct a surface from a cylinder.
```

src/compas/geometry/surfaces/nurbs.py
```diff
@@ -187,26 +244,155 @@ def degree_u(self):
...
+        Returns
+        -------
+        :class:`compas.geometry.NurbsSurface`
+
+        """
+        return nurbssurface_from_frame(cls, frame, *args, **kwargs)
+
+    @classmethod
+    def from_interpolation(cls, points, *args, **kwargs):
+        """Construct a surface from a frame.
```
chenkasirer, line 342:
```suggestion
        """Construct a surface from an interpolation.
```

src/compas/geometry/surfaces/nurbs.py
```diff
@@ -187,26 +244,155 @@ def degree_u(self):
...
             A CAD native surface object.
 
         Returns
         -------
         :class:`compas.geometry.NurbsSurface`
-            A COMPAS NURBS surface.
 
         """
-        return new_nurbssurface_from_native(cls, surface)
+        raise NotImplementedError
```
chenkasirer, line 395: call plugin here?

src/compas/geometry/surfaces/surface.py
```diff
@@ -51,7 +45,9 @@ class Surface(Geometry):
     """
 
     def __new__(cls, *args, **kwargs):
-        return new_surface(cls, *args, **kwargs)
+        if cls is Surface:
+            raise TypeError("Instantiating the base Surface class directly is not allowed.")
```
chenkasirer: I would maybe change the wording a bit. the caller doesn't necessarily know an entirely different class is gonna be created, that's abstracted by the plugin system. if I saw something like this I'd go looking for the right child class to instantiate myself. Instead, they should know the default constructor is not allowed but the other ones yes.

chenkasirer (approved):
LGTM!

PR #1385, opened 2024-08-22
Refinement of various scene functions

tomvanmele:
A few backwards compatible tweaks to the scene functionality, especially refinement of how things are cleared from the viz context...

### Checklist

### DIFF
--- a/src/compas/scene/scene.py
+++ b/src/compas/scene/scene.py
@@ -76,6 +76,14 @@ class Scene(Tree):
         # type: () -> list[SceneObject]
         return [node for node in self.nodes if not node.is_root]  # type: ignore
 
+    @property
+    def context_objects(self):
+        # type: () -> list
+        guids = []
+        for obj in self.objects:
+            guids += obj.guids
+        return guids
+
     def add(self, item, parent=None, **kwargs):
         # type: (compas.geometry.Geometry | compas.datastructures.Datastructure, SceneObject | TreeNode | None, dict) -> SceneObject
         """Add an item to the scene.
@@ -108,21 +116,32 @@ class Scene(Tree):
         super(Scene, self).add(sceneobject, parent=parent)
         return sceneobject
 
-    def clear(self):
-        # type: () -> None
-        """Clear everything from the current context of the scene."""
+    def clear_context(self, guids=None):
+        # type: (list | None) -> None
+        """Clear all objects from the visualisation context."""
+        clear(guids)
 
-        clear()
+    def clear(self, clear_scene=True, clear_context=True):
+        # type: (bool, bool) -> None
+        """Clear all objects inside the scene.
 
-    def clear_objects(self):
-        # type: () -> None
-        """Clear all objects inside the scene."""
+        Parameters
+        ----------
+        context : bool, optional
+            Clear all objects from the context as well.
 
+        """
         guids = []
+
         for sceneobject in self.objects:
             guids += sceneobject.guids
             sceneobject._guids = None
-        clear(guids=guids)
+
+            if clear_scene:
+                self.remove(sceneobject)
+
+        if clear_context:
+            self.clear_context(guids)
 
     def draw(self):
         """Draw the scene."""
@@ -132,8 +151,6 @@ class Scene(Tree):
 
         before_draw()
 
-        self.clear_objects()
-
         drawn_objects = []
         for sceneobject in self.objects:
             if sceneobject.show:
@@ -142,3 +159,8 @@ class Scene(Tree):
         after_draw(drawn_objects)
 
         return drawn_objects
+
+    def redraw(self):
+        """Redraw the scene."""
+        self.clear(clear_scene=False, clear_context=True)
+        self.draw()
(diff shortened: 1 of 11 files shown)


gonzalocasas (commented):
There's a big chunk of commented out code in the mesh object, is that replaced by anything or just removed functionality? 

Did you have the chance to verify this change also for the Grasshopper scene objects?

tomvanmele:
the commented out part is (mostly) related to labels, which i think we should handle independently from the individual types of objects in order to avoid repetition. have left it there as a reminder...

tomvanmele:
haven't checked GH, but the default behaviour of all functionality remains the same. so i don't expect any problems...

tomvanmele:
> the commented out part is (mostly) related to labels, which i think we should handle independently from the individual types of objects in order to avoid repetition. have left it there as a reminder...

in any case, the removed code paths were not (obviously) accessible nor documented so i assume there is no real issue.

src/compas/scene/scene.py
```diff
@@ -108,21 +116,32 @@ def add(self, item, parent=None, **kwargs):
...
+        # type: (bool, bool) -> None
+        """Clear all objects inside the scene.
 
-    def clear_objects(self):
-        # type: () -> None
-        """Clear all objects inside the scene."""
+        Parameters
+        ----------
+        context : bool, optional
+            Clear all objects from the context as well.
```
gonzalocasas: This could indeed have a simple test to cover it, because the `clear`/`clear_context` is already not clear enough by the naming, so, at least adding a test would ensure that this is not broken in the future by accident.

src/compas_rhino/scene/meshobject.py
```diff
@@ -215,7 +151,7 @@ def draw(self):
                 disjoint=self.disjoint,
             )
 
-            # geometry.Transform(transformation_to_rhino(self.worldtransformation))
+            geometry.Transform(transformation_to_rhino(self.worldtransformation))
```
gonzalocasas, line 158: This is weird, do you know why it was commented out and why it's needed now?

src/compas_rhino/scene/meshobject.py
```diff
@@ -264,7 +200,9 @@ def draw_vertices(self):
             elif self.group:
                 self.add_to_group(self.group, guids)
 
-        self._guids_vertices = guids
+        self._guid_vertex = dict(zip(guids, vertices))
+
+        self._guids += guids
```
gonzalocasas, line 209: This causes an error if the scene has not been drawn at least once because `self._guids` is `None`. Maybe should be initialized on `__init__`

gonzalocasas (changes requested):
I added some inline comments, but the biggest prob right now is that GH mesh object doesn't work at all (I guess it needs the same refactoring regarding `vertex_xyz`).
The other issue is that the removal of `draw_vertexlabels`, `draw_edgelabels`, `draw_facelabels`, `draw_vertexnormals`, `draw_facenormals` is not so undocumented as mentioned. We use them pretty much every single time in workshops and intro classes to compas. I don't think it can just go away.

tomvanmele:
> I added some inline comments, but the biggest prob right now is that GH mesh object doesn't work at all (I guess it needs the same refactoring regarding `vertex_xyz`). The other issue is that the removal of `draw_vertexlabels`, `draw_edgelabels`, `draw_facelabels`, `draw_vertexnormals`, `draw_facenormals` is not so undocumented as mentioned. We use them pretty much every single time in workshops and intro classes to compas. I don't think it can just go away.

ok. i can put those back. it is actually not crucial for the PR

src/compas_rhino/scene/meshobject.py
tomvanmele: scene objects can have a transformation. before it was applied through the vertex caching. now it is done explicitly here...

src/compas_rhino/scene/meshobject.py
tomvanmele: ha weird maybe i forgot to push this

tomvanmele:
will review and push new soon

src/compas_rhino/scene/meshobject.py
gonzalocasas: if you have a `scene.draw()` before trying something like `scene_object.draw_faces()` it works, but I guess it's a valid case to just draw the faces before drawing the full scene:

```
import compas
from compas.colors import Color


# Load a mesh from a sample OBJ file
mesh = Mesh.from_obj(compas.get("tubemesh.obj"))

scene = Scene()
sp = scene.add(mesh)

sp.draw_faces()  # this line fails if draw() is after
# drawing scene after faces causes the exception
# scene.draw()
```

tomvanmele: shoud be fixed now

Licini:
LGTM! but will leave @gonzalocasas to say final yes : )

gonzalocasas (approved):
LGTM! One last request though: if the `draw_vertexnormals` and `draw_facenormals` will remain commented out, please add their removal to the changelog

PR #1391, opened 2024-09-29
Fixed `compas_ghpython` is not importable in Rhino8

chenkasirer:
* Fixed `compas_ghpython` not importable in Rhino 8 CPython components, mostly by:
    * Avoiding `compas_rhino.drawing` imports within `compas_ghpython`
    * Modifying the plugin discovery logic to catch a new type of exception.
* Retired deprecated module `compas_ghpython.utilities`.

### Checklist


src/compas/plugins.py
```diff
@@ -33,6 +35,17 @@
     "PluginDefaultNotAvailableError",
 ]
 
+if compas.RHINO:
+    import System  # type: ignore  # pragma: no cover
```
tomvanmele: i would not add these pragma annotations. we don't really care about the coverage notifications, so i would also not pollute the codebase to reduce their frequency...

src/compas/plugins.py
```diff
@@ -33,6 +35,17 @@
...
+
+    DotNetException = System.Exception  # pragma: no cover
+else:
+
+    class DummyDotNetException(BaseException):
+        pass
+
+    DotNetException = DummyDotNetException
```
tomvanmele: why is this done in two steps?

src/compas/plugins.py
```diff
@@ -401,15 +414,15 @@ def try_import(self, module_name):
...
         try:
             module = __import__(module_name, fromlist=["__name__"], level=0)
             self._cache[module_name] = True
 
         # There are two types of possible failure modes:
         # 1) cannot be imported, or
         # 2) is a python 3 module and we're in IPY, which causes a SyntaxError
-        except (ImportError, SyntaxError):
+        # 3) in Rhino8 we may get a nasty DotNet Exception when trying to import a module
+        except (ImportError, SyntaxError, DotNetException):
```
tomvanmele, line 425: does this mean, when not in Rhino, we accept `ImportError`, `SyntaxError` AND `BaseException`? does this mean we accept ALL exceptions?

src/compas_ghpython/scene/meshobject.py
```diff
@@ -147,11 +146,21 @@ def draw_faces(self):
...
+
+    @staticmethod
+    def _create_ngon(vertex_count):
+        if vertex_count < 3:
+            return
+        if vertex_count == 3:
+            return [0, 1, 2]
+        if vertex_count == 4:
+            return [0, 1, 2, 3]
+        return list(range(vertex_count))
```
tomvanmele: no reason to attach this to the object

src/compas_ghpython/scene/volmeshobject.py
```diff
@@ -115,3 +114,13 @@ def draw_cells(self):
...
+
+    @staticmethod
+    def _create_ngon(vertex_count):
+        if vertex_count < 3:
+            return
+        if vertex_count == 3:
+            return [0, 1, 2]
+        if vertex_count == 4:
+            return [0, 1, 2, 3]
+        return list(range(vertex_count))
```
tomvanmele: same

tomvanmele (commented):
other than a few questions, i think this is great!

src/compas/plugins.py
chenkasirer: what do you mean two steps? why I'm not using `BaseException` directly?

src/compas/plugins.py
chenkasirer: outside Rhino we would accept (in addition to the other two) a `DummyDotNetException` which is never really raised anywhere, so as good as ignored
Outside of Rhino any import error should always be one of `ImportError` and `SyntaxError` anyways

src/compas/plugins.py
tomvanmele: yes, why not `class DotNetException(BaseException): pass`

tomvanmele (approved):
LGTM, with one comment

src/compas/plugins.py
chenkasirer: no other reason except it never crossed my mind :)

PR #1399, opened 2024-10-15
Fix `scaled` in `Box`

gonzalocasas:
Tiny bug fix

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 * Fixed `RuntimeError` when using `compas_rhino.unload_modules` in CPython`. 
+* Fixed `scaled` in `Box`.
 
 ### Removed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 * Fixed `RuntimeError` when using `compas_rhino.unload_modules` in CPython`. 
 * Fixed `scaled` in `Box`.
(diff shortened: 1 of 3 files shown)


src/compas/geometry/shapes/box.py
```diff
@@ -575,6 +575,28 @@ def scale(self, factor):
         self.ysize *= factor
         self.zsize *= factor
 
+    def scaled(self, x, y=None, z=None):
```
chenkasirer, line 585: why not make `scale` the same and then use it here?

gonzalocasas: yes, of course, you're totally right, changed it here: https://github.com/compas-dev/compas/pull/1399/commits/79596048e0284084a0ec4b360f193d8fb6deaae2

this is now a breaking change because the argument names have changed, even if positionally it would be a non-breaking change

CHANGELOG.md
```diff
@@ -12,6 +12,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 * Fixed `RuntimeError` when using `compas_rhino.unload_modules` in CPython`. 
+* Fixed `scaled` in `Box`.
```
chenkasirer: I assume the original issue is about `Geometry.scaled` not working quite right for `Box` (I think it's using `transfome(Scale(...)))`..? did it just move `Box.frame` or something? perhaps a better description of what used to happen here?

chenkasirer (approved):
LGTM!
just a small comment on the log entry

CHANGELOG.md
gonzalocasas: Nope, actually, `Box.scaled()` fails completely with `TypeError: scale() got an unexpected keyword argument 'x'`

gonzalocasas: Updated the changelog!

Issue #1402, opened 2024-10-22
`Data.copy` silently adding `name` attribute


yck011522:
**Describe the bug**
I found that the Data.copy() mechanism somehow added the `name` attribute in the `json_string` after the copy. This affects native compass geometry classes (e.g. Frame) and also classes that I have inherited from data.

The addition of this attribute only happens in the `to_jsonstring()` but not if I read the `__data__`. However, it changes the result of `sha256()` and the copied object will return a differnt hash. (see example code below)

Now, I'm not sure if this is the intended behavior (for version control?). My goal is to use a hash function to compare the data content of the objects. I assumed the `sha256()` would be for this purpose, but maybe it is not? If not, can you maybe clarify what is the best practice for comparing the data content between two objects, especially when the data contains geometry, str, and list of things. 

**To Reproduce**
The following example shows not only a problem related to the addition of `name` attribute, but also floating point difference during the copy of the frame. Both of which would throw off the hash comparison.
```
if __name__ == "__main__":
    import compas
    print(compas.__version__)

    frame = Frame([1, 2, 3], [0.1, 0.2, 0.3])

    print(" to_jsonstring(): ")
    print(frame.to_jsonstring())
    print(frame.copy().to_jsonstring())

    print(" __data__: ")
    print(frame.__data__)
    print(frame.copy().__data__)

    print(" sha256(): ")
    print(frame.sha256())
    print(frame.copy().sha256())
```

output:
```
2.4.2
 to_jsonstring():
{"dtype": "compas.geometry/Frame", "data": {"point": [1.0, 2.0, 3.0], "xaxis": [0.2672612419124244, 0.5345224838248488, 0.8017837257372731], "yaxis": [-0.16903085094570336, 0.8451542547285167, -0.50709255283711]}, "guid": "fe05d20f-69b4-4bc0-ba25-88e8d1404933"}
{"dtype": "compas.geometry/Frame", "data": {"point": [1.0, 2.0, 3.0], "xaxis": [0.26726124191242445, 0.5345224838248489, 0.8017837257372732], "yaxis": [-0.16903085094570341, 0.8451542547285167, -0.5070925528371101]}, "name": "Frame", "guid": "3a58794d-273c-4285-b875-cbff06e68c34"}
 __data__:
[...]

PR #1403, opened 2024-10-24
Fix bug in data copy name

tomvanmele:

### Checklist

### DIFF
--- a/tests/compas/data/test_copy.py
+++ b/tests/compas/data/test_copy.py
@@ -1,0 +1,21 @@
+from compas.data import Data
+
+
+class TestData(Data):
+    @property
+    def __data__(self):
+        return {}
+
+
+def test_copy():
+    data = TestData()
+
+    assert data.name == data.copy().name
+    assert data.__dtype__ == data.copy().__dtype__
+
+    assert data.guid != data.copy().guid
+    assert data.__jsondump__() != data.copy().__jsondump__()
+    assert data.to_jsonstring() != data.copy().to_jsonstring()
+
+    assert data.__jsondump__(minimal=True) == data.copy().__jsondump__(minimal=True)
+    assert data.to_jsonstring(minimal=True) == data.copy().to_jsonstring(minimal=True)
--- a/tests/compas/data/test_copy.py
+++ b/tests/compas/data/test_copy.py
@@ -7,15 +7,57 @@ class TestData(Data):
         return {}
 
 
-def test_copy():
+def test_copy_noname():
     data = TestData()
+
+    assert data._name is None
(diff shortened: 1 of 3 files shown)


tests/compas/data/test_copy.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+class TestData(Data):
+    @property
+    def __data__(self):
+        return {}
+
+
+def test_copy():
+    data = TestData()
```
Licini, line 11: Perhaps you need to assign a name in your test before copying the object, in order the cover Line294. Otherwise it is just `None`, right?

tomvanmele: the problem does not arise when a name is assigned, but when there isn't. when no name is assigned, the name attribute is not included in the serialisation info. however, (before this PR) when a copy is made the name attribute is assigned regardless. this means that the serialised info before and after copy don't match. that is what this PR fixes. the test as written, would have failed before this PR...

tomvanmele:
this is partially related to #1402

yck011522:
Thanks for the fix. One more additional thought:

I'm looking at your test file, and it appears that you intended to have the `.copy()` function return a new object that has a new guid. While I have no problem with that, I wonder if you would consider allowing the user to pass an optional flag, so that they can do a guid-preserving copy?

Such that the following can be true too:
```
assert data.__jsondump__() == data.copy(preserve_guid=True).__jsondump__()
assert data.to_jsonstring() == data.copy(preserve_guid=True).to_jsonstring()
```

tomvanmele:
@yck011522 i can, but that raises questions about the value of a guid since both copies will have the same "unique" identifier...

this is also the case before and after serialisation btw...

tests/compas/data/test_copy.py
Licini: ah I see the reason. I was suggesting this only because of the Codecov warning: 
image
How about doing this test both with and without assigning name?

Licini (approved):
Looking good! But better if possible to get rid of Codedev warning : )

yck011522:
@tomvanmele 
I think my experience with other software is that copy will result in new GUID, while serialization and deserialization should preserve it. Except in the case that there is a GUID collision in the 'current document' (probably not relevant to us as we don't have the concept of a document object table. 
I don't know where this experience came from, probably just form experience using Rhino and GH. 

However, I would still like to see an option for the user to do copy while preserving GUID, there are use cases that may need it.

tomvanmele:
how about now folks?

yck011522:
Yea, I think its great.

yck011522 (approved):

PR #1407, opened 2024-11-01
fixed Sphere._vertices used before populated

chenkasirer:
Small fix in `Sphere` of issue causing `Sphere.edges` and `Sphere.faces` to fail if calls before first call to `Sphere.vertices`.

### Checklist

### DIFF
--- a/tests/compas/geometry/test_shpere.py
+++ b/tests/compas/geometry/test_shpere.py
@@ -1,0 +1,15 @@
+import pytest
+
+from compas.geometry import Frame
+from compas.geometry import Sphere
+
+
+@pytest.fixture
+def sphere():
+    return Sphere(450.0, Frame.worldXY())
+
+
+def test_sphere_discretization(sphere):
+    assert len(sphere.edges) == 496
+    assert len(sphere.faces) == 256
+    assert len(sphere.vertices) == 242
--- a/tests/compas/geometry/test_shpere.py
+++ b/tests/compas/geometry/test_shpere.py
@@ -10,6 +10,10 @@ def sphere():
 
 
 def test_sphere_discretization(sphere):
-    assert len(sphere.edges) == 496
-    assert len(sphere.faces) == 256
-    assert len(sphere.vertices) == 242
+    expected_face_count = sphere.resolution_v * sphere.resolution_u
+    expected_vertex_count = (sphere.resolution_v - 1) * sphere.resolution_u + 2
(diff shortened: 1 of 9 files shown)


tests/compas/geometry/test_shpere.py
```diff
@@ -0,0 +1,15 @@
...
+
+@pytest.fixture
+def sphere():
+    return Sphere(450.0, Frame.worldXY())
+
+
+def test_sphere_discretization(sphere):
+    assert len(sphere.edges) == 496
+    assert len(sphere.faces) == 256
+    assert len(sphere.vertices) == 242
```
tomvanmele: would be even better to relate these assertions to numbers derived from the discretisation params (u, v)

tomvanmele:
this is happening because i (wrongly) imagined/assumed that faces or edges would be accessed after vertices, and i wanted to avoid recomputing the vertices over and over again, if not necessary.

intention was good, implementation bad :)

would be nice if we can find another way to handle this...

chenkasirer:
Well I mean edges and faces are cached so the the compute methods should only be called once.. I assumed the vertices are not cached for the case resolution is changed, but since edges and faces are perhaps we should cache vertices as well?

We could maybe use a reset computed kind of decorator to invalidate caches..

tomvanmele:
would be nice but i don't see an easy solution. we could monitor the shape's own properties but modifications can be made at a deeper level and then things become tricky. for example

```python
cylinder = Cylinder(radius=1.0, height=1.0)

# all of the following would have to trigger a cache reset
cylinder.radius = ...
cylinder.height = ...
cylinder.resolution_u = ...
cylinder.resolution_v = ...
cylinder.frame = ...
cylinder.frame.point = ...
cylinder.frame.point.x = ...
...
```

not sure how we can elegantly track this.
the direct property changes are easy, but the nested/deeper ones are not so obvious to me.

tomvanmele:
anyway perhaps that is not the point of this PR...

tomvanmele (approved):
LGTM

chenkasirer:
> not sure how we can elegantly track this. the direct property changes are easy, but the nested/deeper ones are not so obvious to me.

I see, indeed tricky. 
honestly I don't think we need to guarantee consistency if someone manipulates deep attributes.
I guess the only thing we can do about that is make these shapes immutable by e.g. returning copies from getters but not sure if we want to do that.

PR #1409, opened 2024-11-08
Fix friday

tomvanmele:

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,8 +20,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Fixed bug in `VolMesh.vertex_cells`.
 * Fixed bug in `VolMesh.is_halfface_on_boundary`.
 
-
 ### Removed
+
+* Removed `VolMesh.halfface_adjacent_halfface` because of general nonsensicalness, and because it is (and probably always has been) completely broken.
 
 
 ## [2.5.0] 2024-10-25
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,11 +10,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 * Added key conversion map to `compas.colors.ColorDict` to avoid serialisation problems with tuple keys when used in combination with edges.
+* Added `Scene.find_all_by_itemtype`.
 
 ### Changed
 
 * Fixed bug in `VolMesh.delete_cell`.
 * Fixed `NoneType` error when calling `compas.geometry.Sphere.edges`.
+* Fixed bug in `VolMesh.vertex_halffaces`.
+* Fixed bug in `VolMesh.vertex_cells`.
+* Fixed bug in `VolMesh.is_halfface_on_boundary`.
 
 
 ### Removed
--- a/src/compas/datastructures/volmesh/volmesh.py
+++ b/src/compas/datastructures/volmesh/volmesh.py
@@ -383,7 +383,7 @@ class VolMesh(Datastructure):
 
     @classmethod
     def from_vertices_and_cells(cls, vertices, cells):
-        # type: (list[list[float]], list[list[list[int]]]) -> VolMesh
+        # type: (list[list[float]] | dict[int, list[float]], list[list[list[int]]]) -> VolMesh
         """Construct a volmesh object from vertices and cells.
 
         Parameters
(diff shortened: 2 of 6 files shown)


tomvanmele:
btw, @gonzalocasas the real changes to volmesh are in lines 1764 and onwards. the rest is just annoying code coverage feedabck because i added a few `# type: ignore` comments

tomvanmele:
the section i commented out is a function that is not only broken , but also doesn't seem to make much sense. will remove it if there is no objection...

src/compas/scene/meshobject.py
```diff
@@ -106,16 +106,6 @@ def mesh(self, mesh):
...
-    @property
-    def transformation(self):
-        # type: () -> compas.geometry.Transformation | None
-        return self._transformation
-
-    @transformation.setter
-    def transformation(self, transformation):
-        # type: (compas.geometry.Transformation) -> None
-        self._transformation = transformation
-
```
Licini, line 118: No transformation anymore?

Licini (approved):
LGTM

src/compas/datastructures/volmesh/volmesh.py
```diff
@@ -3108,36 +3111,36 @@ def halfface_opposite_halfface(self, halfface):
...
-        The adjacent face belongs a to one of the cell neighbors over faces of the initial cell.
-        A face and its adjacent face share two common vertices.
-
-        """
-        u, v = halfedge
-        cell = self.halfface_cell(halfface)
-        nbr_halfface = self._cell[cell][v][u]
-        nbr_cell = self._plane[u][v][nbr_halfface]
-        return None if nbr_cell is None else self._cell[nbr_cell][v][u]
+    # def halfface_adjacent_halfface(self, halfface, halfedge):
```
gonzalocasas: This would be a breaking change

src/compas/datastructures/volmesh/volmesh.py
```diff
@@ -3108,37 +3111,6 @@ def halfface_opposite_halfface(self, halfface):
         nbr = self._plane[w][v][u]
         return None if nbr is None else self._cell[nbr][w][v]
 
-    def halfface_adjacent_halfface(self, halfface, halfedge):
```
gonzalocasas, line 3111: This would be a breaking change, it shouldn't be included if this is going to be a patch release.

src/compas/scene/meshobject.py
gonzalocasas, line 118: Same here, this is a breaking change, and maybe even not that small. It's not clear why this is needed, and what's the new way of doing the same thing if this property goes away

src/compas/scene/scene.py
```diff
@@ -245,3 +245,11 @@ def find_by_itemtype(self, itemtype):
         for obj in self.objects:
             if isinstance(obj.item, itemtype):
                 return obj
+
+    def find_all_by_itemtype(self, itemtype):
```
gonzalocasas, line 249: docstring?

gonzalocasas (commented):
Tested `is_halfface_on_boundary` and it works, I'm worried about the deletion of public methods, because this is not going to be a new major release

src/compas/datastructures/volmesh/volmesh.py
tomvanmele: sure but the method absolutely doesn't work and i don't even know what it is supposed to do...

src/compas/scene/meshobject.py
tomvanmele: this is just duplication form functionality of the base class

tomvanmele: the parent already defines it so it should not have been here to begin with. functionality will not change...

src/compas/scene/scene.py
tomvanmele: sure

src/compas/scene/meshobject.py
gonzalocasas: Ah! Got it!

src/compas/datastructures/volmesh/volmesh.py
tomvanmele: it can't be that anyone has been using this because the description is nonsensical and the implementation doesn't work. we can leave it in, but i don't know how to fix it because i don't even know what it is supposed to do...

src/compas/datastructures/volmesh/volmesh.py
gonzalocasas: hehe ok, let's remove it, maybe adding a clear notice in the changelog

CHANGELOG.md
```diff
@@ -20,9 +20,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Fixed bug in `VolMesh.vertex_cells`.
 * Fixed bug in `VolMesh.is_halfface_on_boundary`.
 
-
 ### Removed
 
+* Removed `VolMesh.halfface_adjacent_halfface` because of general nonsensicalness, and because it is (and probably always has been) completely broken.
```
gonzalocasas, line 25: 😝

gonzalocasas (approved):

Issue #1411, opened 2024-11-14
unify_cycles / face_adjaccency


romanarust:
**Describe the bug**

I have a set of [vertices and faces](https://gist.githubusercontent.com/romanarust/58e5dfea928ecd2207dd21d5b512ba4e/raw/7c0cad520963c2d2fa013e0b9d48285934de23d7/vertices_faces.json) from a correct, closed mesh, but the function `unify_cycles` fails. The reason for this is that when there are more than [100 faces](https://github.com/compas-dev/compas/blob/46b286311ccf0212f65fa202684c978a1d6bf560/src/compas/topology/orientation.py#L122), a different function (`_face_vertices`) is used for face adjacency. However, in my case, this function produces a different, and wrong result than `face_adjacency`, because the parameters of this function aren’t dependent on the input.

How would it be best to proceed: I could propose a fix, but should the parameters be determined based on the input (i.e determine the radius on the longest edge?), or should the user be allowed to select which function to use (`unify_cycles(fast=True)`)? My understanding is that the `_face_adjacency` function was intended to increase performance. Wouldn’t it make sense for this to be user-controlled?

tomvanmele:
hi romana,

is `radius=10` the source of the problem?

tomvanmele:
perhaps both?

* user is allowed to choose, with the default switch for over 100 faces
* radius is None by default and in that case is computed based on some kind of heuristic
* radius is available as a top-level parameter so it can be set by the user in case of awkward mesh geometries

romanarust:
ok thank you, I'll propose something

PR #1413, opened 2024-11-14
update unify cycles

romanarust:
As noted in https://github.com/compas-dev/compas/issues/1411 there is a problem with face_adjacency on meshes with arbitrary-sized faces.  I've now exposed the parameters `radius` and `nmax` from `compas.topology._face_adjacency` to `compas.topology.face_adjacency` and further propagated them to `unify_cycles` and `Mesh.unify_cycles`.

### Checklist


src/compas/datastructures/mesh/mesh.py
```diff
@@ -2932,9 +2932,18 @@ def quads_to_triangles(self, check_angles=False):
                     del self.facedata[face]
 
     # only reason this is here and not on the halfedge is because of the spatial tree
-    def unify_cycles(self, root=None):
+    def unify_cycles(self, root=None, nmax=None, radius=None):
         """Unify the cycles of the mesh.
 
+        Parameters
+        ----------
+        root : str, optional
```
tomvanmele: type of root is `int`

src/compas/topology/orientation.py
```diff
@@ -166,6 +172,10 @@ def unify_cycles(vertices, faces, root=None):
         The faces of the mesh defined as lists of vertex indices.
     root : str, optional
```
tomvanmele: same here

src/compas/topology/orientation.py
```diff
@@ -117,10 +121,12 @@ def face_adjacency(points, faces):
...
-    f = len(faces)
-
-    if f > 100:
-        return _face_adjacency(points, faces)
+    if nmax is not None:
+        if radius is not None:
+            return _face_adjacency(points, faces, nmax=nmax, radius=radius)
+        return _face_adjacency(points, faces, nmax=nmax)
+    if radius is not None:
+        return _face_adjacency(points, faces, radius=radius)
```
tomvanmele: perhaps it would make sense to refactor `_face_adjacency` such that you can write here
```python
if nmax or radius:
    return _face_adjacency(points, faces, nmax=nmax, radius=radius)
```

tomvanmele:
btw, since the switch is no longer based on the number of faces, it should be easier to test using generated boundary cases...

romanarust:
> btw, since the switch is no longer based on the number of faces, it should be easier to test using generated boundary cases...

correct. do you have something in mind or should I leave the tests for now?

tomvanmele:
was thinking that we could use the meshgrid with awkward proportions in x and y to generate inputs where radius and nmax provide very different results in different directions...

tomvanmele (dismissed):
LGTM

tests/compas/topology/test_unify_cycles.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+def test_unify_cycles():
+    if compas.IPY:
+        return
+    test_data = compas.json_load(os.path.join(HERE, "..", "fixtures", "topology", "vertices_faces.json"))
+    vertices = test_data["vertices"]
+    faces = test_data["faces"]
+    unify_cycles(vertices, faces)
+    unify_cycles(vertices, faces, nmax=29, max_distance=22.4)  # anything below won't work
```
tomvanmele: is it clear why that is the case?
or it is just something you noticed?

tests/compas/topology/test_unify_cycles.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+def test_face_adjacency():
```
tomvanmele: why is this a face adjacency test and not also a test of unify cycles?

tomvanmele (commented):
everything looks good but the tests are a bit confusing.
should we not include an assertion of some sorts?
also, if you know when the tests are supposed to fail, should we then not add a `@pytest.mark.xfail`?

tomvanmele:
@romanarust perhaps just a few things to sort out regarding the test(s)

romanarust:
> @romanarust perhaps just a few things to sort out regarding the test(s)

Thanks! I've added now some `assert` to the tests.

tomvanmele (approved):
LGTM

PR #1415, opened 2024-12-13
Frame inversion

tomvanmele:

### Checklist


src/compas/geometry/frame.py
```diff
@@ -604,6 +604,18 @@ def to_transformation(self):
     # Methods
     # ==========================================================================
 
+    def invert(self):
```
chenkasirer, line 607: not sure but should this be maybe called `flip`?

src/compas/geometry/frame.py
```diff
@@ -604,6 +604,18 @@ def to_transformation(self):
...
+        """Invert the frame while keeping the X axis fixed."""
+        self._yaxis = self.yaxis * -1
+        self._zaxis = None
+
+    def inverted(self):
+        """Return an inverted copy of the frame."""
+        frame = self.copy()  # type: Frame
+        frame.invert()
+        print(frame.xaxis, frame.yaxis, frame.zaxis)
```
chenkasirer: remove `print`?

chenkasirer (approved):
lgtm
just a couple of small comments

tests/compas/geometry/test_frame.py
```diff
@@ -92,3 +92,35 @@ def test_interpolate_frame_start_end():
 
     three_quarter_frame = frame1.interpolate_frame(frame2, 0.75)
     assert TOL.is_allclose([math.degrees(three_quarter_frame.axis_angle_vector.y)], [-67.5], atol=TOL.angular)
+
+
+def test_frame_invert():
+    frame = Frame([0, 0, 0])
+
+    assert TOL.is_close(frame.xaxis.dot([1, 0, 0]), 1.0)
```
chenkasirer, line 100: just out of curiosity what's the advantage of this over `TOL.is_allclose(frame.xaxis, [1, 0, 0])`

src/compas/geometry/frame.py
tomvanmele: in `Vector` it is also `invert`, but we can add an alias in both perhaps

tests/compas/geometry/test_frame.py
tomvanmele: no advantage. am just used to expressing this using dot

PR #1426, opened 2025-01-20
Rectangle xy

### LINKED ISSUE #1422: Unexpected behavior of "bbox_numpy.minimum_area_rectangle_xy"
**Describe the bug**
Tried to compute a xy oriented minimum bounding box. Sometimes it works, sometimes the base rectangle appears to be mirrored.

**To Reproduce**
1. Context: Rhino / Grasshopper
2. Sample Python 3 script in GH

```
# requirements: compas==2.8.1
# requirements: numpy

from compas.geometry import bbox_numpy, Point
import numpy as np

file_path = "..."

compas_mesh = Mesh.from_obj(file_path)

compas_pts, _ = compas_mesh.to_vertices_and_faces()

arr = np.array(compas_pts)

minimum_area_rect = bbox_numpy.minimum_area_rectangle_xy(arr)

pts = []

for pt in minimum_area_rect:
    pts.append(Point(pt[0], pt[1], 0.0))

scene = Scene()
scene.add(pts[0])
scene.add(pts[1])
scene.add(pts[2])
scene.add(pts[3])

a = scene.draw()
```

4. See OBJ files: [meshes.zip](https://github.com/user-attachments/files/18427113/meshes.zip)

**Expected behavior**

This:
image

**Unexpected behavior**

It seems like the minimum area rectangle is mirrored. The only change is moving the mesh on the canvas, with all coordinates remaining positive.
image

**Expected goal**

xy_ombb rotated to align with the coordinate system. 
image

 - OS: iOS
 - Python version == 3.12.8
 - COMPAS version == 2.8.1
 - Python package manager: conda / pip

chenkasirer:
fixes https://github.com/compas-dev/compas/issues/1422

fixed this issue causing `minimum_area_rectangle_xy`  to produce a mirrored result for the same Mesh due to slight variations in vertex values.

### Checklist


tomvanmele:
does this also produce the expected result if you go through the oriented bounding box function? it should detect that the > input data is essentially flat and use the rectangle function automatically...

chenkasirer:
> does this also produce the expected result if you go through the oriented bounding box function? it should detect that the > input data is essentially flat and use the rectangle function automatically...

Haven't tested myself but it appears there's unittest that verifies that https://github.com/compas-dev/compas/blob/2c7ed4ee1bfea86d9547b9932045a41f3375fff7/tests/compas/geometry/test_bbox.py#L67

in any case, this is determined prior to the call to `minimum_area_rectangle_xy` where my change is.
there, everything is calculated exactly as before, the only change is that the centroid of the convex hull is used to determined if a vector perpendicular to the segments points inwards or outwards of the shape. these are quite wild and using mean for the same purpose wasn't as robust.

image

tomvanmele (approved):

Issue #1351, opened 2024-05-02
Rhino scripts default install path is wrong for Rhino 8


obucklin:
`python -m compas_rhino.install` adds the COMPAS packages to `~/.rhinocode/py27-rh8/Lib/site-packages` instead of `\AppData\Roaming\McNeel\Rhinoceros\8.0\scripts`. Grasshopper components cannot then import the modules.  

there is a workaround which is to include 

```
pt = r"C:\Users\obucklin/.rhinocode/py27-rh8/Lib/site-packages"

if pt not in sys.path:
    sys.path.append(pt)
```

at the top of a python component. This fixes IronPython components.

I'm not sure if it's the same problem, but cpython components built with `invoke build-cpython-ghuser-components` do not instantiate on the canvas and result in a `User Object failed to deserialize` error:

![image](https://github.com/compas-dev/compas/assets/55446042/a8cac637-e292-4b60-a312-f7b0147091ab)

Windows 11
Rhino 8 SR6 2024-4-10
COMPAS 2.1.0

Thanks!

petrasvestartas:
I guess ScriptEditor usage for Rhino8 is still experimental as stated in the Compas documentation:
![image](https://github.com/compas-dev/compas/assets/18013985/51949462-797c-44e2-aef4-58bf5504a015)
https://github.com/compas-dev/compas/blob/08f06247cfdd99c4fdb029a4d74430ca0efed444/src/compas_rhino/install.py#L35
Python library installation directory and Grasshopper are two different directories.

What works for me is following:

Grasshopper for ScriptEditor Rhino8 CPython on Windows:
`python src\rhino\gh\componentize_cpy.py src\rhino\gh\cpy %APPDATA%\Grasshopper\UserObjects`

on mac
`python src/rhino/gh/componentize_cpy.py src/rhino/gh/cpy  "/Users/petras/Library/Application Support/McNeel/Rhinoceros/8.0/Plug-ins/Grasshopper (b45a29b1-4343-4035-989e-044e8580d9cf)/UserObjects"`

or the package manager directory (manual upload to Rhino).

The python libraries I install as described in compas main page:
https://compas.dev/compas/latest/userguide/cad.rhino8.html
When a library is ready, I do it via pip, since this is the only possible solution to link a library in script editor is via `# r: `keyword.
Conda environments will be supported "maybe" in the future. This whole thing is painful, but at least it is CPython not IronPython...

ChiaChing-Yen:
[...]

PR #1427, opened 2025-01-20
Rhino8 gh

chenkasirer:
part of https://github.com/compas-dev/compas/issues/1351
needs https://github.com/compas-dev/compas_invocations2/pull/4

* Added CPython implementation of GH components for Rhino8
* Added tasks to yakerize and publish
* Added new workflow for publishing as yak package
* Removed support for `compas_rhino.install -v8.0`

### Checklist


src/compas_ghpython/components_cpython/Compas_FromJson/code.py
```diff
@@ -0,0 +1,20 @@
+"""
+Deserializes JSON into COMPAS objects.
+"""
+
+# r: compas==2.8.1
```
tomvanmele: why pin the version?

tomvanmele: does this work if not at the top of the script?

src/compas_ghpython/components_cpython/Compas_FromJson/code.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+class CompasInfo(component):
+    def RunScript(self, json):
+        if not json:
+            return None
+
+        try:
+            return compas.json_load(json)
+        except:  # noqa: E722
```
tomvanmele: even better would be to not leave the `except` naked, no?

src/compas_ghpython/components_cpython/Compas_RpcCall/code.py
```diff
@@ -0,0 +1,32 @@
...
+Remote Procedure Call: to invoke Python functions outside of Rhino, in the context of the CPython interpreter.
+"""
+
+# r: compas==2.8.1
+from ghpythonlib.componentbase import executingcomponent as component
+
+from compas.rpc import Proxy
+
+
+class CompasRpcCall(component):
```
tomvanmele: maybe consider `CompasRemoteCall`... procedure seems not relevant to mention and `c` and `Call` are the same.

src/compas_ghpython/components_cpython/Compas_FromJson/code.py
chenkasirer: yes it works when it's there.

about the pinning, I was thinking it might be important when e.g. creating a component that uses features available in the latest release only. since Rhino doesn't otherwise manage component->library dependencies, this seems like the only way.

With that in mind, it might make sense that we extend our componentizer to inject these headers from a proper list of dependencies list in the metadata. otherwise, it's a pain in the ass to keep these up to date. maybe I'll change to `compas~=2.8.1` (if that even works..)

tomvanmele: but why not `compas>=2.8.1`

jf---:
Stoked about the yakerize feature!

gonzalocasas (approved):
YAY!!! LGTM! 🚀 
Do we need to update the installation docs to indicate that 8.0 is not installable the same way or is it already updated?

chenkasirer:
> YAY!!! LGTM! 🚀 Do we need to update the installation docs to indicate that 8.0 is not installable the same way or is it already updated?

the Rhino8 installation instructions are already up to date. I will add a section about installation from the package manager in another PR.

PR #1435, opened 2025-02-05
fixed unexpected behavior for opposite parallel planes.

mattiskoh:
`Plane.is_parallel()` does not check for opposite normals and thus returns False for this edge case. 

Not the most elegant fix due to the `TOL.is_close()` hope it's fine like this..

### Checklist

- [] I added new functions/classes and made them available on a second-level import, e.g. `compas.datastructures.Mesh`.

### DIFF
--- a/src/compas/geometry/plane.py
+++ b/src/compas/geometry/plane.py
@@ -390,9 +390,7 @@ class Plane(Geometry):
         True
 
         """
-        is_parallel = TOL.is_close(self.normal.dot(other.normal), 1, rtol=0, atol=tol)
-        is_parallel_neg = TOL.is_close(self.normal.dot(other.normal), -1, rtol=0, atol=tol)
-        return is_parallel or is_parallel_neg
+        return TOL.is_close(abs(self.normal.dot(other.normal)), 1, rtol=0, atol=tol)
 
     def is_perpendicular(self, other, tol=None):
         """Verify if this plane is perpendicular to another plane.
--- a/src/compas/geometry/plane.py
+++ b/src/compas/geometry/plane.py
@@ -384,9 +384,15 @@ class Plane(Geometry):
         >>> plane2 = Plane([1.0, 1.0, 1.0], [0.0, 0.0, 1.0])
         >>> plane1.is_parallel(plane2)
         True
+        >>> plane1 = Plane.worldXY()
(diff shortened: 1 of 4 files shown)


src/compas/geometry/plane.py
```diff
@@ -384,9 +384,15 @@ def is_parallel(self, other, tol=None):
...
         >>> plane1.is_parallel(plane2)
         True
+        >>> plane1 = Plane.worldXY()
+        >>> plane2 = Plane([1.0, 1.0, 1.0], [0.0, 0.0, -1.0])
+        >>> plane1.is_parallel(plane2)
+        True
 
         """
-        return TOL.is_close(self.normal.dot(other.normal), 1, rtol=0, atol=tol)
+        is_parallel = TOL.is_close(self.normal.dot(other.normal), 1, rtol=0, atol=tol)
```
tomvanmele: why not just take the absolute value of the dot product?

tomvanmele (commented):
thanks for noticing this and proposing a fix...

src/compas/geometry/plane.py
mattiskoh: agreed. I've made the changes.

tomvanmele (commented):
the fix is great, but could you add a test that proves the problem is solved?

mattiskoh:
> the fix is great, but could you add a test that proves the problem is solved?

Isn't the Docstring Example sufficient..? Or do you guys not run `doctest` ? Better said why put an example if there is no test for this function...?

tomvanmele:
the examples in the docs are tested. however, examples are not necessarily meant to be formulated as tests, but rather explanations for users on how to use the advertised functionality. in many case some of the examples will indeed overlap with some of the tests, but ideally tests for example also address things like boundary cases, which simple examples usually don't.

currently our test coverage is not nearly sufficient, but we are trying to work towards this. therefore i try to make sure that at least new PRs are accompanied by tests, especially when they fix an issue.

i know it is annoying but it needs to be done...

tomvanmele (approved):

jf---:
nice find @mattiskoh 

I wonder whether the fix can be generalised; that is, is a tolerance is an unsigned value in general? Intuitively the suggested fix might just be broadly applicable. 

The precision class of OCCT utilises abs in a nr of cases

https://dev.opencascade.org/doc/refman/html/class_precision.html

PR #1443, opened 2025-03-05
Update relation between frame and transformation

Licini:
Hey folks, I'd like propose to update the relation between `frame` and `transformation` in `SceneObject`, to align better with how they are currently defined in `compas_model` (https://github.com/BlockResearchGroup/compas_model/blob/b8d7a992956abcfe96d41eb11fecd5a064a12674/src/compas_model/elements/element.py#L271), which I found makes more sense and easier to understand.

- We update the calculation of `worldtransformation` to the simple multiplication of all transformations along the scene tree.
- We re-define the `frame` property as a read-only computed property which is derived from `Frame.from_transformation(self.worldtransformation)`, and it can be think of as the reference frame of the local coordinate system of this object after all the transformations applied.

This relation also aligns better with other CAD systems and libraries like Threejs. And it will give a lot of advantage for `compas_viewer` where I will be able to push the transformation and scene tree into GPU and calculate the multiplication there.

### DIFF
--- a/src/compas/scene/sceneobject.py
+++ b/src/compas/scene/sceneobject.py
@@ -13,6 +13,7 @@ import compas.scene  # noqa: F401
 from compas.colors import Color
 from compas.data import Data
 from compas.datastructures import TreeNode
+from compas.geometry import Frame
 from compas.geometry import Transformation
 
 from .context import clear
@@ -36,10 +37,8 @@ class SceneObject(TreeNode):
         The opacity of the object.
     show : bool, optional
         Flag for showing or hiding the object. Default is ``True``.
-    frame : :class:`compas.geometry.Frame`, optional
-        The local frame of the scene object, in relation to its parent frame.
     transformation : :class:`compas.geometry.Transformation`, optional
-        The local transformation of the scene object in relation to its frame.
+        The local transformation of the scene object in relation to its parent object.
     context : str, optional
         The context in which the scene object is created.
     **kwargs : dict
@@ -56,9 +55,9 @@ class SceneObject(TreeNode):
     guids : list[object]
         The GUIDs of the items drawn in the visualization context.
     frame : :class:`compas.geometry.Frame`
-        The local frame of the scene object, in relation to its parent frame.
+        The frame of the local coordinate system of the scene object.
     transformation : :class:`compas.geometry.Transformation`
-        The local transformation of the scene object in relation to its frame.
+        The local transformation of the scene object in relation to its parent object.
     worldtransformation : :class:`compas.geometry.Transformation`
         The transformation of the scene object in world coordinates.
     color : :class:`compas.colors.Color`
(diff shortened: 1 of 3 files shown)


src/compas/scene/sceneobject.py
```diff
@@ -176,21 +169,17 @@ def transformation(self, transformation):
     @property
     def worldtransformation(self):
         # type: () -> compas.geometry.Transformation
-        frame_stack = [self.frame] if self.frame else []
-        parent = self.parent
+        transformations = [self.transformation] if self.transformation else []
```
tomvanmele, line 173: although probably nobody was using this, there is a slight change in behaviour that should be documented:

there can no longer be a independent transformation wrt to the local frame...

src/compas/scene/sceneobject.py
```diff
@@ -176,21 +169,17 @@ def transformation(self, transformation):
...
+        parent: SceneObject = self.parent
```
tomvanmele: i like your enthusiasm, but this annotation can't be introduced yet in  the current version of COMPAS...

tests/compas/scene/test_scene.py
```diff
@@ -83,22 +83,19 @@ def test_sceneobject_auto_context_discovery_no_context(mocker):
...
         scene = Scene()
         sceneobj1 = scene.add(Box())
-        sceneobj1.frame = Frame([1.0, 0.0, 0.0], xaxis=[1.0, 0.0, 0.0], yaxis=[0.0, 1.0, 0.0])
         sceneobj1.transformation = Translation.from_vector([10.0, 0.0, 0.0])
-        assert sceneobj1.worldtransformation == sceneobj1.frame.to_transformation() * sceneobj1.transformation
+        assert sceneobj1.worldtransformation == Translation.from_vector([10.0, 0.0, 0.0])
+        assert sceneobj1.frame == Frame([10.0, 0.0, 0.0], [1.0, 0.0, 0.0], [0.0, 1.0, 0.0])
 
         sceneobj2 = scene.add(Box(), parent=sceneobj1)
-        sceneobj2.frame = Frame([1.0, 1.0, 0.0], xaxis=[1.0, 0.0, 0.0], yaxis=[0.0, 1.0, 0.0])
```
tomvanmele, line 91: by simply removing the frame, you're not reproducing the same result as before...

tomvanmele (commented):
i think we have to add a bit of documentation

src/compas/scene/sceneobject.py
Licini: sure, I added a note in docstring also mentioning now in changelog

src/compas/scene/sceneobject.py
Licini: removed, i got so used to it 😂

tests/compas/scene/test_scene.py
Licini: Yep, but if you see in next few lines, the assert condition was updated according to the new computing scheme, or perhaps I misunderstood something here?

tomvanmele (approved):

PR #1444, opened 2025-03-05
add group

Licini:
Adding a `Group` object to Scene, which is a custom `SceneObject`.  The usages are similar to Three.js:

```python
from compas.scene import Group

scene = Scene()

group = Group(name="My Group")
scene.add(group)

point = Point(0, 0, 0)
group.add(point)

print(scene)
```
```
<Tree with 3 nodes>
└── <TreeNode: ROOT>
    └── <Group: My Group>
        └── <GeometryObject: Point>
```

### DIFF
--- a/src/compas/scene/group.py
+++ b/src/compas/scene/group.py
@@ -4,7 +4,30 @@ from .sceneobject import SceneObject
 
 
 class Group(SceneObject):
-    """A group of scene objects."""
+    """A group of scene objects.
+
+    Parameters
+    ----------
+    name : str, optional
+        The name of the group.
+
+    Examples
+    --------
+    >>> from compas.scene import Scene
+    >>> from compas.scene import Group
+    >>> from compas.geometry import Point
+    >>> scene = Scene()
+    >>> group = Group(name="My Group")
+    >>> scene.add(group)
+    >>> point = Point(0, 0, 0)
+    >>> group.add(point)
+    >>> print(scene)
+    <Tree with 3 nodes>
+    └── <TreeNode: ROOT>
+        └── <Group: My Group>
+            └── <GeometryObject: Point>
+
+    """
 
     def __new__(cls, *args, **kwargs):
         return object.__new__(cls)
--- a/src/compas/scene/group.py
+++ b/src/compas/scene/group.py
@@ -1,4 +1,5 @@
 from compas.data import Data
+
 from .sceneobject import SceneObject
 
 
--- a/src/compas/scene/group.py
+++ b/src/compas/scene/group.py
@@ -1,0 +1,13 @@
+from compas.data import Data
+from .sceneobject import SceneObject
+
+
+class Group(SceneObject):
+    """A group of scene objects."""
+
+    def __new__(cls, *args, **kwargs):
+        return object.__new__(cls)
+
+    def __init__(self, **kwargs):
+        name = kwargs.pop("name", "Group")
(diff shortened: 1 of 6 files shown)


src/compas/scene/group.py
```diff
@@ -0,0 +1,37 @@
...
+    >>> print(scene)
+    <Tree with 3 nodes>
+    └── <TreeNode: ROOT>
+        └── <Group: My Group>
+            └── <GeometryObject: Point>
+
+    """
+
+    def __new__(cls, *args, **kwargs):
+        return object.__new__(cls)
```
Licini: We need to overwrite `____new____` to revert to the default behavior of normal `object`, so I can initialize a instance directly without looking for a registered item.

src/compas/scene/group.py
```diff
@@ -0,0 +1,37 @@
...
+
+    def __init__(self, **kwargs):
+        name = kwargs.pop("name", "Group")
+        super(Group, self).__init__(item=Data(name=name), **kwargs)
```
Licini: Currently `SceneObject` will not accept `None` as item, we can either create this simple `Data` instance just holding the name or we can modify `SceneObject` to accept `item=None`

src/compas/scene/group.py
gonzalocasas: I think this comment will be needed in code, otherwise we will not remember why this is like this.

src/compas/scene/group.py
gonzalocasas: Maybe same here

gonzalocasas (approved):
I think this looks good, it's surprising that such a small change creates this feature, I guess this is just giving it a cleaner name to abusing empty `SceneObject`s

src/compas/scene/group.py
Licini: Done!

src/compas/scene/sceneobject.py
```diff
@@ -101,7 +101,7 @@ def __init__(
         **kwargs  # type: dict
     ):  # fmt: skip
         # type: (...) -> None
-        if not isinstance(item, Data):
+        if not isinstance(item, (Data, type(None))):
```
tomvanmele: should this not be: `if item and not isinstance(item, Data)`?

Licini:
> I think this looks good, it's surprising that such a small change creates this feature, I guess this is just giving it a cleaner name to abusing empty `SceneObject`s

@gonzalocasas  I realized an issue when I'm adding group to the serialization test, where `Data` can not be serialized without `__ data __` implemented, which means we still have to extend `Data` even it is not really doing much here..., in this case allowing `SceneObject` to take `None` item is perhaps a better route. It can also reduce footprint of serialized `Scene` especially when we have a lot of groups. Could you take a look again let me know what you think?

tomvanmele (approved):
LGTM just a minor remark

src/compas/scene/sceneobject.py
Licini: Indeed, looks more elegant, updated!

tomvanmele:
Also, i was wondering if it would be cleaner to use a `factory` classmethod on `SceneObject`, instead of overwriting `__new__`

Licini:
> Also, i was wondering if it would be cleaner to use a `factory` classmethod on `SceneObject`, instead of overwriting `__new__`

Sure, Factory pattern could be more readable than __ new __, I can experiment the idea a bit in another PR

PR #1446, opened 2025-03-14
Add make_solid method to RhinoBrep to cap planar holes

papachap:
Adds `make_solid()` method to `RhinoBrep` for converting a brep to a solid if it is a strip in the Rhino backend.

### Checklist

- [X] I added a line to the `CHANGELOG.md` file in the `Unreleased` section under the most fitting heading (e.g. `Added`, `Changed`, `Removed`).
- [X] I ran all tests on my computer and it's all green (i.e. `invoke test`).
- [X] I ran lint on my computer and there are no errors (i.e. `invoke lint`).
- [X] I have added necessary documentation (if appropriate)

### DIFF
--- a/src/compas_rhino/geometry/brep/brep.py
+++ b/src/compas_rhino/geometry/brep/brep.py
@@ -684,9 +684,9 @@ class RhinoBrep(Brep):
         None
 
         """
-        if not self._brep.IsSolid:
-            capped_brep = self._brep.CapPlanarHoles(TOL.absolute)
-            if capped_brep:
-                self._brep = capped_brep
+        if not self.is_solid:
+            result = self._brep.CapPlanarHoles(TOL.absolute)
+            if result:
+                self._brep = result
             else:
                 raise BrepInvalidError("Failed to convert Brep to solid")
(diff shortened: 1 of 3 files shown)


src/compas_rhino/geometry/brep/brep.py
```diff
@@ -674,3 +675,18 @@ def flip(self):
 
         """
         self._brep.Flip()
+
+    def make_solid(self):
```
chenkasirer: I'm wondering if we shouldn't call this just `cap_holes`. 
Wasn't able to fully understand but after having a look at the [OCC implementation](https://github.com/compas-dev/compas_occ/blob/e5f3349d7587f267e0e8524da0ed7c3c7667f049/src/compas_occ/brep/brep.py#L1650C5-L1650C17) of `make_solid()` I'm not convinced these do the same thing.

tomvanmele: `make_solid` in `compas_occ` indeed does something else...

obucklin: I came to say the same thing. Is there another occ function that caps planar holes?

tomvanmele: there currently isn't a method that closes a hole in an existing brep, but this could be added...

papachap: @chenkasirer @tomvanmele @obucklin I now see my misconception, so thanks for pointing it out. I’ve renamed the method to `cap_planar_holes()` in `compas_rhino.RhinoBrep` and added it to `compas.geometry.Brep`.

chenkasirer (approved):
LGTM!

Issue #1451, opened 2025-03-25
`Brep.from_curves` implementation for `RhinoBrep`


obucklin:
As @chenkasirer mentioned in #1447 , we should start catching up on Brep implementations in compas_rhino. 

I will begin by implementing `Brep.from_curves`

PR #1452, opened 2025-03-25
RhinoBrep.from_curves

obucklin:
solves #1451 
This adds implementation for `Brep.from_curves` for rhino. 

### Checklist


obucklin:
@chenkasirer do I ping someone for a review here? should I just be patient?

src/compas_rhino/geometry/brep/brep.py
```diff
@@ -302,6 +302,39 @@ def from_cylinder(cls, cylinder):
...
+        """Create a RhinoBreps from a list of planar face boundary curves.
+
+        Parameters
+        ----------
+        curves : : list of :class:`~compas.geometry.Curve` or :class:`~compas.geometry.Polyline`
+            The planar curves that make up the face borders of brep faces.
+
+        Returns
+        -------
+        list of :class:`~compas_rhino.geometry.RhinoBrep`
```
chenkasirer, line 316: I get that `CreatePlanarBreps` returns a list of Breps but I imagined this function creating a solid from the outline of multiple faces of the (eventually) same brep. I'm also wondering because of the original docstring specifies a single Brep as the return type. There's currently no OCC implementation so nothing to compare to, but I'm simply not sure what's the intention here. 

Could you maybe provide a use-case for this method that illustrates what we should be able to make with it?

src/compas_rhino/geometry/brep/brep.py
```diff
@@ -302,6 +302,39 @@ def from_cylinder(cls, cylinder):
...
+        """
+        if not isinstance(curves, list):
+            curves = [curves]
+        faces = []
+        for curve in curves:
+            if isinstance(curve, Polyline):
+                rhino_curve = polyline_to_rhino_curve(curve)
+            else:
+                rhino_curve = curve_to_rhino(curve)
+            face = Rhino.Geometry.Brep.CreatePlanarBreps(rhino_curve)
```
chenkasirer: as per the [documentation](https://developer.rhino3d.com/api/rhinocommon/rhino.geometry.brep/createplanarbreps#(curve,double)) this version of this method is deprecated. use the version that takes tolerance instead.

src/compas_rhino/geometry/brep/brep.py
```diff
@@ -302,6 +302,39 @@ def from_cylinder(cls, cylinder):
...
+            if face is None:
+                raise BrepError("Failed to create face from curve: {} ".format(curve))
+            if isinstance(face, list) and len(face) > 1:
```
chenkasirer: if `face` is not `None` it's safe to assume `face` is a list so this check is superfluous.

src/compas_rhino/geometry/brep/brep.py
obucklin: > I get that `CreatePlanarBreps` returns a list of Breps but I imagined this function creating a solid from the outline of multiple faces of the (eventually) same brep. 

this is the general use case. however if not all of the faces are joinable, how do you decide which ones to get rid of?

>I'm also wondering because of the original docstring specifies a single Brep as the return type. There's currently no OCC implementation so nothing to compare to, but I'm simply not sure what's the intention here.

I can either update the docstring to return a list of Breps, or I can return None if the faces don't join into one single Brep. I would typically go for the former and have the user decide if they only accept one single Brep or if they use multiple Breps. 

> Could you maybe provide a use-case for this method that illustrates what we should be able to make with it? 

I was developing this to build a brep of a timber plate in the way that the `compas_model.Plate._compute_geometry()` does it, that is by building faces from polyline points. That type builds a mesh from vertices and I want to build a Brep, obviously. sample code looks like this:

```
        for i in range(len(self.outline_a)-1):
            a = self.outline_a.points[i]
            b = self.outline_a.points[i+1]
            c = self.outline_b.points[i+1]
            d = self.outline_b.points[i]
            curves.append(NurbsCurve.from_points([a, b, c, d, a], degree=1))
        plate_geo = Brep.from_curves(curves)
        if len(plate_geo) != 1:
            raise ValueError("The curves could not be joined into a single Brep.")  
        else:   
            plate_geo = plate_geo[0] 
```

chenkasirer: no, fair enough!

src/compas_rhino/geometry/brep/brep.py
```diff
@@ -302,6 +302,39 @@ def from_cylinder(cls, cylinder):
         rhino_cylinder = cylinder_to_rhino(cylinder)
         return cls.from_native(rhino_cylinder.ToBrep(True, True))
 
+    @classmethod
+    def from_curves(cls, curves):
+        """Create a RhinoBreps from a list of planar face boundary curves.
+
+        Parameters
+        ----------
+        curves : : list of :class:`~compas.geometry.Curve` or :class:`~compas.geometry.Polyline`
```
chenkasirer: little typo
```suggestion
        curves : list of :class:`~compas.geometry.Curve` or :class:`~compas.geometry.Polyline`
```

chenkasirer (approved):
Nice! LGTM!

Issue #1457, opened 2025-04-07
Character in `compas/scene/group.py` causes failure in Rhino 7


obucklin:
Something in the docstring in  `compas/scene/group.py` causes an error, causing compas.scene to not work. 

The Draw GH component gives the following error when adding to canvas:
```
Runtime error (SyntaxErrorException): Non-ASCII character '\xe2' in file C:\Users\xxx\AppData\Roaming\McNeel\Rhinoceros\7.0\scripts\compas\scene\group.py on line 24, but no encoding declared; see http://www.python.org/peps/pep-0263.html for details

Traceback:
  line 17, in <module>, "C:\Users\xxx\AppData\Roaming\McNeel\Rhinoceros\7.0\scripts\compas\scene\__init__.py"
  line 7, in script

```

**To Reproduce**
1. Open Rhino 7 and Grasshopper
2. Sdrop compas `Draw` component on canvas
3. component in error state
4. copy error
5. paste error here.

**Desktop (please complete the following information):**
 - OS: Windows 11
 - IronPython 
 
**Additional context**
Thanks!!

Licini:
thx! @obucklin this is now fixed in main branch, we will make a release soon

PR #1458, opened 2025-04-09
Group

Licini:
* Fixed a bug when printing an empty `Tree`.
* Fixed a bug in `Group` for IronPython where the decoding declaration was missing. #1457 
* Fixed a bug where a `Group` without name could not be added to the scene.


chenkasirer (approved):
LGTM!

tomvanmele (approved):

PR #1460, opened 2025-05-06
serialize and deserilize mro

Licini:
This is to add the ability to deserializing a `datastructure` implementation like `Mesh`, `Graph`, `Cellnetwork` etc. to its closest available parent class. 

For example we want to be able to serialize a `FormDiagram` object using `compas_tna` into a json. And open same json in another environment without `compas_tna` installed, while loading the same object as its closest available class `Mesh`.

An example behaviour:
```python
from compas import json_dumps
from compas import json_loads

# Some custom class that inherits from Mesh but is not available when deserializing
class MyCustomMesh(Mesh):
    pass

mesh = MyCustomMesh()
serialized = json_dumps(mesh)

# delete the access to our custom class
del MyCustomMesh

# fallback to Mesh if MyCustomMesh is not available
loaded = json_loads(serialized)
assert loaded.__class__ == Mesh
```

This is done through inserting an `mro` (method resolving order) field into the json dump state, so the decoder can later pickup and use as a backup if `dtype` is not directly found

### DIFF
--- a/src/compas/data/data.py
+++ b/src/compas/data/data.py
@@ -76,7 +76,7 @@ class Data(object):
         return "{}/{}".format(".".join(self.__class__.__module__.split(".")[:2]), self.__class__.__name__)
 
     @classmethod
-    def __cls_dtype__(cls):
+    def __clstype__(cls):
         return "{}/{}".format(".".join(cls.__module__.split(".")[:2]), cls.__name__)
 
     @property
--- a/src/compas/data/data.py
+++ b/src/compas/data/data.py
@@ -75,6 +75,10 @@ class Data(object):
     def __dtype__(self):
         return "{}/{}".format(".".join(self.__class__.__module__.split(".")[:2]), self.__class__.__name__)
 
+    @classmethod
+    def __cls_dtype__(cls):
+        return "{}/{}".format(".".join(cls.__module__.split(".")[:2]), cls.__name__)
+
     @property
     def __data__(self):
         raise NotImplementedError
--- a/src/compas/data/encoders.py
+++ b/src/compas/data/encoders.py
@@ -40,7 +40,7 @@ except (ImportError, SyntaxError):
     numpy_support = False
 
 
-def cls_from_dtype(dtype):  # type: (...) -> Type[Data]
+def cls_from_dtype(dtype, mro=None):  # type: (...) -> Type[Data]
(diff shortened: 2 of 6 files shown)


src/compas/data/data.py
```diff
@@ -75,6 +75,10 @@ def __init__(self, name=None):
     def __dtype__(self):
         return "{}/{}".format(".".join(self.__class__.__module__.split(".")[:2]), self.__class__.__name__)
 
+    @classmethod
+    def __cls_dtype__(cls):
```
tomvanmele: `__clstype__` or `__ctype__` perhaps?

src/compas/data/encoders.py
```diff
@@ -40,14 +40,16 @@
     numpy_support = False
 
 
-def cls_from_dtype(dtype):  # type: (...) -> Type[Data]
+def cls_from_dtype(dtype, mro=None):  # type: (...) -> Type[Data]
```
tomvanmele: i would not call this "mro". method resolution order refers to the mechanism that Python uses to determine which method should be called in a multiple inheritance scenario. but that is not what we do here...

src/compas/datastructures/datastructure.py
```diff
@@ -21,6 +21,41 @@ def __init__(self, attributes=None, name=None):
         self._aabb = None
         self._obb = None
 
+    def __get_mro__(self):
```
tomvanmele: `__clstree__` or `__inheritance__` or something like that?

src/compas/data/data.py
Licini: ok clstype then, as ctype is a exiting function (C compatible data types)

src/compas/data/encoders.py
Licini: indeed, we use "inheritance" then

src/compas/datastructures/datastructure.py
Licini: ok, `__inheritance__`  and as property

Licini:
@tomvanmele how does this look now?

tomvanmele (approved):

gonzalocasas:
Very cool change! Could you please revise the data serialization doc page, and add this information? https://compas.dev/compas/latest/userguide/advanced.serialisation.html

gonzalocasas (approved):

PR #1461, opened 2025-05-14
Scene Data Store

Licini:
Hey guys a rather large PR, there are several components to this, let me explain:
## 1. Scene data store
Previously we inherent `Scene` from `Tree` and `SceneObject` from `TreeNode`. It was nice but the serialization was not very straightforward, because we don't want to serialize `Data` item directly under `SceneObject` since they might be reused. So we had to customize the serialization and deserialization to only store `guid` of item under `SceneObject`, then group all the distinct items as a list under scene.  

This PR basically makes the idea more explicit by reflecting this mechanism directly as the attributes of `Scene`. Under `Scene`, we have 3 main aspects: `datastore`, `objectstore` and `tree`. The `datastore` is a bucket of unique `Data` items, the `objectstore` is a bucket of `SceneObjects`, both keyed by their `guids`. And `tree` stores hierarchy of the scene objects, but nothing more. 

For `SceneObject`, instead maintaining a live link to `Data` item, we only actually store `guid` of that item and use the datastore of the scene to retrieve it (with minimal overhead). 

The benefit is a much more transparent serialization and deserialization process: now both `Scene` and `SceneObject` can be json dumped/loaded without any customization using our default encoder/decoder. The structure of two classes directly reflect the serialized data with almost no change.

This also makes it easier for further extension of `Scene` for example if you want to use a datastore that contains very large geometry but with lazy loading, you can pass in a custom `datastore` yourself when creating the scene, using dict-like object which can be queried with a `guid`. 

For example a minimal serialized scene can look like this now:
```
{
    "data": {
        "attributes": {},
        "datastore": {
            "07b131a8-6edc-407a-8944-174a274fec5b": {
                "data": {
                    # ...
                },
                "dtype": "compas.geometry/Box",
                "guid": "07b131a8-6edc-407a-8944-174a274fec5b"
            }
        },
        "name": "Scene",
        "objectstore": {
            "22fd4e9c-66fe-481a-9392-c07c7fea2522": {
                "data": {
                    "item": "07b131a8-6edc-407a-8944-174a274fec5b",
                    # ...
                },
                "dtype": "compas.scene/GeometryObject",
                "guid": "22fd4e9c-66fe-481a-9392-c07c7fea2522",
                "name": "box1"
            },
            "bea576ac-641e-400b-9083-0ed018003af0": {
                "data": {
                    "item": "07b131a8-6edc-407a-8944-174a274fec5b",
                    # ...
                },
                "dtype": "compas.scene/GeometryObject",
                "guid": "bea576ac-641e-400b-9083-0ed018003af0",
                "name": "box2"
            }
        },
        "tree": {
            "data": {
                "attributes": {},
                "root": {
                    "children": [
                        {
                            "children": [
                                {
                                    "name": "bea576ac-641e-400b-9083-0ed018003af0"
                                }
                            ],
                            "name": "22fd4e9c-66fe-481a-9392-c07c7fea2522"
                        }
                    ],
                    "name": "Scene"
                }
            },
            "dtype": "compas.datastructures/Tree",
            "guid": "ef7305fd-cabc-45c2-a66e-e3cdc367ec25",
            "inheritance": []
        }
    },
    "dtype": "compas.scene/Scene",
    "guid": "ac6b3bc4-bda2-4c5c-9d0e-8496d3590cb5",
    "inheritance": []
}
```

## 2. SceneObjectFactory
We have been using `__new__` for automatically choosing appropriate `SceneObject` class.  It has been working well most of time. But it becomes annoying if one wants to initialize a custom `SceneObject` class like `Group`, where I just want a straightforward initiation without triggering auto class detection. @tomvanmele has been advocating for more explicit Factory method instead `__new__`, to make the constructor behavior more predictable and more type-hint friendly. Which I quite agree. 

This does not change the external APIs, users still use `Scene.add` the same way, which internally calls the factory method now instead of the class constructor.

## 3. Others
I added more tests on `Scene` to make sure it behaves correctly. In summary, although there are a big internal re-organization in `Scene` and `SceneObject`, the user-facing APIs aren't changed, there also shouldn't be any changes needed for all the extended `SceneObjects`. I will drop some in-line comments below


src/compas/scene/scene.py
```diff
@@ -108,19 +113,49 @@ def add(self, item, parent=None, **kwargs):
...
         """
 
-        parent = parent or self.root
+        if "context" in kwargs:
+            if kwargs["context"] != self.context:
+                raise Exception("Object context should be the same as scene context: {} != {}".format(kwargs["context"], self.context))
+            del kwargs["context"]  # otherwist the SceneObject receives "context" twice, which results in an error
+
+        # Create a corresponding new scene object
+        sceneobject = SceneObjectFactory.create(item=item, context=self.context, scene=self, **kwargs)
```
tomvanmele: since this class has only 1 static method, perhaps if might as well just be a function? using a class seems overkill...

src/compas/scene/scene.py
```diff
@@ -108,19 +113,49 @@ def add(self, item, parent=None, **kwargs):
...
         else:
-            if "context" in kwargs:
-                if kwargs["context"] != self.context:
-                    raise Exception("Object context should be the same as scene context: {} != {}".format(kwargs["context"], self.context))
-                del kwargs["context"]  # otherwist the SceneObject receives "context" twice, which results in an error
-            sceneobject = SceneObject(item=item, context=self.context, **kwargs)  # type: ignore
-        super(Scene, self).add(sceneobject, parent=parent)
+            if not isinstance(parent, SceneObject):
+                raise ValueError("Parent is not a SceneObject.", parent)
+            parent_node = self.tree.get_node_by_name(parent.guid)
```
tomvanmele: should we not add a method for finding a node "by guid" instead of misusing the `get_node_by_name` method?

getting something by name to me implies that there might be multiple objects that might be found/returned.

tomvanmele: perhaps `tree.get(guid)` exists as default, whereas `tree.get_by_...()` provides alternatives

src/compas/scene/sceneobject.py
```diff
@@ -22,7 +21,50 @@
...
-class SceneObject(TreeNode):
+class SceneObjectFactory:
+    """Factory class for creating appropriate SceneObject instances based on input item type.
+
+    This factory encapsulates the logic for selecting the right SceneObject subclass
+    for a given data item, making the creation process more explicit and easier to understand.
+    """
+
+    @staticmethod
+    def create(item=None, scene=None, **kwargs):
```
tomvanmele: as mentioned, perhaps it is overkill to use a class for this, since there is only 1 static method...

src/compas/scene/sceneobject.py
```diff
@@ -22,7 +21,50 @@
...
+            item._scene = scene
+            return item
+
+        sceneobject_cls = get_sceneobject_cls(item, **kwargs)
+
+        # Create and return an instance of the appropriate scene object class
+        return sceneobject_cls(item=item, scene=scene, **kwargs)
+
+
+class SceneObject(Data):
```
tomvanmele, line 63: at some point, we should perhaps introduce `Serializable` as base class before `Data`, because i find it odd that the scene objects are also data, while they are meant to encapsulate data types.

only reason to make them `Data` is to add the serialisation behaviour...

src/compas/scene/sceneobject.py
```diff
@@ -123,29 +169,69 @@ def __init__(
...
         return {
-            "item": str(self.item.guid),
-            "settings": self.settings,
-            "children": [child.__data__ for child in self.children],
+            "item": self._item,
+            "name": self.name,
+            "color": self.color,
+            "opacity": self.opacity,
+            "show": self.show,
+            "transformation": self.transformation,
```
tomvanmele, line 173: cool, this makes much more sense than having a separate settings dict

src/compas/scene/sceneobject.py
```diff
@@ -199,55 +285,32 @@ def contrastcolor(self, color):
...
-        else:
-            if "context" in kwargs:
-                if kwargs["context"] != self.context:
-                    raise Exception("Child context should be the same as parent context: {} != {}".format(kwargs["context"], self.context))
-                del kwargs["context"]  # otherwist the SceneObject receives "context" twice, which results in an error
-            sceneobject = SceneObject(item=item, context=self.context, **kwargs)  # type: ignore
-
-        super(SceneObject, self).add(sceneobject)
-        return sceneobject
+        # type: (compas.data.Data, dict) -> compas.scene.SceneObject
```
tomvanmele: type info should come before the docstring

tomvanmele: i think

src/compas/scene/scene.py
Licini: yep that make sense

src/compas/scene/scene.py
Licini: I understand, using `name` can sound strange indeed. But if we use something like `get_by_guid`, it is also not entirely right, because this `guid` here we use for querying is the `guid` of a `SceneObject`, not the `guid` of the `TreeNode` itself, they are still two different objects of distinct classes, It doesn't feel right for them to share the same `guid`.

Licini: The most appropriate naming in my opinion would be `sceneobject_guid`, to achieve that we can either: 
1. store it in the built-in `attributes` dict of `Data` class. The downside would be adding an extra "attribute" dict layer in every node in serialization. 
2. Add a custom `SceneNode` class like what we did before, we can still make the serialization compact, with the price of adding another class that does only this.
3. Other better ideas? ...

src/compas/scene/sceneobject.py
Licini: sure done!

src/compas/scene/sceneobject.py
Licini: yep indeed

src/compas/scene/scene.py
tomvanmele: add a wrapper method to the scene itself, instead of using the method of the tree directly?

src/compas/scene/sceneobject.py
Licini: This is tricky... `Serializable` is definitely more precise here, meanwhile I think the term `Data` is too generic that can be interpreted in many ways. 
Some "external opinions" here: https://chatgpt.com/share/6825ac7d-0408-800f-bd76-fd084568c57f

src/compas/scene/scene.py
Licini: ok, sounds like a pretty good mid point! will do this way!

src/compas/scene/context.py
Licini, line 1: Consolidating `_get_sceneobject_cls` and `get_sceneobject_cls` into one function. Also cleaned up the kwargs. I removed `sceneobject_type` parameter because now you can call constructor directly with `__new__` removed.

src/compas/scene/group.py
```diff
@@ -27,16 +27,3 @@ class Group(SceneObject):
...
-        return object.__new__(cls)
-
-    @property
-    def __data__(self):
-        # type: () -> dict
-        data = {
-            "settings": self.settings,
-            "children": [child.__data__ for child in self.children],
-        }
-        return data
```
Licini, line 42: These customizations are no longer needed anymore, thanks to the more straight forward serialization/deserialization

src/compas/scene/scene.py
```diff
@@ -43,47 +44,51 @@ class Scene(Tree):
...
-                if "item" in child_node:
-                    guid = child_node["item"]
-                    sceneobject = parent.add(items[guid], **settings)
-                else:
-                    sceneobject = parent.add(Group(**settings))
-                add(child_node, sceneobject, items)
-
-        add(data["root"], scene, items)
-
-        return scene
```
Licini, line 71: We directly serialize `datastore`, `objectstore` and `tree` here. And during deserialization we can read them as they are without using a custom `__from__data__` method.

src/compas/scene/scene.py
```diff
@@ -43,47 +44,51 @@ class Scene(Tree):
...
-
-    def __init__(self, name="Scene", context=None):
-        # type: (str, str | None) -> None
-        super(Scene, self).__init__(name=name)
-        super(Scene, self).add(TreeNode(name="ROOT"))
+    def __init__(self, context=None, datastore=None, objectstore=None, tree=None, **kwargs):
```
Licini, line 55: Users can passing their own customs stores and tree, if they know what they are doing.

src/compas/scene/sceneobject.py
```diff
@@ -123,29 +165,69 @@ def __init__(
...
     def scene(self):
         # type: () -> compas.scene.Scene | None
-        return self.tree
+        return self._scene
 
     @property
     def item(self):
         # type: () -> compas.data.Data
-        return self._item
+        return self.scene.datastore[self._item]
```
Licini, line 188: `item` now becomes a getter that retrieve the live object from datastore, what we actually store as `_item` is the `guid`

src/compas/scene/sceneobject.py
```diff
@@ -123,29 +165,69 @@ def __init__(
...
+
+    @property
+    def parentnode(self):
+        # type: () -> compas.datastructures.Node | None
+        return self.node.parent
+
+    @property
+    def childnodes(self):
+        # type: () -> list[compas.datastructures.Node]
+        return self.node.children
```
Licini, line 220: "shortcut" properties that allows us to still work with `SceneObject` like a `TreeNode`

src/compas/scene/scene.py
```diff
@@ -43,47 +44,51 @@ class Scene(Tree):
...
     @property
     def objects(self):
         # type: () -> list[SceneObject]
-        return [node for node in self.nodes if not node.is_root]  # type: ignore
+        return list(self.objectstore.values())
 
     @property
     def context_objects(self):
-        # type: () -> list
+        # type: () -> list[SceneObject]
```
tomvanmele: these are not scene objects but the guids of the objects created as representation in the visualisation context

src/compas/scene/scene.py
```diff
@@ -230,10 +264,10 @@ def find_by_name(self, name):
         :class:`SceneObject`
 
         """
-        return self.get_node_by_name(name=name)
+        return next((obj for obj in self.objects if obj.name == name), None)
```
tomvanmele, line 267: return type can be None

src/compas/scene/sceneobject.py
```diff
@@ -22,7 +21,46 @@
 from .descriptors.protocol import DescriptorProtocol
 
 
-class SceneObject(TreeNode):
+def SceneObjectFactory(item=None, scene=None, context=None, **kwargs):
```
tomvanmele: function name breaks naming conventions. why not just snake case?

tomvanmele: type hint?

src/compas/scene/sceneobject.py
```diff
@@ -84,33 +122,37 @@ class SceneObject(TreeNode):
...
-        return super(SceneObject, cls).__new__(sceneobject_cls)
-
     def __init__(
         self,
         item=None,  # type: compas.data.Data | None
         name=None,  # type: str | None
         color=None,  # type: compas.colors.Color | None
         opacity=1.0,  # type: float
         show=True,  # type: bool
-        frame=None,  # type: compas.geometry.Frame | None
```
tomvanmele, line 98: perhaps this should be mentioned in the log...

src/compas/scene/sceneobject.py
```diff
@@ -123,29 +165,69 @@ def __init__(
...
     @property
     def item(self):
         # type: () -> compas.data.Data
-        return self._item
+        return self.scene.datastore[self._item]
+
+    @property
+    def node(self):
+        # type: () -> compas.datastructures.TreeNode
+        return self.scene.get_sceneobject_node(self)
```
tomvanmele: should we not cache this?

tomvanmele:
@Licini shall we try to finish this?

src/compas/scene/scene.py
Licini: ah right my mistake

src/compas/scene/sceneobject.py
Licini: I forget to switch pattern from class naming

src/compas_ghpython/components/Compas_ToRhinoGeometry/code.py
```diff
@@ -4,12 +4,12 @@
...
+from compas.scene import sceneobject_factory
 
 
 class CompasToRhinoGeometry(component):
     def RunScript(self, cg):
         if not cg:
             return None
 
-        return SceneObject(item=cg).draw()
+        return sceneobject_factory(item=cg).draw()
```
Licini, line 15: @gonzalocasas @chenkasirer Please note here, with the factory pattern, you need call factory function in GH components instead of constructor

Licini:
> @Licini shall we try to finish this?

Done with the requested changes

src/compas_ghpython/components/Compas_ToRhinoGeometry/code.py
gonzalocasas: This is a breaking change that affects a lot of external code. It needs a major release (compas 3!). Also, I don't like the method name `sceneobject_factory`, it feels old, could we add the factory method as a class/static method of `SceneObject`, like `SceneObject.create()`? the word factory on its own sounds too much java-like to my taste :P

Licini: > This is a breaking change that affects a lot of external code. It needs a major release (compas 3!). Also, I don't like the method name `sceneobject_factory`, it feels old, could we add the factory method as a class/static method of `SceneObject`, like `SceneObject.create()`? the word factory on its own sounds too much java-like to my taste :P

Ok 🤣 I had impression that these are the only places SceneOjbect constructor are called. But sure `SceneObject.create()` is a good idea to me too. Regarding the full PR, I have been having more discussions @tomvanmele  on potential issues and corresponding reworks for data store, so I will close this one for now and re-submit another one that only concerns the serialization part. We can deal with `__new__` in the future in compas 3

PR #1469, opened 2025-06-02
Update devday acknowledgements

domrice:
On the T-shirt, it says `compas.devday(2025, open=True)`. The current implementation does not use the boolean, which is anyway called `is_open`, not `open`.

### Checklist

### DIFF
--- a/src/compas/__init__.py
+++ b/src/compas/__init__.py
@@ -275,7 +275,7 @@ def get_bunny(localstorage=None):
     return bunny
 
 
-def devday(year, *, open):
+def devday(year, open=True):
     import zlib
     import base64
 
--- a/src/compas/__init__.py
+++ b/src/compas/__init__.py
@@ -275,9 +275,10 @@ def get_bunny(localstorage=None):
     return bunny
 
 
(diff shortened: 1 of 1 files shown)


src/compas/__init__.py
```diff
@@ -275,9 +275,10 @@ def safe_extract(tar, path=".", members=None, numeric_owner=False):
     return bunny
 
 
-def devday(year, is_open):
+def devday(year, *, open):
```
tomvanmele: should this not have a default argument of `True`?

gonzalocasas: yep, let's change it to 
```suggestion
def devday(year, open=True):
```

gonzalocasas (changes requested):
Thanks for catching this @dominikreisach !! If you agree the small suggestion posted above, we merge it

src/compas/__init__.py
domrice: Sure, let's do it this way!

tomvanmele (approved):

gonzalocasas (approved):

PR #1470, opened 2025-06-02
try data store again

Licini:
Another attempt for data store only update. Please comments below...

### DIFF
--- a/src/compas/scene/scene.py
+++ b/src/compas/scene/scene.py
@@ -43,38 +43,36 @@ class Scene(Tree):
     @property
     def __data__(self):
         # type: () -> dict
-        items = {str(object.item.guid): object.item for object in self.objects if object.item is not None}
         return {
             "name": self.name,
             "root": self.root.__data__,  # type: ignore
-            "items": list(items.values()),
+            "datastore": self.datastore,
         }
 
     @classmethod
     def __from_data__(cls, data):
         # type: (dict) -> Scene
-        scene = cls(data["name"])
-        items = {str(item.guid): item for item in data["items"]}
-
-        def add(node, parent, items):
+        scene = cls(data["name"], datastore=data["datastore"])
+        def add(node, parent):
             for child_node in node.get("children", []):
                 settings = child_node["settings"]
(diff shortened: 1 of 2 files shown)


src/compas/scene/scene.py
```diff
@@ -43,38 +43,36 @@ class Scene(Tree):
...
         }
 
     @classmethod
     def __from_data__(cls, data):
         # type: (dict) -> Scene
-        scene = cls(data["name"])
-        items = {str(item.guid): item for item in data["items"]}
-
-        def add(node, parent, items):
+        scene = cls(data["name"], datastore=data["datastore"])
```
Licini, line 55: in this version, we still allow direct serialization and deserialization of `datastore` attribute

src/compas/scene/scene.py
```diff
@@ -108,6 +106,9 @@ def add(self, item, parent=None, **kwargs):
             The scene object associated with the item.
         """
 
+        if isinstance(item, Group):
+            raise Exception("Use Scene.add_group to add a group to the scene.")
+
```
Licini, line 111: Letting users to interact with a `Group` instance before properly attached to `Scene` is fragile, this is added to prevent that and ask user to use `scene.add_group` instead, so the links are property established.

src/compas/scene/scene.py
```diff
@@ -118,8 +119,18 @@ def add(self, item, parent=None, **kwargs):
...
         super(Scene, self).add(sceneobject, parent=parent)
+
         return sceneobject
+    
+    def add_group(self, name=None, parent=None, **kwargs):
+        # type: (str, SceneObject | TreeNode | None, dict) -> Group
+        parent = parent or self.root
+        group = Group(name=name, **kwargs)
+        super(Scene, self).add(group, parent=parent)
+        return group
```
Licini, line 133: Another reason we must have this function is that: `Group` class is not registered with an `item` and can not automatically reconstructed with correct extension class in a different context like `Viewer`. For example in Viewer we must overwrite this function to use `compas_viewer.scene.Group` instead of the original one

PR #1472, opened 2025-06-04
Small fixes

tomvanmele:
* fixed context detection (default should be None and not Viewer)
* fixed discretisation of polyhedron into edges (when data comes from numpy)
* fixed graph construction from pointcloud which used degree parameter in nonsensical way
* added isocurve_u and isocurve_v for spherical and cylindrical surfaces

### DIFF
--- a/src/compas/datastructures/graph/graph.py
+++ b/src/compas/datastructures/graph/graph.py
@@ -166,8 +166,8 @@ class Graph(Datastructure):
         default_node_attributes=None,
         default_edge_attributes=None,
         name=None,
-        **kwargs,
-    ):
+        **kwargs
+    ):  # fmt: skip
         super(Graph, self).__init__(kwargs, name=name)
         self._max_node = -1
         self.node = {}
--- a/src/compas/datastructures/graph/graph.py
+++ b/src/compas/datastructures/graph/graph.py
@@ -5,6 +5,8 @@ from __future__ import print_function
 from ast import literal_eval
 from itertools import combinations
 from random import sample
+from random import choice
+from random import shuffle
 
 import compas
 
@@ -132,7 +134,10 @@ class Graph(Datastructure):
 
     def __before_json_dump__(self, data):
         data["node"] = {repr(key): attr for key, attr in data["node"].items()}
-        data["edge"] = {repr(u): {repr(v): attr for v, attr in nbrs.items()} for u, nbrs in data["edge"].items()}
+        data["edge"] = {
+            repr(u): {repr(v): attr for v, attr in nbrs.items()}
+            for u, nbrs in data["edge"].items()
+        }
         return data
 
     def __after_json_load__(self, data):
@@ -140,7 +145,10 @@ class Graph(Datastructure):
         nodes = data["node"] or {}
         edges = data["edge"] or {}
         data["node"] = {l_e(node): attr for node, attr in nodes.items()}
-        data["edge"] = {l_e(u): {l_e(v): attr for v, attr in nbrs.items()} for u, nbrs in edges.items()}
+        data["edge"] = {
+            l_e(u): {l_e(v): attr for v, attr in nbrs.items()}
+            for u, nbrs in edges.items()
+        }
         return data
 
     @classmethod
@@ -159,7 +167,13 @@ class Graph(Datastructure):
         graph._max_node = data.get("max_node", graph._max_node)
         return graph
 
-    def __init__(self, default_node_attributes=None, default_edge_attributes=None, name=None, **kwargs):
+    def __init__(
+        self,
+        default_node_attributes=None,
+        default_edge_attributes=None,
+        name=None,
+        **kwargs,
+    ):
         super(Graph, self).__init__(kwargs, name=name)
         self._max_node = -1
         self.node = {}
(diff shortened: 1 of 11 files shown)


src/compas/geometry/surfaces/cylindrical.py
```diff
@@ -162,6 +163,39 @@ def from_three_points(cls, a, b, c):
...
+        Returns
+        -------
+        :class:`compas.geometry.Circle`
+
+        """
+        point = self.center + self.frame.zaxis * v
+        xaxis = self.frame.xaxis
+        yaxis = self.frame.yaxis
+        frame = Frame(point, xaxis, yaxis)
+        return Circle(radius=self.radius, frame=frame)
```
gonzalocasas, line 197: I understand the logical reason for returning `Circle` and `Line`, but wouldn't make more sense to return `Curve` instances always? That way both `_u` and `_v` would be symmetrical and also this API would be consistent with the one of `Surface`

tests/compas/geometry/test_polyhedron.py
```diff
@@ -12,6 +12,6 @@ def test_polyhedron():
     assert polyhedron.faces == faces
     assert polyhedron.name == name
     assert polyhedron.points == vertices
-    assert polyhedron.lines == [(a, b) for a, b in pairwise(vertices + vertices[:1])]
+    assert polyhedron.lines == [(a, b) for a, b in pairwise(vertices[-1:] + vertices)]
     assert polyhedron.points[0] == vertices[0]
     assert polyhedron.points[-1] != polyhedron.points[0]
```
gonzalocasas, line 19: Could you add a test for `edges()` also since the implementation has changed now?

src/compas/scene/context.py
```diff
@@ -112,9 +112,9 @@ def detect_current_context():
         return "Rhino"
     if compas.is_blender():
         return "Blender"
-    other_contexts = [v for v in ITEM_SCENEOBJECT.keys()]
-    if other_contexts:
-        return other_contexts[0]
+    # other_contexts = [v for v in ITEM_SCENEOBJECT.keys()]
+    # if other_contexts:
+    #     return other_contexts[0]
```
gonzalocasas, line 117: This looks dangerous to commit, isn't that going to break stuff, like the viewer?

src/compas/geometry/surfaces/spherical.py
```diff
@@ -191,6 +191,44 @@ def from_points(cls, points):
...
+        :class:`compas.geometry.Circle`
+
+        """
+        x = self.point_at(u=0, v=v)
+        y = self.point_at(u=0.25, v=v)
+        origin = self.center + self.frame.zaxis * (x - self.center).dot(self.frame.zaxis)
+        xaxis = x - origin
+        yaxis = y - origin
+        frame = Frame(origin, xaxis, yaxis)
+        return Circle(radius=xaxis.length, frame=frame)
```
gonzalocasas, line 230: Same comment here about aligning these methods to a unified API that returns `Curve` always

gonzalocasas (changes requested):
generally lgtm except the scary commented out context code. The other things are more like an open question

src/compas/datastructures/graph/graph.py
```diff
@@ -159,7 +161,13 @@ def __from_data__(cls, data):
         graph._max_node = data.get("max_node", graph._max_node)
         return graph
 
-    def __init__(self, default_node_attributes=None, default_edge_attributes=None, name=None, **kwargs):
+    def __init__(
+        self,
+        default_node_attributes=None,
+        default_edge_attributes=None,
+        name=None,
+        **kwargs,
```
gonzalocasas: what is ironpython's problem with this line?

src/compas/geometry/surfaces/cylindrical.py
tomvanmele: we don't have a functional general curve

src/compas/geometry/surfaces/spherical.py
tomvanmele: this is not possible because the base curve class just defines an api but no functionality

src/compas/scene/context.py
tomvanmele: viewer and notebook set the context explicitly so they are unaffected. the fact that we return the viewer instead of none is just by coincidence and actually masks bugs and creates unpredictable behavior

tomvanmele (commented):
apparently i started a review so i have to leave a comment to submit my comments :)

src/compas/datastructures/graph/graph.py
tomvanmele: it can't deal with the trailing comma

src/compas/scene/context.py
Licini: Just to confirm, I did few tests with viewer too they are not affected

src/compas/geometry/surfaces/cylindrical.py
tomvanmele: afaik that is also not possible. you need a way to evaluate the curve and this depends on the type of curve it is...

src/compas/datastructures/graph/graph.py
gonzalocasas: 😝

src/compas/scene/context.py
gonzalocasas: Good, then let's delete instead of comment out

gonzalocasas (approved):

Issue #1456, opened 2025-04-06
bbox_numpy to mesh.aabb method integration or Box class extra features


2twenity:
# Feature Request

Hello everyone, I'd like to implement some extra features as a developer to Box so the mesh.aabb returns more flexible object. Ideal case would be a box that is fully based on numpy and represented simply as 2 diagonal points (2 numpy arrays). Maybe some flag could be added to mesh.aabb so we can choose which sort of a box we expect. 

## Details

**Problem statement**
I'm playing with compas_model bvh.from_meshes method and trying to implement bvh based on aabb nodes. To calculate parameters like SAH (Surface Are Heuristic) the following box methods are required:
1) Boxes summing or union method
2) Box surface area
3) Box centroid

**Solution Proposal**
I'll paste some rough logic of what is supposed to happen

1) __add__ method or "union" which will allow to sum 2 instances of a Box class
```python
def __add__(self, box):
    """Returns a 3D summary box that contains both of summed boxes"""
    if isinstance(box, SuperchargedBox):
        all_points = self.points + box.points
            return self.from_points(all_points)
    else:
        raise TypeError("Instances belong to different classes")
```

2) Surface are calculations
```python
def compute_surface_area(self):
        return (self.xsize * self.ysize + self.xsize * self.zsize + self.ysize * self.zsize) * 2
```

3) Centroid
```python
def centroid(self) -> Point:
        """Returns a Centroid of the Box"""
        arr_max = np.asarray([self.xmax, self.ymax, self.zmax])
        arr_min = np.asarray([self.xmin, self.ymin, self.zmin])

        res = (arr_max + arr_min) / 2
        return Point(*res)
```

**Alternatives**
If the box is represented as 2 numpy arrays (2 points of the diagonal). So none of the extra calculations called on constructors methods. It'll also allow to make infinitely small box and infinitely big boxes for summing/union operations.  

Also, as I mentioned, some sort of flexibility when making calling a mesh.aabb would be useful.
```python
Mesh.aabb(numpy_box=True) -> bbox_numpy
```
Or these changes could be made in compas_model. I'd appreciate your feedback on these changes.

tomvanmele:
hi @2twenity 

i will have a look at your proposal asap (probably not before early next week though)

best,
tom

PR #1473, opened 2025-06-08
added bounding box computations for Datastructure class, added to_point method for Mesh class

domrice:

### Checklist

### DIFF
--- a/tests/compas/datastructures/test_mesh.py
+++ b/tests/compas/datastructures/test_mesh.py
@@ -1289,6 +1289,9 @@ def test_face_attributes_includes_all_defaults(box):
 
 
 def test_compute_aabb():
+    if compas.IPY:
+        pytest.skip("Skipping test under IronPython.")
+
     mesh = Mesh.from_obj(compas.get("tubemesh.obj"))
     aabb = mesh.compute_aabb()
 
@@ -1298,6 +1301,9 @@ def test_compute_aabb():
 
 
 def test_compute_obb():
+    if compas.IPY:
+        pytest.skip("Skipping test under IronPython.")
+
     mesh = Mesh.from_obj(compas.get("tubemesh.obj"))
     obb = mesh.compute_obb()
 
--- a/tests/compas/datastructures/test_mesh.py
+++ b/tests/compas/datastructures/test_mesh.py
@@ -1287,26 +1287,20 @@ def test_face_attributes_includes_all_defaults(box):
 # bounding volumes
 # --------------------------------------------------------------------------
 
+if not compas.IPY:
 
-def test_compute_aabb():
-    if compas.IPY:
-        pytest.skip("Skipping test under IronPython.")
+    def test_compute_aabb():
+        mesh = Mesh.from_obj(compas.get("tubemesh.obj"))
+        aabb = mesh.compute_aabb()
 
-    mesh = Mesh.from_obj(compas.get("tubemesh.obj"))
-    aabb = mesh.compute_aabb()
+        assert isinstance(aabb, Box)
+        assert len(aabb.points) == 8
+        assert aabb.contains_points(mesh.to_points())
 
-    assert isinstance(aabb, Box)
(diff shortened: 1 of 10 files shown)


gonzalocasas (approved):
LGTM! Thanks!

gonzalocasas:
@tomvanmele unless you have any additional feedback on this one, I think we can merge it

domrice:
Question: Should we add test methods for `Datastructure.compute_obb()` and `Datastructure.compute_aabb()`? It was not clear to me where to add them.

gonzalocasas:
@dominikreisach you could add them, yes. I would add them to `test_mesh.py` (despite the fact that the entry point is in the parent class).

domrice:
@gonzalocasas Surely, I will create the tests it right away.

I also found a feature request that we could implement in the same turn, see [here](https://github.com/compas-dev/compas/issues/1456).

tomvanmele:
> @tomvanmele unless you have any additional feedback on this one, I think we can merge it

sure, will have a look later today...

tomvanmele:
`numpy` is imported in a location not liked by IronPython...

tests/compas/datastructures/test_mesh.py
```diff
@@ -1264,3 +1281,26 @@ def test_face_attributes_includes_all_defaults(box):
...
+    assert aabb.contains_points(mesh.to_points())
+
+
+def test_compute_obb():
+    mesh = Mesh.from_obj(compas.get("tubemesh.obj"))
+    obb = mesh.compute_obb()
+
+    assert isinstance(obb, Box)
+    assert len(obb.points) == 8
+    assert obb.contains_points(mesh.to_points())
```
tomvanmele: IronPython will not like this

tomvanmele:
the PR is good, we just have to resolve one of the API conflicts before we can merge.

currently, `Datastructure` defines (as it should), the properties `aabb` and `obb` with corresponding compute methods `compute_aabb` and `compute_obb`.

in `Mesh`, the properties `aabb` and `obb` are masked by the methods `aabb()` abnd `obb()`. in my opinion the latter should be removed. technically that would be a breaking change though...

tomvanmele:
btw, to make it even better, ideally, `to_points` is also added to/implemented in `Graph`, `CellNetwork`, `VolMesh` and `Tree`

domrice:
> btw, to make it even better, ideally, `to_points` is also added to/implemented in `Graph`, `CellNetwork`, `VolMesh` and `Tree`

Done for `Graph`, `CellNetwork`, and `VolMesh`. It is not clear to me how to extract points from the `Tree` or `TreeNode`classes though.

How do we resolve the IronPython issue in the test function?

tomvanmele:
yes, indeed, including `Tree` is not meaningful.
to deal with IronPython you have to make the test conditional using `if not compas.IPY`...

tomvanmele:
> in `Mesh`, the properties `aabb` and `obb` are masked by the methods `aabb()` abnd `obb()`. in my opinion the latter should be removed. technically that would be a breaking change though...

@gonzalocasas opinions?

gonzalocasas:
> > in `Mesh`, the properties `aabb` and `obb` are masked by the methods `aabb()` abnd `obb()`. in my opinion the latter should be removed. technically that would be a breaking change though...
> 
> @gonzalocasas opinions?

I understand, I guess it's borderline ok to go for this breaking change, we should document it well-enough in the CHANGELOG

domrice:
> yes, indeed, including `Tree` is not meaningful. to deal with IronPython you have to make the test conditional using `if not compas.IPY`...

Thanks, all set now.

IMHO, what if we just add this for now and remove the masking methods `aabb()` and `obb()` in `Mesh` for compas 3?

gonzalocasas:
> IMHO, what if we just add this for now and remove the masking methods `aabb()` and `obb()` in `Mesh` for compas 3?

That's very reasonable indeed

PR #1505, opened 2025-12-10
plane_to_rhino bug: add missing y-axis vector argument

jwwang0-0:

- [X] Bug fix in a **backwards-compatible** manner.

This PR fixes an issue where plane_to_rhino() constructed a Rhino.Geometry.Plane with only two arguments (origin point + x-axis). Rhino’s Plane constructor requires three arguments:

- origin point
- x-axis
- y-axis

The current code will direct to another constructor :

public Plane(
[Point3d ] origin
[Vector3d ] normal
)

And will cause troubles like 

image

### Checklist

### DIFF
--- a/src/compas_rhino/conversions/geometry.py
+++ b/src/compas_rhino/conversions/geometry.py
@@ -57,7 +57,7 @@ def plane_to_rhino(plane):
     :rhino:`Rhino.Geometry.Plane`
 
     """
-    return Rhino.Geometry.Plane(point_to_rhino(plane[0]), vector_to_rhino(plane[1]))
+    return Rhino.Geometry.Plane(point_to_rhino(plane[0]), vector_to_rhino(plane[1]), vector_to_rhino(plane[2]))
 
 
 def frame_to_rhino_plane(frame):


src/compas_rhino/conversions/geometry.py
```diff
@@ -57,7 +57,7 @@ def plane_to_rhino(plane):
     :rhino:`Rhino.Geometry.Plane`
 
     """
-    return Rhino.Geometry.Plane(point_to_rhino(plane[0]), vector_to_rhino(plane[1]))
+    return Rhino.Geometry.Plane(point_to_rhino(plane[0]), vector_to_rhino(plane[1]), vector_to_rhino(plane[2]))
```
tomvanmele, line 60: a COMPAS plane only has a point and a normal (it is not a frame). index 0 will return the base point, and index 1, the normal.

accessing index 2 on a plane object will raise a `KeyError`.

tomvanmele: the redirection to the constructor that uses a point and a normal is correct. i think the functionality you are looking for is `frame_to_rhino`...

jwwang0-0: Oh, thank you for the explanation! I'm wrong. That makes sense. I think it would be better to include a data type check in the code to verify whether the input is a compas_plane or a compas_frame. As it stands, it’s a bit confusing, especially since in Rhino plane is defined in a way that’s more similar to a compas_frame.

gonzalocasas:
Should we close this? Or is anyone planning to add the type check as a fix instead?

tomvanmele:
I don't think there is anything to be fixed... it is true that the plane/frame thing in relation to rhino is confusing, but that is because rhino doesn't distinguish between the two.

i think it can be closed...

PR #1506, opened 2025-12-11
Fixed Blender installer

gonzalocasas:

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+* Fixed Blender installation process to support the new "Extensions" mechanism in Blender 4.2+.
+
 ### Removed
 
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -89,6 +89,8 @@ current_version = "2.15.0"
 message = "Bump version to {new_version}"
 commit = true
 tag = true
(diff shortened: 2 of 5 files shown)


gonzalocasas:
What do you guys think of going in this direction? 

https://github.com/user-attachments/assets/e73b296c-a8c7-489a-bc91-66c8fe0ae663

tomvanmele:
> What do you guys think of going in this direction?

can you describe the gist of the new approach? not 100% clear to me from the video...

gonzalocasas:
> > What do you guys think of going in this direction?
> 
> can you describe the gist of the new approach? not 100% clear to me from the video...

The new approach is what seems to be the official way to create Blender extensions. Now `compas_blender` is a blender extension because it includes a bunch of extra metadata and entrypoints, "extensions" are new in Blender 4.2+ so this only works from that version onwards. Once that is in place, we simply ship the `compas_blender` wheel without dependencies, which can be installed as shown in the video (`Edit -> Preferences -> Get extensions -> Install from disk...`). If this works, we should also release the wheel onto the blender extensions website, so that it can be installed just by searching. 

After installing it, the wheel contain much except enabling a little "Preferences" panel in the COMPAS extension, and that panel has a button to install dependencies. When you click that, the remaining dependencies are pip installed (by subprocessing pip with blender's python executable).

If you all agree with it, I would like to clean up this PR a bit, and try to get it merged and release on 2.x, because the Blender installation process we have documented didn't work well for me and it's very convoluted.

tomvanmele:
i agree 💯

PR #1521, opened 2026-06-16
Deferred loading of NumPy and SciPy dependencies

josemmo:
## Problem

`import compas.geometry` takes several seconds due to top-level NumPy and SciPy imports in `compas.geometry` and `compas.data`. This triggers a cascade of around 470 transitive imports at startup, even when those libraries are not used.
 
While the COMPAS codebase already defers heavy imports in several places[^1][^2][^3], some modules don't follow this practice, which defeats the effort made elsewhere.

[^1]: https://github.com/compas-dev/compas/blob/e03e40975ebec56ab98b4cfaead1eedf70241cb4/src/compas/geometry/polyhedron.py#L395
[^2]: https://github.com/compas-dev/compas/blob/e03e40975ebec56ab98b4cfaead1eedf70241cb4/src/compas/geometry/_core/distance.py#L595
[^3]: https://github.com/compas-dev/compas/blob/e03e40975ebec56ab98b4cfaead1eedf70241cb4/src/compas/datastructures/mesh/mesh.py#L404

## Impact of the PR
 
|         | Import time |
|---------|-------------|
| Before  | ~2.21s      |
| After   | ~0.14s      |
| Speedup | **~15x**    |

## How to reproduce

1. Setup a blank project:

```bash
uv init
uv add compas==2.15.1 snakeviz
```

2. Profile with SnakeViz:

```ps1
# Clear cache first for a cold-start measurement
Get-ChildItem -Directory -Recurse -Force -Filter "__pycache__" -ErrorAction SilentlyContinue | Remove-Item -Recurse -Force -ErrorAction SilentlyContinue

uv run python -c 'import cProfile; cProfile.run("import compas.geometry", "before.prof")'
uv run snakeviz before.prof
```

The profile will show `_io.open_code` dominating runtime across 500+ file reads. This is Python pulling in source/bytecode files from NumPy/SciPy from disk.

For my particular device, it took 2.21 seconds to run `import compas.geometry`. Note this time may vary significant depending on the machine hardware and cache status.

3. Capture an import time log:

```ps1
uv run python -X importtime -c "import compas.geometry" 2>&1 | Out-File -FilePath before.txt -Encoding utf8
```

The log will contain around 470 NumPy/SciPy imports with high cumulative times. For example:

```
import time: self [us] | cumulative | imported package
import time:      1271 |      63580 | numpy
import time:     27278 |      39595 | numpy.testing._private.utils
import time:       507 |      49062 | numpy.testing
import time:       727 |     160679 | scipy.linalg
import time:       694 |      94077 | scipy.optimize
import time:       439 |      20286 | scipy.spatial.transform
import time:       539 |      79419 | scipy.spatial
```

## How to verify fix

Repeat the steps above after switching to this branch:
 
```bash
uv add "git+https://github.com/josemmo/compas@performance/deferred-loading"
```
 
The profile should be significantly shorter (0.14 seconds in my case) and the import log should contain no NumPy or SciPy entries.

## Alternatives considered
 
- **[`lazy_loader`](https://github.com/scientific-python/lazy-loader) ([SPEC 1](https://scientific-python.org/specs/spec-0001/))**: The approach used by NumPy, scikit-image, NetworkX, and other libraries. Compatible with Python 3.9+, but would require restructuring large parts of the codebase.
 
- **`lazy` keyword ([PEP 810](https://peps.python.org/pep-0810/))**: Native language support, but requires Python 3.15+, which is outside COMPAS's current support range.

### DIFF
--- a/src/compas_rhino/geometry/brep/trim.py
+++ b/src/compas_rhino/geometry/brep/trim.py
@@ -8,7 +8,6 @@ from compas.geometry import BrepTrim
 from compas_rhino.geometry import RhinoNurbsCurve
 
 from .edge import RhinoBrepEdge
-
 from .vertex import RhinoBrepVertex
 
 
--- a/src/compas_rhino/geometry/curves/curve.py
+++ b/src/compas_rhino/geometry/curves/curve.py
@@ -6,6 +6,7 @@ import Rhino.Geometry  # type: ignore
 
 from compas.geometry import Curve
 from compas.geometry import Plane
+from compas_rhino.conversions import ConversionError
 from compas_rhino.conversions import box_to_compas
 from compas_rhino.conversions import plane_to_compas_frame
 from compas_rhino.conversions import plane_to_rhino
@@ -14,7 +15,6 @@ from compas_rhino.conversions import point_to_rhino
 from compas_rhino.conversions import polyline_to_compas
 from compas_rhino.conversions import transformation_to_rhino
 from compas_rhino.conversions import vector_to_compas
-from compas_rhino.conversions import ConversionError
 
 
 class RhinoCurve(Curve):
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Renamed `Tolerance.units` to `Tolerance.unit` to better reflect the documented properties. Left `units` with deprecation warning.
 * Fixed `NotImplementedErorr` when calling `BrepLoop.vertices`.
 * Fixed `python -m compas` to detect extensions based on `importlib` rather than `pkg_resources`.
+* Deferred loading of NumPy and SciPy dependencies.
 
 ### Removed
 
--- a/src/compas/data/encoders.py
+++ b/src/compas/data/encoders.py
@@ -14,7 +14,7 @@ from .data import Data  # noqa: F401
 from .exceptions import DecoderError
 
 IDictionary = None
-numpy_support = False
+numpy_support = None  # type: bool | None
 dotnet_support = False
 
 # We don't do this from `compas.IPY` to avoid circular imports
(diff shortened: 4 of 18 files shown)


src/compas_rhino/geometry/curves/curve.py
tomvanmele, line 1: i guess this is included by accident?

src/compas_rhino/geometry/brep/trim.py
tomvanmele, line 1: i guess this is included by accident?

josemmo: (See previous comment)

src/compas_rhino/geometry/curves/curve.py
josemmo: No, this is intentional. The linter of the repository applied those corrections during the pre-commit hook.

I put them in a separate commit (2f5932258fd62cfae482104e77a0df2faa5063fa) in case you want to cherry pick them.

josemmo:
> can we add a benchmark to check the impact on repeated calls to functions like `normrow`, `transform_points_numpy`, `trimesh_pull_points_numpy`, ...?

Thanks for raising this point!

I don't think a benchmark is necessary here because Python caches all previously imported modules in `sys.modules`. On subsequent calls to any of these functions, an import statement is effectively a dictionary lookup that takes tenths of nanoseconds, which is completely negligible compared to the time spent on the underlying NumPy/SciPy operations that those functions perform.

Unless functions like `normrow`/`transform_points_numpy`/etc are called millions of times in a "for" loop or 60 times per second in an infinite loop, this change should not have any negative impact at all. That said, if there's a specific case you're worried about, please let me know and I'd be happy to take a look :)

josemmo:
Hi @chenkasirer,

> would it be on the first call to a numpy-backed function?

Yep, it will load the module on the first call to a function importing NumPy/SciPy.

---

> seems easy to re-introduce an import to numpy somewhere and re-introduce this import delay

Luckily for us, there's a Ruff rule precisely for this! https://docs.astral.sh/ruff/rules/banned-module-level-imports/

If you're happy with this approach, I can update the PR to modify the pyproject.toml file:

```toml
[tool.ruff.lint]
select = ["E", "F", "I", "TID253"]

[tool.ruff.lint.flake8-tidy-imports]
banned-module-level-imports = ["numpy", "scipy"]
```

PR #1512, opened 2026-02-17
Base Py3 upgrade

tomvanmele:
### What type of change is this?

- [ ] Bug fix in a **backwards-compatible** manner.
- [ ] New feature in a **backwards-compatible** manner.
- [ ] Breaking change: bug fix or new feature that involve incompatible API changes.
- [ ] Other (e.g. doc update, configuration, etc)

### Checklist

_Put an `x` in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your code._

- [ ] I added a line to the `CHANGELOG.md` file in the `Unreleased` section under the most fitting heading (e.g. `Added`, `Changed`, `Removed`).
- [ ] I ran all tests on my computer and it's all green (i.e. `invoke test`).
- [ ] I ran lint on my computer and there are no errors (i.e. `invoke lint`).
- [ ] I added new functions/classes and made them available on a second-level import, e.g. `compas.datastructures.Mesh`.
- [ ] I have added tests that prove my fix is effective or that my feature works.
- [ ] I have added necessary documentation (if appropriate)

PR #1524, opened 2026-07-06
Fix pca_numpy conditioning: SVD the centered data matrix, not the covariance matrix

### LINKED ISSUE #1522: bestfit_plane_numpy returns wrong normals on near-collinear point clouds (covariance step squares the conditioning)
## Describe the bug

`bestfit_plane_numpy` (via `pca_numpy`) fits by building the covariance matrix and SVD-ing it:

```python
# src/compas/geometry/pca_numpy.py
C = Y.T.dot(Y) / (n - 1)
u, s, vT = svd(C, full_matrices=False)
```

Forming `Y.T @ Y` squares the condition number: a point cloud whose second in-plane extent is ~1e-8 of its first (a thin sliver — still a perfectly well-defined plane) has a covariance eigenvalue ratio of ~1e-16, at machine epsilon, so the smallest principal direction is lost to rounding and the returned plane normal tips into the sliver plane. An SVD of the centered data matrix `Y` itself (whose right singular vectors are mathematically the same) only has to resolve a ~1e-8 singular-value ratio and recovers the normal to full precision.

Slivers like this are routine in AEC data — footprints of thin panels, wall edges, extracted face outlines — which is where we hit it (migrating a timber-engineering geometry engine onto COMPAS; our legacy data-matrix fit and `bestfit_plane_numpy` disagreed by up to 1.31 per normal component on real inputs).

## To Reproduce

```python
import numpy as np
from compas.geometry import bestfit_plane_numpy

rng = np.random.default_rng(7)

# A thin sliver in a known plane: long axis ~1.0, short in-plane axis ~1e-8.
n = 200
t = rng.uniform(-1.0, 1.0, n)          # long in-plane direction
s = rng.uniform(-1e-8, 1e-8, n)        # short in-plane direction (near-collinear)
pts_local = np.column_stack([t, s, np.zeros(n)])  # exactly planar, normal = +Z

# Rotate into a generic orientation so nothing is axis-aligned.
a, b = 0.6, -1.1
Rx = np.array([[1, 0, 0], [0, np.cos(a), -np.sin(a)], [0, np.sin(a), np.cos(a)]])
Rz = np.array([[np.cos(b), -np.sin(b), 0], [np.sin(b), np.cos(b), 0], [0, 0, 1]])
R = Rz @ Rx
pts = pts_local @ R.T + np.array([100.0, -40.0, 7.0])
true_normal = R @ np.array([0.0, 0.0, 1.0])

_, normal = bestfit_plane_numpy(pts)
normal = np.asarray(normal) / np.linalg.norm(normal)

Y = pts - pts.mean(axis=0)
normal_svd = np.linalg.svd(Y, full_matrices=False)[2][2]

for name, nrm in (("bestfit_plane_numpy", normal), ("data-matrix SVD", normal_svd)):
    ang = np.degrees(np.arccos(np.clip(abs(nrm @ true_normal), -1, 1)))
    print(f"{name:20s}: angle off = {ang:.6f} deg")
```

Output (compas 2.15.1, numpy 2.4.6, CPython 3.13, macOS arm64):

```
bestfit_plane_numpy : angle off = 11.638995 deg
data-matrix SVD     : angle off = 0.000001 deg
```

Shrinking the short axis to `1e-9` makes it worse (`37.28 deg` vs `0.000004 deg`). The points are *exactly* planar in both cases — the fit problem is perfectly posed; only the covariance construction loses it.

Confirmed present on current `main` as well as 2.15.1 (the `C = Y.T.dot(Y) / (n - 1)` construction is unchanged).

## Expected behavior

The normal of an exactly-planar (if skinny) point cloud should be recovered — the direct SVD of the centered coordinates does so to machine precision.

## Suggested fix

In `pca_numpy`, run the SVD on the centered data matrix directly:

```python
u, s, vT = svd(Y, full_matrices=False)   # right singular vectors == eigenvectors of C
eigenvalues = (s ** 2) / (n - 1)          # if the current return contract needs them
```

Right singular vectors of `Y` are identical (in exact arithmetic) to the eigenvectors of `C`, so callers see the same results on well-conditioned inputs — only the ill-conditioned ones improve. Cost-wise both routes are O(n) once the `Y.T @ Y` product is counted; for typical point-cloud sizes the difference is noise.

Happy to open a PR with the change + tests along these lines if you agree with the direction.

mengxihex:
Fixes #1522.

## What

`pca_numpy` formed the covariance matrix `C = Y.T @ Y / (n - 1)` and SVD'd it. Forming `C` squares the condition number, so for near-degenerate inputs (an almost collinear point cloud whose smallest extent is ~1e-8 of its largest — still an exactly planar, perfectly posed fit) the smallest principal direction drowns at machine epsilon and comes back wrong: `bestfit_plane_numpy` normals were off by 11.6° (1e-8 aspect) to 37.3° (1e-9 aspect) on the reproduction in #1522.

This PR runs the SVD on the centered data matrix `Y` directly. The right-singular vectors of `Y` are exactly the eigenvectors of `C`, and the singular values are rescaled (`s**2 / (n - 1)`) so the returned eigenvalues keep their meaning (variance along each principal direction). Well-conditioned inputs return the same results as before; only the ill-conditioned ones improve. Cost is equivalent (both routes are O(n·dim²)).

## Checks

- New `tests/compas/geometry/test_pca_numpy.py`: (1) well-conditioned parity with the covariance route (eigenvalues match `eigvalsh(C)`, eigenvectors diagonalize `C`), (2) the near-collinear cloud recovers its smallest direction to <1e-4°, (3) `bestfit_plane_numpy` regression on the same cloud (fails on the previous implementation with ~11.6° error).
- Full `tests/compas/geometry` suite: 519 passed (includes the `pca_numpy` consumers `bestfit_numpy`, `bbox_numpy`, `icp_numpy`).
- CHANGELOG entry added under Unreleased → Changed.

One behavioural note for review: SVD sign indeterminacy means individual principal directions may flip sign relative to the old implementation on some inputs (the sign was already unspecified in both routes).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+* Changed `pca_numpy` to run the SVD on the centered data matrix instead of the covariance matrix: forming the covariance squares the condition number, so near-degenerate inputs (e.g. almost collinear point clouds) returned wrong principal directions — `bestfit_plane_numpy` normals were off by 11–37 degrees on exactly planar sliver clouds (#1522). Well-conditioned results are unchanged (same eigenvectors; eigenvalues rescaled to keep their variance meaning).
 * Changed `Tolerance` class to no longer use singleton pattern. `Tolerance()` now creates independent instances instead of returning the global `TOL`. 
 * Renamed `Tolerance.units` to `Tolerance.unit` to better reflect the documented properties. Left `units` with deprecation warning.
 * Fixed `NotImplementedErorr` when calling `BrepLoop.vertices`.
--- a/src/compas/geometry/pca_numpy.py
+++ b/src/compas/geometry/pca_numpy.py
@@ -54,34 +54,32 @@ def pca_numpy(data):
     # across all observations
     Y = X - mean
 
-    # covariance matrix of spread
-    # note: there is a covariance function in NumPy...
-    # the shape of the covariance matrix is dim x dim
-    # for example, if the data are 2D point coordinates, the shape of C is 2 x 2
-    # the diagonal of the covariance matrix contains the variance of each variable
-    # the off-diagonal elements of the covariance matrix contain the covariance
-    # of two independent variables
-    C = Y.T.dot(Y) / (n - 1)
-
-    # assert C.shape[0] == dim, "The shape of the covariance matrix is not correct."
-
-    # SVD of covariance matrix
-    u, s, vT = svd(C, full_matrices=False)
+    # SVD of the spread matrix directly — deliberately NOT of the covariance
+    # matrix C = Y.T @ Y / (n - 1).
+    # The right-singular vectors of Y are exactly the eigenvectors of C, but
+    # forming C squares the condition number: for near-degenerate data (e.g.
+    # an almost collinear point cloud, whose smallest extent is ~1e-8 of its
+    # largest) the smallest principal direction drowns in floating-point
+    # rounding and the returned directions are wrong, while the SVD of Y
+    # recovers them to full precision. See issue #1522.
+    u, s, vT = svd(Y, full_matrices=False)
 
     # eigenvectors
     # ------------
     # note: the eigenvectors are normalized
     # note: vT is exactly what it says it will be => the transposed eigenvectors
     # => take the rows of vT, or the columns of v
-    # the right-singular vectors of C (the columns of V or the rows of Vt)
-    # are the eigenvectors of CtC
+    # the right-singular vectors of Y (the columns of V or the rows of Vt)
+    # are the eigenvectors of C = Y.T @ Y / (n - 1)
     eigenvectors = vT
 
     # eigenvalues
     # -----------
-    # the nonzero singular values of C are the square roots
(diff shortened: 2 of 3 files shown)


mengxihex:
Fixed the `windows-ironpython` failure: the new test file used the `@` matmul operator, which is a parse-time SyntaxError under IronPython 2.7. It now follows the `test_bbox.py` idiom — py2-parseable, `if compas.IPY: return` guards, numpy imported inside the guarded test bodies. Verified locally: 3/3 new tests + the full `tests/compas/geometry` suite pass on CPython. Workflows on the new commit are awaiting approval.

gonzalocasas:
@tomvanmele ping ;)

tomvanmele:
@mengxihex perhaps add yourself to the author list :)

mengxihex:
thanks a lot @tomvanmele! Long-time admirer of compas, the abstraction layer AEC has been missing to scale. Glad to contribute.

tomvanmele:
btw @mengxihex i am currently modernizing the codebase in this PR https://github.com/compas-dev/compas/pull/1512
a big part of this will be the improved integration of numpy/scipy in the core algorithms, without having to care about ipy and/or py2. if you have opinions there, feel free to weigh in...

